PR #5, opened 2025-04-25
Initial Plumbing for vfio-user Server

blitz:
This PR adds:

- a dummy XHCI device
- the necessary boilerplate to make this a `vfio-user` `Backend`
- the plumbing to start the `vfio-user` server on startup.

This is enough to get the device recognized by Linux in the VM. `usbvfiod` will then crash, because it triggers unimplemented code paths. But progress!

One important issue to be fixed to make meaningful progress is this: https://github.com/rust-vmm/vfio-user/pull/34 If this isn't merged soon, we might have to fork it.

### DIFF
--- a/src/cli.rs
+++ b/src/cli.rs
@@ -3,7 +3,10 @@
 //! The main external constraint here is that we need to be compatible
 //! to the vfio-user [Backend Program
 //! Conventions](https://github.com/nutanix/libvfio-user/blob/master/docs/vfio-user.rst#backend-program-conventions).
-use std::path::PathBuf;
+use std::{
+    os::fd::RawFd,
+    path::{Path, PathBuf},
+};
 
 use clap::Parser;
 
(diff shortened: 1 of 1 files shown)


src/cli.rs
```diff
@@ -13,4 +20,10 @@ pub struct Cli {
     /// increase verbosity.
     #[arg(short, long, action = clap::ArgAction::Count)]
     pub verbose: u8,
+
+    /// The path where to create a listening Unix domain socket.
+    ///
+    /// This is the path where Cloud Hypervisor will connect to usbvfiod.
+    #[arg(short, long)]
+    pub socket_path: PathBuf,
```
snue, line 28: nit: we are leaving out the `fd` option on purpose currently?

I prepared the following recently:

```suggestion
    /// Provide the vfio-user socket as file-descriptor.
    #[arg(long, conflicts_with = "socket_path")]
    pub fd: Option<RawFd>,
    /// Provide the vfio-user socket as path.
    #[arg(long, required_unless_present = "fd")]
    pub socket_path: Option<PathBuf>,
```

blitz: We agreed to implement the command line option and trigger a `todo!` when someone uses it.

snue (approved):

PR #8, opened 2025-05-02
Add Smoke Test

blitz:
Add a simple smoke test that checks whether the USB controller pops up in the VMs PCI config space.

 As the xhci driver currently hangs for quite a long time, we only
check whether the kernel sees it during boot up. This makes the test succeed in a couple of seconds. If we wait for VM userspace to come up, this currently takes more than a minute.

We can refactor the test later when the xhci driver attaches cleanly.

@xanderio The Github CI currently needs 80 seconds to run the NixOS integration test. Should I set up a Hercules CI runner?

### DIFF
--- a/nix/tests.nix
+++ b/nix/tests.nix
@@ -1,0 +1,96 @@
+# This file contains integration tests for usbvfiod.
+{ lib, pkgs, usbvfiod }:
+let
+  # For the VM that we start in Cloud Hypervisor, we re-use the netboot image.
+  netbootNixos = lib.nixosSystem {
+    inherit (pkgs) system;
+
+    modules = [
+      "${pkgs.path}/nixos/modules/installer/netboot/netboot-minimal.nix"
+
+      # Cloud Hypervisor Guest Convenience
+      ({ config, ... }: {
+
+        boot.kernelParams = [
+          # Use the serial console for kernel output.
+          #
+          # The virtio-console is an option as well, but is not
+          # compiled into the NixOS kernel and would be inconvenient.
+          "console=ttyS0"
+        ];
+
+        # Enable debug verbosity.
+        boot.consoleLogLevel = 7;
+
+        # Silence the useless stateVersion warning. We have no state to keep.
+        system.stateVersion = config.system.nixos.release;
+      })
+    ];
+  };
+
+  netboot =
+    let
+      inherit (netbootNixos) config;
+
+      kernelTarget = pkgs.stdenv.hostPlatform.linux-kernel.target;
+    in
+    {
+      initrd = "${config.system.build.netbootRamdisk}/initrd";
+      kernel = "${config.system.build.kernel}/${kernelTarget}";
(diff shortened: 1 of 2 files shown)


nix/tests.nix
```diff
@@ -0,0 +1,106 @@
...
+      };
+    };
+
+    testScript = ''
+      start_all()
+
+      machine.wait_for_unit("cloud-hypervisor.service")
+
+      # Check whether the USB controller pops up.
+      machine.wait_until_succeeds("grep -Fq 'pci 0000:00:02.0: [1b36:000d]' ${cloudHypervisorLog}")
```
blitz, line 103: We could also poll `journalctl` if we want to avoid the log file in `/tmp`.

nix/tests.nix
```diff
@@ -0,0 +1,106 @@
...
+              --cmdline ${lib.escapeShellArg netboot.cmdline} \
+              --initramfs ${netboot.initrd} \
+              --user-device socket=${usbvfiodSocket}
+          '';
+        };
+      };
+
+      virtualisation = {
+        cores = 2;
+        memorySize = 8192;
```
snue, line 93: This (and the 4G guest) seem a little excessive for the simple functionality test. Could we get away with 2G/1G for host/guest? A lot of cloud-hypervisor integration tests seem to work fine with 512MiB guests.

nix/tests.nix
```diff
@@ -0,0 +1,106 @@
...
+      };
+    };
+
+    testScript = ''
+      start_all()
```
snue, line 98: nit: it seems the first access of the `machine` will also just start it if necessary, and this line is not strictly required.

nix/tests.nix
snue: I have no strong preference here currently. The logfile in `/tmp` is fine.

nix/tests.nix
```diff
@@ -0,0 +1,106 @@
...
+
+    nodes.machine = { pkgs, ... }: {
+      environment.systemPackages = [
+        pkgs.cloud-hypervisor
+        pkgs.tmux
+
+        (pkgs.writeShellScriptBin "start-chv" ''
+          exec cloud-hypervisor  \
+
+        '')
```
snue, line 62: Do you use this in interactive sessions? Doesn't look very useful to me.

snue (commented):
Left some comments/questions. The approach looks good to me. Thanks for carefully pointing out the pitfalls regarding using this as a production setup example.

nix/tests.nix
blitz: This is a left-over. Will fix.

nix/tests.nix
blitz: The initramfs is 512M compressed. I've reduced this to 2/4GiB for guest/host. We can finetune this later. With the Nix store on a block device, we can go way down.

snue (approved):
LGTM. Test succeeds and runs as part of CI.

PR #9, opened 2025-05-06
DMA Support

blitz:
This PR works towards the infrastructure for the virtual XHCI controller to be able to access guest memory. For this we introduce:

- `MemorySegment`: a wrapper around a mmap-ed memory
- `DynamicBus`: a wrapper around `Bus` that allows modifying the bus

We assemble `MemorySegments` into a bus and then hand this to the XHCI controller as target for DMA.

I would cleanup the structure of the crate a bit next. We have the Particle device emulation code in the `usbvfiod` library and the other components just as modules in the binary. This is a bit awkward now. But I didn't want to shuffle this around in this PR as well. :)

## TODO

- [x] Finish implementation of `MemorySegment` (and add tests)
- [x] Add `SAFETY` comments for any unsafe code.
- [x] Create a `Bus` of all memory regions we get from `dma_map` (using `MemorySegment`)
- [x] Wire everything up to be used from the XHCI backend

### DIFF
--- a/src/memory_segment.rs
+++ b/src/memory_segment.rs
@@ -63,8 +63,9 @@ pub struct MemorySegment {
 impl MemorySegment {
     /// Creates a memory segment from a file.
     ///
-    /// The file is only used for mapping and will not be read or
-    /// written to.
+    /// The `File` object is only used for memory-mapping and will not
+    /// be read or written to. We only access the underlying file via
+    /// the memory mapping.
     pub fn new_from_fd(
         fd: &File,
         file_offset: u64,
--- a/src/memory_segment.rs
+++ b/src/memory_segment.rs
@@ -1,0 +1,256 @@
+//! Provide a [`BusDevice`] abstraction over a piece of a mmap'able
+//! file.
+
+use std::{
+    fs::File,
+    sync::{
+        atomic::{AtomicU16, AtomicU32, AtomicU64, AtomicU8, Ordering},
+        Arc,
+    },
+};
+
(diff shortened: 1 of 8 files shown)


src/memory_segment.rs
```diff
@@ -0,0 +1,252 @@
...
+    }
+}
+
+impl BusDevice for MemorySegment {
+    fn size(&self) -> u64 {
+        self.size
+    }
+
+    fn read(&self, req: Request) -> u64 {
+        assert!(req.addr.checked_add(req.size.into()).unwrap() <= self.size);
```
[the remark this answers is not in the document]
snue: Ah, no. It's correct. This computes the first address behind the access.

src/memory_segment.rs
```diff
@@ -0,0 +1,256 @@
...
+pub struct MemorySegment {
+    size: u64,
+    mapping: Arc<Mapping>,
+}
+
+impl MemorySegment {
+    /// Creates a memory segment from a file.
+    ///
+    /// The file is only used for mapping and will not be read or
+    /// written to.
```
snue, line 67: I found this a bit confusing with the `File` abstraction vs a file descriptor. We still read/write the (anonymous/memfd) "file" through the mapping.

```suggestion
    /// The File is only used for mapping and will not be read or
    /// written to.
```

snue (approved):
:tada: Very nice!

src/memory_segment.rs
blitz: Good point. I reworded this.

PR #10, opened 2025-05-09
cargo: update to vfio-bindings 0.4.0

blitz:
This was recently released in our journey to depend less on inputs from Git repos.

I've pinned vfio-user to avoid pulling in different versions of `vmm-sys-utils`. I'll work upstream towards a release there as well.


snue (approved):
Thanks for chasing after releases for these dependencies. Update looks good.

PR #11, opened 2025-05-12
Start handling MMIO Region access

snue:
This is work on the way to make the XHCI driver attach successfully.

* Change the PCI device trait to better work with the VFIO region access
* Add constants for XHCI register offsets
* Return defaults for basic register access

This currently bails out when the XHCI driver configures the single device slot (in `CONFIG`).

### DIFF
--- a/nix/tests.nix
+++ b/nix/tests.nix
@@ -22,6 +22,9 @@ let
         # Enable debug verbosity.
         boot.consoleLogLevel = 7;
 
+        # Convenience packages for interactive use
+        environment.systemPackages = [ pkgs.pciutils pkgs.usbutils ];
+
         # Silence the useless stateVersion warning. We have no state to keep.
         system.stateVersion = config.system.nixos.release;
       })
--- a/src/device/pci/traits.rs
+++ b/src/device/pci/traits.rs
@@ -54,34 +54,21 @@ pub trait PciDevice: Debug {
 
     /// Write a value to an I/O region.
     ///
-    /// The device may or may not claim the request. If the request is claimed, this function
-    /// returns `Some(())` otherwise `None`.
-    ///
-    /// A device that only claims requests via standard BARs can use
-    /// [`try_match_bar`](super::config_space::ConfigSpace::try_match_bar) to implement this function.
-    ///
     /// # Parameters
     ///
-    /// - `kind`: Specifies the type of request.
+    /// - `region`: Identifies the targeted I/O region (BAR).
     /// - `req`: The offset and size of the request. Offsets are relative to the beginning of each
     ///          I/O region.
     /// - `value`: The value to be written.
-    #[must_use]
-    fn try_io_write(&self, kind: RequestKind, req: Request, value: u64) -> Option<()>;
+    fn write_io(&self, region: u32, req: Request, value: u64);
 
     /// Read a value from an I/O region.
     ///
-    /// The device may or may not claim the request. If the request is claimed, this function
-    /// returns `Some(value)` otherwise `None`.
-    ///
-    /// A device that only claims requests via standard BARs can use
-    /// [`try_match_bar`](super::config_space::ConfigSpace::try_match_bar) to implement this function.
-    ///
     /// # Parameters
     ///
-    /// - `kind`: Specifies the type of request.
+    /// - `region`: Identifies the targeted I/O region (BAR).
     /// - `req`: The offset and size of the request. Offsets are relative to the beginning of each
     ///          I/O region.
     #[must_use]
-    fn try_io_read(&self, kind: RequestKind, req: Request) -> Option<u64>;
+    fn read_io(&self, region: u32, req: Request) -> u64;
 }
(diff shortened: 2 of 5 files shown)


nix/tests.nix
```diff
@@ -22,6 +22,9 @@ let
         # Enable debug verbosity.
         boot.consoleLogLevel = 7;
 
+        # Convenience packages for interactive use
+        environment.systemPackages = [ pkgs.pciutils pkgs.usbutils ];
```
phip1611, line 26:
```suggestion
        environment.systemPackages = with pkgs; [
          pciutils 
          usbutils
        ];
```

src/device/pci/config_space.rs
```diff
@@ -222,6 +222,36 @@ impl ConfigSpaceBuilder {
         self
     }
 
+    /// Add a Base Address Register (BAR) for a non-prefetchable 64-bit memory region.
+    ///
+    /// This is the typical BAR type for MMIO regions.
+    ///
+    /// Size must be a power of 2 and at least 16 bytes, but 4 KiB is the recommended minimum.
```
phip1611, line 229: I think we should follow Rust best practices for documentation

```suggestion
    /// # Arguments
    /// - `index`: ...
    /// - `size`: Size must be a power of 2 and at least 16 bytes, but 4 KiB is the recommended minimum.
```

snue: The whole function duplicates the existing 32 bit BAR logic and comment and is still missing tests. We can probably clean up the comment for both a bit.

nix/tests.nix
blitz: Instead of nit-picking on the Nix formatting we should just enable the Nix git-hooks check. :)

src/device/pci/config_space.rs
blitz: If the 64-bit BAR is not required to make progress, I'd like to remove it from this PR. The problems here are that it:

- doesn't handle BARs that are >4G
- has no tests

and of course it's not required at all right now.

src/device/pci/traits.rs
```diff
@@ -54,34 +54,27 @@ pub trait PciDevice: Debug {
...
     /// # Parameters
     ///
+    /// - `region`: Identifies the targeted I/O region (BAR).
     /// - `kind`: Specifies the type of request.
     /// - `req`: The offset and size of the request. Offsets are relative to the beginning of each
     ///          I/O region.
     /// - `value`: The value to be written.
-    #[must_use]
-    fn try_io_write(&self, kind: RequestKind, req: Request, value: u64) -> Option<()>;
+    fn write_io(&self, region: u32, kind: RequestKind, req: Request, value: u64);
```
blitz, line 66: Is `kind` relevant? I would opt to remove it. Same for `read_io`.

src/xhci_backend.rs
```diff
@@ -124,6 +137,24 @@ impl ServerBackend for XhciBackend {
                 },
             ),
 
+            0 => self.device.write_io(
+                0,
+                RequestKind::Memory,
+                Request::new(offset, RequestSize::try_from(data.len() as u64).unwrap()),
```
blitz, line 143: For integer conversions I prefer `try_into().unwrap()`, because it will do the right thing or fail loudly. Especially for `usize -> u64` this has no cost for all platforms we care for.

Otherwise, for `unwrap()` I would appreciate a comment why `unwrap` is safe.

src/device/pci/traits.rs
```diff
@@ -54,34 +54,27 @@ pub trait PciDevice: Debug {
 
     /// Write a value to an I/O region.
     ///
-    /// The device may or may not claim the request. If the request is claimed, this function
-    /// returns `Some(())` otherwise `None`.
-    ///
-    /// A device that only claims requests via standard BARs can use
-    /// [`try_match_bar`](super::config_space::ConfigSpace::try_match_bar) to implement this function.
+    /// The device must claim the request.
```
blitz, line 57: This sentence is just confusing if you don't know what was here before. I would just remove it.

blitz (changes requested):
Will approve/merge if you comment on / fix the `MemoryKind` and `unwrap` comments.

nix/tests.nix
snue: We already have git-hooks for `nixpkgs-fmt`, `deadnix`, and `statix` enabled. They run on `nix flake -L check`. I have not seen those complain. Are we missing anything that would have reformatted here?

src/device/pci/config_space.rs
snue: I reverted to the regular 32 bit bar. We still see the same access by the driver.

nix/tests.nix
blitz: No. It's fine.

blitz (approved):

src/xhci_backend.rs
blitz: Yes, it does look much worse. :(

PR #13, opened 2025-05-15
Handle more initial MMIO register access

snue:
This PR starts tracking and returning some internal state during MMIO register access. The xhci driver probe still fails.

The visible guest driver probe sequence changes a little from:

```
$ sudo dmesg | grep xhci
[    1.371933] xhci_hcd 0000:00:03.0: xHCI Host Controller
[    1.372179] xhci_hcd 0000:00:03.0: new USB bus registered, assigned bus number 1
[    1.373995] xhci_hcd 0000:00:03.0: No Extended Capability registers, unable to set up roothub
[    1.511130] xhci_hcd 0000:00:03.0: Host halt failed, -110
[    1.511380] xhci_hcd 0000:00:03.0: Host controller not halted, aborting reset.
[    1.511733] xhci_hcd 0000:00:03.0: can't setup: -12
[    1.511938] xhci_hcd 0000:00:03.0: USB bus 1 deregistered
[    1.512180] xhci_hcd 0000:00:03.0: init 0000:00:03.0 fail, -12
[    1.512431] xhci_hcd 0000:00:03.0: probe with driver xhci_hcd failed with error -12
```

to:
```
$ sudo dmesg | grep xhci 
[    1.383109] xhci_hcd 0000:00:03.0: xHCI Host Controller
[    1.383344] xhci_hcd 0000:00:03.0: new USB bus registered, assigned bus number 1
[    1.384923] xhci_hcd 0000:00:03.0: No Extended Capability registers, unable to set up roothub
[    1.385705] xhci_hcd 0000:00:03.0: can't setup: -12
[    1.385987] xhci_hcd 0000:00:03.0: USB bus 1 deregistered
[    1.386258] xhci_hcd 0000:00:03.0: init 0000:00:03.0 fail, -12
[    1.388043] xhci_hcd 0000:00:03.0: probe with driver xhci_hcd failed with error -12
```
(the host controller "halt"/reset looks correct now).

The xhci driver finishes by writing a 0 to the primary event ring segment table size, which should result in "undefined behavior" (cannot stop the primary event ring). At this point, we need to provide more capabilities to describe the XHCI controller setup.

### DIFF
--- a/src/device/pci/xhci.rs
+++ b/src/device/pci/xhci.rs
@@ -4,12 +4,13 @@
 //! [here](https://www.intel.com/content/dam/www/public/us/en/documents/technical-specifications/extensible-host-controler-interface-usb-xhci.pdf).
 
 use std::sync::Mutex;
+use tracing::{debug, warn};
 
 use crate::device::{
     bus::{BusDeviceRef, Request, SingleThreadedBusDevice},
     pci::{
         config_space::{ConfigSpace, ConfigSpaceBuilder},
-        constants::xhci::{capability, offset, OP_BASE, RUN_BASE},
+        constants::xhci::{capability, offset, operational::crcr, MAX_SLOTS, OP_BASE, RUN_BASE},
         traits::PciDevice,
     },
 };
@@ -23,6 +24,25 @@ pub struct XhciController {
 
     /// The PCI Configuration Space of the controller.
     config_space: ConfigSpace,
+
+    /// The current Run/Stop status of the controller.
+    running: bool,
+
+    /// The current Run/Stop status of the command ring.
+    command_ring_running: bool,
+
+    /// Internal Command Ring position.
+    command_ring_dequeue_pointer: u64,
+
+    /// Intrernal Consumer Cycle State for the next TRB.
+    consumer_cycle_state: bool,
+
+    /// Configured device slots.
+    slots: Vec<()>,
+
+    /// Device Context Array
+    /// TODO: currently just the raw pointer configured by the OS
+    device_contexts: Vec<u64>,
 }
 
 impl XhciController {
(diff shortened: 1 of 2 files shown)


src/device/pci/xhci.rs
```diff
@@ -23,6 +31,28 @@ pub struct XhciController {
 
     /// The PCI Configuration Space of the controller.
     config_space: ConfigSpace,
+
+    /// The current Run/Stop status of the controller.
+    running: bool,
+
+    /// The current Run/Stop status of the command ring.
+    command_ring_running: bool,
```
phip1611, line 39: Do `running` and `command_ring_running` depend on each other? Then a state machine might be a better design?

src/device/pci/xhci.rs
```diff
@@ -59,11 +182,26 @@ impl PciDevice for Mutex<XhciController> {
         assert_eq!(region, 0);
 
         match req.addr {
+            // xHC Operational Registers
```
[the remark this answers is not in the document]
snue: This stems from the simplified `SingleThreadedBusDevice` trait the `XhciController` currently makes use of. See https://github.com/cyberus-technology/usbvfiod/blob/b8492c028eb46ee811efc5965a5f14cf0d80e0c5/src/device/bus.rs#L266-L296

This helps us get going quickly and looks totally sufficient for now. We may refactor if this turns out to be a bottleneck later.

src/device/pci/xhci.rs
```diff
@@ -4,16 +4,24 @@
...
     bus::{BusDeviceRef, Request, SingleThreadedBusDevice},
     pci::{
         config_space::{ConfigSpace, ConfigSpaceBuilder},
-        constants::xhci::{offset, OP_BASE, RUN_BASE},
+        constants::xhci::{capability, offset, operational::crcr, MAX_SLOTS, OP_BASE, RUN_BASE},
         traits::PciDevice,
     },
 };
 
+/// A Basic Event Ring.
```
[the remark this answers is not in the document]
snue: It doesn't do anything yet. It just remembers what the host driver configures. The better description and more functionality is left for later.  This will be refactored once we actually handle events on that ring.

In general, we don't want to repeat too much information from the XHCI specification. The Command and Event Rings are central concepts of host controller operation. There's a single Command Ring and one Event Ring associated with each "Interruptor" (Interrupt register set).

src/device/pci/xhci.rs
```diff
@@ -41,6 +71,99 @@ impl XhciController {
...
+            event_ring: EventRing::default(),
+            slots: vec![],
+            device_contexts: vec![],
+        }
+    }
+
+    /// Obtain the current host controller status as defined for the `USBSTS` register.
+    #[must_use]
+    pub fn status(&self) -> u64 {
+        !self.running as u64 & 0x1u64
```
snue, line 87: @blitz I remembered your preference for `try_into().unwrap()` in all these cases, but they all mandate a type annotation :crying_cat_face:

blitz: `u64::from(self.running)` works. No need for `try_*`.

src/device/pci/xhci.rs
```diff
@@ -41,6 +71,99 @@ impl XhciController {
                 // TODO Should be a 64-bit BAR.
                 .mem32_nonprefetchable_bar(0, 4 * 0x1000)
                 .config_space(),
+            running: false,
+            command_ring_running: false,
+            command_ring_dequeue_pointer: 0,
+            consumer_cycle_state: false,
+            event_ring: EventRing::default(),
+            slots: vec![],
+            device_contexts: vec![],
```
blitz, line 80: You can probably replace these by `..Default::default()`.

blitz (approved):
Let's keep the ball rolling here. 👍🏻

src/device/pci/xhci.rs
```diff
@@ -88,11 +224,16 @@ impl PciDevice for Mutex<XhciController> {
 
             // xHC Operational Registers
             offset::USBCMD => 0,
-            offset::USBSTS => 0x1,   /* HCHalted */
+            offset::USBSTS => self.lock().unwrap().status(),
+            offset::CRCR => self.lock().unwrap().command_ring_status(),
+            offset::CRCR_HI => 0,
             offset::PAGESIZE => 0x1, /* 4k Pages */
-            offset::CONFIG => 0,     /* No device slots enabled */
+            offset::CONFIG => self.lock().unwrap().config(),
```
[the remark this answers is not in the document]
snue: I'll leave that for later. It would remove duplication and allow for less ad-hoc public methods. For now, we don't need to take the lock for returning a bunch of defaults.

src/device/pci/xhci.rs
snue: Not so easy, unfortunately. I can't simply derive `Default` for `XhciController` (Bus Device and Config Space don't play along nicely). And implementing Default manually doesn't look like an improvement.

PR #15, opened 2025-05-19
Make the xhci driver attach

snue:
Handle runtime interrupt registers to the point where the guest `xhci` driver attaches and exposes two virtual buses.

```
$ lsusb -t
/:  Bus 001.Port 001: Dev 001, Class=root_hub, Driver=xhci_hcd/1p, 20000M/x2
/:  Bus 002.Port 001: Dev 001, Class=root_hub, Driver=xhci_hcd/1p, 20000M/x2
```

The MSI-X capability is not exposed to the guest because the VMM fails to configure it. The `broken pipe` errors it sees are not fatal, though.

The vfio backend dies in a `todo!()`:

```
2025-05-19T16:44:07.143245Z TRACE usbvfiod::xhci_backend: read  region 0 offset 0x3020+4
2025-05-19T16:44:07.143265Z TRACE usbvfiod::xhci_backend: write region 0 offset 0x3020+4 val [2, 0, 0, 0]
2025-05-19T16:44:07.143282Z TRACE usbvfiod::xhci_backend: read  region 0 offset 0x40+4
2025-05-19T16:44:07.143302Z TRACE usbvfiod::xhci_backend: write region 0 offset 0x40+4 val [1, 0, 0, 0]
2025-05-19T16:44:07.143309Z DEBUG usbvfiod::device::pci::xhci: controller started with cmd 0x1
2025-05-19T16:44:07.143327Z TRACE usbvfiod::xhci_backend: read  region 0 offset 0x44+4
2025-05-19T16:44:07.143365Z TRACE usbvfiod::xhci_backend: read  region 0 offset 0x10+4
2025-05-19T16:44:07.145418Z TRACE usbvfiod::xhci_backend: read  region 0 offset 0x440+4

thread 'main' panicked at src/device/pci/xhci.rs:266:18:
not yet implemented
note: run with `RUST_BACKTRACE=1` environment variable to display a backtrace
```

And we see a warning with stack trace in guest `dmesg` during port detection. There are some more details in the commit messages.

I also updated the integration test goal post.

### DIFF
--- a/nix/tests.nix
+++ b/nix/tests.nix
@@ -93,7 +93,8 @@ in
       machine.wait_for_unit("cloud-hypervisor.service")
 
       # Check whether the USB controller pops up.
-      machine.wait_until_succeeds("grep -Fq 'pci 0000:00:02.0: [1b36:000d]' ${cloudHypervisorLog}")
+      machine.wait_until_succeeds("grep -Fq 'usb usb1: Product: xHCI Host Controller' ${cloudHypervisorLog}")
+      machine.wait_until_succeeds("grep -Fq 'hub 1-0:1.0: 1 port detected' ${cloudHypervisorLog}")
     '';
   };
 }
--- a/nix/tests.nix
+++ b/nix/tests.nix
@@ -87,13 +87,15 @@ in
       };
     };
 
+    # The nested CI runs are really slow.
+    globalTimeout = 3600;
     testScript = ''
       start_all()
 
       machine.wait_for_unit("cloud-hypervisor.service")
 
       # Check whether the USB controller pops up.
-      machine.wait_until_succeeds("grep -Fq 'usb usb1: Product: xHCI Host Controller' ${cloudHypervisorLog}")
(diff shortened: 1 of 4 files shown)


snue:
The test fails with timeout. But it succeeds locally, building it in the sandbox or just running the `.driver`. Not sure what the issue is on the CI runner.

blitz:
> The test fails with timeout. But it succeeds locally, building it in the sandbox or just running the `.driver`. Not sure what the issue is on the CI runner.

The Github runner uses nested-nested virtualization and is _very_ slow. So I think it just runs into the timeout. You can try increasing the timeout. This is not a good solution going forward though.

@snue Can you check with @xanderio to see whether can enable a Hercules CI runner for the Cyberus org? I can also look into it tomorrow. It should speed up things considerably.

blitz (approved):

snue:
> The Github runner uses nested-nested virtualization and is _very_ slow. So I think it just runs into the timeout. You can try increasing the timeout. This is not a good solution going forward though.

Increasing the timeout fixed it indeed. It is very slow.

> @snue Can you check with @xanderio to see whether can enable a Hercules CI runner for the Cyberus org? I can also look into it tomorrow. It should speed up things considerably.

I reached out. But you can follow up on that.

PR #18, opened 2025-05-23
ci: disable Github Actions in favor of Hercules CI

blitz:
Due to running in a VM, the Github Actions CI was incredibly slow. We are seeing almost half an our of CI time. Use Hercules CI to speed things up. There our integration tests don't run in doubly-nested VMs.

**Important:** Our Hercules CI agent will **not** run in forked repositories. This would be a security issue. For blitz/usbvfiod it works because I have my own Hercules CI agent. In general, you need (at least) one agent per Github account.

See the dashboard for my fork here: https://hercules-ci.com/github/blitz/usbvfiod The commit status seems to be only annotated in my fork: https://github.com/blitz/usbvfiod/tree/hercules-ci

This means that:

1. either we set up a Hercules CI agent for each of our developers,
2. we host our development branches in this cyberus-technology/usbvfiod.

2 might be easier, but we have to keep the branch chaos at bay. Maybe we can do this with branch rules so we can limit the creation of branches to `$developer/$feature` or at least to `dev/$feature`.

@snue What do you think?

PS. While I was here I updated our dependencies and disabled gitlint (see commit message).


flake.nix
```diff
@@ -74,7 +74,6 @@
               typos.enable = true;
               deadnix.enable = true;
               statix.enable = true;
-              gitlint.enable = true;
```
snue, line 77: yes, that is just annoying currently :disappointed: 

How would you resolve the automated flake lock file commit if you couldn't just disable it? Can you pass `--no-verify` transparently?

snue (commented):
I think option 2 is easy and sufficient for now. And hopefully it enables better integration with the pull request UI. We can enforce branch hygiene with rules. We should definitely follow them. I have no issues with a `$user/$branch` naming scheme.

I see new clippy issues on your CI: https://hercules-ci.com/accounts/github/blitz/derivations/%2Fnix%2Fstore%2Fszib5sxrf90pz6vi538svb2qambrhg3a-usbvfiod-clippy-0.1.0.drv/log?via-job=0e5a1523-a8f6-4741-89c8-607210679cef

blitz:
The Hercules runner works: https://hercules-ci.com/github/cyberus-technology/usbvfiod/jobs/2

PR #19, opened 2025-05-27
Fix race condition in smoke-test

lbeierlieb:
On my system, the smoke-test consistently (five attempts) failed, because the systemd cloud-hypervisor.service failed ("Failed to open hypervisor interface"). Manually restarting the service in an interactive test run worked, so we assume there is a race condition between the loading of KVM and the execution of the cloud-hypervisor.service.
This PR ensures that KVM is loaded during the second stage of the boot process and is thus reliably available for the start of cloud-hypervisor.service.

For reference, the log of the failed cloud-hypervisor.service:
image
The image also shows that kvm was available during interactive use but apparently not when the cloud-hypervisor.service launched.

### DIFF
--- a/nix/tests.nix
+++ b/nix/tests.nix
@@ -55,6 +55,7 @@ in
     name = "usbvfiod Smoke Test";
 
     nodes.machine = { pkgs, ... }: {
+      boot.kernelModules = [ "kvm" ];
       systemd.services.usbvfiod = {
         wantedBy = [ "multi-user.target" ];
 


nix/tests.nix
```diff
@@ -55,6 +55,7 @@ in
     name = "usbvfiod Smoke Test";
 
     nodes.machine = { pkgs, ... }: {
+      boot.kernelModules = [ "kvm-amd" "kvm-intel" ];
```
snue, line 58: Can we just call for `kvm` and leave out the architecture? I would hope `depmod` solves that for us.

blitz (approved):

nix/tests.nix
lbeierlieb: We took `[ "kvm-amd" "kvm-intel" ]` from one @blitz 's configurations; I agree that loading the conflicting modules looks weird. Unfortunately, I am not able to reproduce the error today... let's load `kvm` and see if the issues arises some time again in the future.

PR #21, opened 2025-05-27
Wire up interrupts to the guest

snue:
This provides a simple `InterruptLine` trait and an implementation for the `EventFd` based interrupts of `vfio-user`.

I also fixed the announced interrupt support. It is sufficient to only provide the correct number of MSI-X vectors (currently 1).

I added a user service to dump diagnostic information from the guest to the cloud hypervisor log. Doing it as a user unit that starts on the automatic login has less potential for ugly interleaved messages. It still races with printing the shell prompt, but the interleaving looks safe.

This PR also injects a bogus interrupt when the R/S bit is toggled active. This isn't compliant with the spec, but I couldn't spot the driver complaining anywhere. The interrupt is registered in the guest, which the smoke test now checks for.

### DIFF
--- a/src/device/pci/xhci.rs
+++ b/src/device/pci/xhci.rs
@@ -167,6 +167,12 @@ impl XhciController {
         self.running = usbcmd & 0x1 == 0x1;
         if self.running {
             debug!("controller started with cmd {usbcmd:#x}");
+
+            // XXX: This is just a test to see if we can generate interrupts.
+            // This will be removed once we generate interrupts in the right
+            // place, (e.g. generate a Port Connect Status Event) and test it.
+            self.interrupt_line.interrupt();
+            debug!("signalled a bogus interrupt");
         } else {
             debug!("controller stopped with cmd {usbcmd:#x}");
         }
--- a/nix/tests.nix
+++ b/nix/tests.nix
@@ -25,6 +25,23 @@ let
         # Convenience packages for interactive use
         environment.systemPackages = [ pkgs.pciutils pkgs.usbutils ];
 
+        # Add user services that run on automatic login.
+        systemd.user.services = {
+          dump-diagnostics = {
+            description = "Dump diagnostic information";
+            wantedBy = [ "default.target" ];
+
+            serviceConfig = {
+              ExecStart = pkgs.writeShellScript "diagnostic-dump" ''
+                echo Dumping Diagnostics
+                cat /proc/interrupts
+              '';
+              StandardOutput = "journal+console";
+              StandardError = "journal+console";
+            };
+          };
+        };
+
         # Silence the useless stateVersion warning. We have no state to keep.
         system.stateVersion = config.system.nixos.release;
       })
@@ -98,6 +115,9 @@ in
       # Check whether the USB controller pops up.
       machine.wait_until_succeeds("grep -Fq 'usb usb1: Product: xHCI Host Controller' ${cloudHypervisorLog}", timeout=3000)
       machine.wait_until_succeeds("grep -Fq 'hub 1-0:1.0: 1 port detected' ${cloudHypervisorLog}")
+
+      # Read the diagnostic information after login.
+      machine.wait_until_succeeds("grep -Eq '\s+1\s+PCI-MSIX.*xhci_hcd' ${cloudHypervisorLog}")
     '';
   };
 }


src/device/pci/xhci.rs
```diff
@@ -19,6 +19,8 @@ use crate::device::{
     },
 };
 
+use super::traits::InterruptLine;
```
snue, line 22: Should probably be qualified as `crate::` instead?

src/xhci_backend.rs
```diff
@@ -256,6 +278,28 @@ impl ServerBackend for XhciBackend {
...
+            })
+            .collect();
+
+        self.controller.lock().unwrap().set_irqs(
+            self.interrupts
+                .clone()
+                .into_iter()
+                .map(|irq| irq.clone() as _)
+                .collect(),
+        );
```
snue, line 302: I had a lot of issues with `: Sized` requirements on the `dyn` type. There is probably a shorter way to construct the vector for the controller that also doesn't require storing `self.interrupts` at all.

src/device/pci/traits.rs
```diff
@@ -70,3 +70,24 @@ pub trait PciDevice: Debug {
...
+}
+
+/// A dummy interrupt line that is intended to be used by devices whose interrupts aren't wired to
+/// anything.
+#[derive(Default, Debug, Clone, Copy)]
+pub struct DummyInterruptLine {}
+
+impl InterruptLine for DummyInterruptLine {
+    fn interrupt(&self) {}
+}
```
snue, line 93: Full disclosure: this code is from the unpublished Particle project, as the rest of PCI code. Should credit @phip1611 and @kaldonir in the commit.

blitz: Is there anything that speaks against importing `interrupt_line.rs` as is at `src/device/interrupt_line.rs`?

src/device/pci/xhci.rs
```diff
@@ -63,6 +65,9 @@ pub struct XhciController {
 
     /// The minimum interval in 250ns increments between interrupts.
     interrupt_moderation_interval: u64,
+
+    /// The interrupt line triggered to signal device events.
+    interrupt_line: Option<Arc<dyn InterruptLine>>,
```
blitz, line 70: This should not be optional. If the caller is not interested in the interrupt, they can pass an unconnected (dummy) interrupt line. Then the rest of the code doesn't have to handle the special case.

src/device/pci/xhci.rs
```diff
@@ -92,7 +97,21 @@ impl XhciController {
...
             interrupt_management: 0,
             interrupt_moderation_interval: runtime::IMOD_DEFAULT,
+            interrupt_line: None,
+        }
+    }
+
+    /// Configure the interrupt line(s) for the controller.
+    ///
+    /// The [`XhciController`] uses these to issue interrupts for events.
+    pub fn set_irqs(&mut self, irqs: Vec<Arc<dyn InterruptLine>>) {
```
blitz, line 107: I find this interface weird. If our controller only supports a single interrupt why can we pass a vector here?

If you always want the caller to pass the right number of interrupt lines, you can make this a `pub fn connect_irqs(&mut self, irqs: &[Arc<dyn InterruptLine>; MAX_INTRS])`. It's a bit more awkward for the caller, but you can't hand in anything wrong.

At the moment, to keep it simple I'm also ok with `pub fn connect_irq(&mut self, irq: Arc<dyn InterruptLine>)`.

That being said, **if it's possible**, can we get rid of multi-stage construction and just hand the interrupt lines in via `new`?

src/device/pci/xhci.rs
blitz: The `super` way is nice, because it will not change if we reorganize the code. I don't have a strong opinion here, though.

src/device/pci/traits.rs
snue: I found it more convenient to integrate it in the common "traits" module. Splitting it out seemed a bit cluttered for the minimum functionality. But this is a matter of taste only.

src/device/pci/xhci.rs
snue: Indeed, this has the same issues as the API on the vfio_user crate, which expects a `Vec` of exactly `VFIO_PCI_NUM_IRQS` elements, but does not enforce it. Here it's even weirder because we only expect the single interrupt.

I don't see a requirement to support more in the near future at least.

This leaks the value of the constant into the API (and requires a cast to `usize`):
`pub fn connect_irqs(&mut self, irqs: &[Arc<dyn InterruptLine>; MAX_INTRS as usize])`
It works but I'm not a huge fan of that, either. Maybe we go with the simple single IRQ interface for now.

snue:
@blitz I can rebase this if you want to merge #22 first.

blitz (approved):

PR #22, opened 2025-05-28
Library Removal plus Fallout

blitz:
So far we had a `[bin]` target and a `[lib]` target in our Cargo.toml (implicitly via the presence of `main.rs` and `lib.rs`). There was no good rule of thumb which code goes where, so I folded everything into the `[bin]` target for now.

This creates a lot of unused code warnings. I don't want to remove all of the unused code, because some might come in handy. I've removed some that seemed very pointless.

There is a tiny functional change in this PR:

- We now tell the vfio-user client about our BARs.

Please review commit-by-commit.


src/xhci_backend.rs
```diff
@@ -82,20 +84,57 @@ impl XhciBackend {
...
-                    argsz: size_of::<vfio_region_info>() as u32,
-                    index: i,
-                    ..Default::default()
-                },
+                    VFIO_PCI_BAR0_REGION_INDEX
+                    | VFIO_PCI_BAR1_REGION_INDEX
+                    | VFIO_PCI_BAR2_REGION_INDEX
+                    | VFIO_PCI_BAR3_REGION_INDEX
+                    | VFIO_PCI_BAR4_REGION_INDEX
+                    | VFIO_PCI_BAR5_REGION_INDEX => {
```
snue, line 112: Thank you! Did you notice any misbehavior without it? The bars showed up fine in the guest and also worked, apparently.

blitz:
> LGTM. Thanks for the cleanup. Do you plan to squash the fixup commit before merge? Does GitHub do it automatically?

I'll fix it.

PR #28, opened 2025-06-04
Add a virtual usb device to test

snue:
Configure the NixOS smoke integration test with a virtual XHCI controller and virtual USB storage device.
We will pass this storage device to the guest through the `usbvfiod` service eventually. For now, it is not used by the guest.

I added a check that we can read the device contents in the host at least (so it works, before we pass it to the guest).

While here, this also adds the `dyndbg` messages for the guest xhci driver.

### DIFF
--- a/nix/tests.nix
+++ b/nix/tests.nix
@@ -69,6 +69,11 @@ let
   # well.
   usbvfiodSocket = "/tmp/usbvfio";
   cloudHypervisorLog = "/tmp/chv.log";
+
+  # Provide a raw file as usb stick test image.
+  usbDiskImage = pkgs.writeText "usb-stick-image.raw" ''
+    This is an uninitialized drive.
+  '';
 in
 {
   integration-smoke = pkgs.nixosTest {
@@ -105,6 +110,13 @@ in
       virtualisation = {
         cores = 2;
         memorySize = 4096;
+        qemu.options = [
+          # A virtual USB XHCI controller in the host ...
+          "-device qemu-xhci,id=host-xhci,addr=10"
+          # ... with an attached usb stick.
+          "-drive if=none,id=usbstick,format=raw,snapshot=on,file=${usbDiskImage}"
+          "-device usb-storage,bus=host-xhci.0,drive=usbstick"
+        ];
       };
     };
 
@@ -121,6 +133,9 @@ in
 
       # Read the diagnostic information after login.
       machine.wait_until_succeeds("grep -Eq '\s+1\s+PCI-MSIX.*xhci_hcd' ${cloudHypervisorLog}")
+
+      # Check whether the virtual usb stick is available in the host.
+      machine.succeed('grep -Fq "uninitialized drive" /dev/sda')
     '';
   };
 }


nix/tests.nix
```diff
@@ -17,6 +17,9 @@ let
           # The virtio-console is an option as well, but is not
           # compiled into the NixOS kernel and would be inconvenient.
           "console=ttyS0"
+          # Enable dyndbg messages for the XHCI driver.
+          "xhci_pci.dyndbg==pmfl"
+          "xhci_hcd.dyndbg==pmfl"
```
snue, line 22: Well, I see the options on the guest kernel `cmdline`, but I am not seeing the expected messages in the `/tmp/chv.log`. :crying_cat_face:

nix/tests.nix
```diff
@@ -102,12 +112,22 @@ in
       virtualisation = {
         cores = 2;
         memorySize = 4096;
+        qemu.options = [
+          # A virtual USB XHCI controller in the host ...
+          "-device qemu-xhci,id=host-xhci,addr=10"
+          # ... with an attached usb stick.
+          "-drive if=none,id=usbstick,format=raw,file=${usbDiskImagePath}"
```
blitz, line 119:
```suggestion
          "-drive if=none,id=usbstick,format=raw,snapshot=on,file=${usbDiskImage}"
```

Does this work to avoid the writable image?

nix/tests.nix
blitz: From the [docs](https://docs.kernel.org/admin-guide/dynamic-debug-howto.html), I would have expected something like `xhci_pci.dyndbg` to just do the right thing. Let me try...

snue: It works for me on local tests (running usbvfiod and cloud-hypervisor). Just in the nixosTest I am missing the output.

nix/tests.nix
blitz: This would also unbreak the interactive driver:

```
In [1]: start_all()
start all VMs
machine: starting vm
mke2fs 1.47.2 (1-Jan-2025)
qemu-system-x86_64: -drive if=none,id=usbstick,format=raw,file=/tmp/usb-stick-image: Could not open '/tmp/usb-stick-image': No such file or directory
---------------------------------------------------------------------------
ConnectionResetError                      Traceback (most recent call last)

```

Or at least it would Just Work™ without having to remember to run some magic commands.

snue: Yes! That gets rid of the permission denied errors. I'll use that instead and remove the extra tmpfile handling. Thank you!

nix/tests.nix
blitz: This works for me. Not sure why `consoleLogLevel` needs to be changed. According to the docs, setting it to debug should suffice.

```nix
        # Enable debug logging for the XHCI controller.
        boot.extraModprobeConfig = ''
          options xhci_pci dyndbg=+pt
          options xhci_hcd dyndbg=+pt
        '';

        # Print everything.
        boot.consoleLogLevel = 15;
```

snue: Well, it was off by one. We can either supply `ignore_loglevel` to just print everything or `loglevel=8` to print everything "smaller than 8".

PR #29, opened 2025-06-05
Add abstractions for the event ring

lbeierlieb:
This PR adds abstractions for event ring handling itself as well as the Transfer Request Blocks (TRBs) that the XHCI controller can place on the event ring.

# TRBs

The controller requires the possibility to construct representations of the events it wants to signal to the driver (ideally checked by the type system as much as possible, further check with assertions to detect implausible values), and a way to retrieve the 16-byte representation it has to write on the event ring.

Individual structs encode the event-specific information (currently `CommandCompletionEventTrbData` and `PortStatusChangeEventTrbData` are implemented); the enum `EventTrb` aggregates the different events. Each event requires the implementation of a matching, publicly available constructor. Further, each event struct needs to implement `to_bytes`, which the public `EventTrb::to_bytes` can call.

The cycle bit is not part of the TRB representation and only set by `EventTrb::to_bytes`. The caller has to make sure the specified cycle bit value is appropriate with the current state of the event ring, where the TRB will be placed.

# Event Ring
The new Event Ring implementation offers methods for initial configuration and for runtime use (read/write values of MMIO registers, enqueue event TRBs).

We remove the old minimal internal implementation of the event ring in the XHCI controller and replace it with the new abstraction. Additionally, an event gets enqueued for testing right after the controller starts and before a test interrupt is fired.

## Limitations
The ring cannot yet wrap around. The Linux driver supplies us with a single segment for 256 TRBs. The `enqueue` method works well 256 times, after that we crash.
Despite our choice to support only a single segment for the event ring, the wrap-around logic is not trivial.
In my opinion, the current implementation should allow us to get far enough to fully configure USB devices, which we should focus on before refining the event ring implementation.

### DIFF
--- a/src/device/pci/rings.rs
+++ b/src/device/pci/rings.rs
@@ -1,0 +1,138 @@
+//! Abstractions of the rings (Event Ring, Command Ring, Transfer Ring) of a
+//! USB3 Host (XHCI) controller.
+//!
+//! The specification is available
+//! [here](https://www.intel.com/content/dam/www/public/us/en/documents/technical-specifications/extensible-host-controler-interface-usb-xhci.pdf).
+
+use tracing::debug;
+
+use crate::device::bus::{BusDeviceRef, Request, RequestSize};
+
+use super::trb::EventTrb;
+
+/// The Event Ring: A unidirectional means of communication, allowing the XHCI
+/// controller to send events to the driver.
+///
+/// This implementation is a simplified version of the full mechanism specified
+/// in the XHCI specification. We assume that the Event Ring Segment Table only
+/// holds a single segment.
+#[derive(Debug, Default, Clone)]
+pub struct EventRing {
+    /// The address of the Event Ring Segment Table.
+    /// This field directly corresponds with the ERSTBA register(s) in the
+    /// XHCI's MMIO region.
+    base_address: u64,
+    /// The dequeue pointer as the driver reports it.
+    /// This field directly corresponds with the ERDP register(s) in the
+    /// XHCI's MMIO region.
+    /// When the ring is not empty, the pointer indicates the address of the
+    /// last processed TRB.
+    /// When the ring is empty, the pointer is equal to the enqueue pointer
+    /// (EREP).
+    dequeue_pointer: u64,
+    /// no public access, indicated via cycle bit
+    /// The enqueue pointer; the spec refers to it as the EREP.
+    /// It is an internal variable of the XHCI controller.
+    /// The driver implicitly knows it reached the enqueue pointer (and thus
+    /// can conclude the ring is empty), when it detects a cycle-bit mismatch
+    /// at ERDP.
+    enqueue_pointer: u64,
+    /// The number of TRBs that fits into the current segment.
+    /// The count is initialized from the size field of an Event Ring Segment
+    /// Table Entry. Once the count reaches 0, we have to advance to the next
+    /// segment—because we only support one, we move back to the start of the
+    /// same segment.
+    trb_count: u32,
+    /// The producer cycle state.
+    /// The driver tracks cycle state as well and can deduce the enqueue
+    /// pointer by detecting cycle-state mismatches.
+    /// Initially, the state has to be true (corresponds to TRB cycle bits
+    /// equal to 1), so new TRBs can be written over the zero-initialized
+    /// memory. Later, the cycle_state has to flip after every full pass of the
+    /// event ring (i.e., in our case, when we move from the back of the
+    /// segment to the front of the single segment).
+    cycle_state: bool,
+}
+
+impl EventRing {
+    /// Configure the Event Ring.
+    /// Amongst setting the base address of the Event Ring Segment Table,
+    ///
+    /// # Parameters
+    ///
+    /// - `erstba`: base address of the Event Ring Segment Table
+    /// - `dma_bus`: the bus to use for DMA accesses
+    pub fn configure(&mut self, erstba: u64, dma_bus: BusDeviceRef) {
+        assert_eq!(erstba & 0x3f, 0, "unaligned event ring base address");
+
+        self.base_address = erstba;
+        self.enqueue_pointer = dma_bus.read(Request::new(erstba, RequestSize::Size8));
+        self.trb_count = dma_bus.read(Request::new(erstba + 8, RequestSize::Size4)) as u32;
+        self.cycle_state = true;
+
+        debug!("event ring segment table is at {:#x}", erstba);
+        debug!(
+            "initializing event ring enqueue pointer with base address of the first (and only) segment: {:#x}",
+            self.enqueue_pointer
+        );
+        debug!(
+            "retrieving TRB count of the first (and only) event ring segment from the segment table: {}",
+            self.trb_count
+        );
+    }
+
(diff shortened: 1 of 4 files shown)


src/device/pci/rings.rs
```diff
@@ -0,0 +1,138 @@
...
+
+impl EventRing {
+    /// Configure the Event Ring.
+    /// Amongst setting the base address of the Event Ring Segment Table,
+    ///
+    /// # Parameters
+    ///
+    /// - `erstba`: base address of the Event Ring Segment Table
+    /// - `dma_bus`: the bus to use for DMA accesses
+    pub fn configure(&mut self, erstba: u64, dma_bus: BusDeviceRef) {
```
snue, line 65: You pass the `dma_bus` to all the methods that need to access the guest memory. That makes it easier to construct the type (derive Default). But I don't think the bus ref should ever change between calls and is basically glued to the ring. Were there other considerations to this?

src/device/pci/xhci.rs
```diff
@@ -148,6 +149,12 @@ impl XhciController {
         if self.running {
             debug!("controller started with cmd {usbcmd:#x}");
 
+            // XXX: This is just a test to see if we can enqueue events.
+            // This will be removed once we send events in the right place
+            // (e.g., command completion events on doorbell writes)
+            let trb = EventTrb::new_port_status_change_event_trb(0);
+            self.event_ring.enqueue(&trb, self.dma_bus.clone());
```
snue, line 156: In case there is a device to expose, we should actually generate this kind of event here, right? It's not really misplaced.

The smoke test currently does not see the difference whether this is here or not. But it should visibly change the progress in the debug logs.

snue (approved):
I left remarks, some nits and typos. Nothing that strictly needs to change before a merge.

I mostly wonder if we really need to pass the `bus_ref` all the time. But any refactoring can be done later as well. The ring implementation will see more work.

src/device/pci/trb.rs
```diff
@@ -0,0 +1,183 @@
...
+            command_trb_pointer,
+            command_completion_parameter,
+            completion_code,
+            slot_id,
+        })
+    }
+}
+
+impl CommandCompletionEventTrbData {
+    fn to_bytes(&self) -> [u8; 16] {
```
blitz, line 94: Can you add a happy path unit test? Doesn't have to be exhaustive.

src/device/pci/rings.rs
```diff
@@ -0,0 +1,138 @@
...
+    /// last processed TRB.
+    /// When the ring is empty, the pointer is equal to the enqueue pointer
+    /// (EREP).
+    dequeue_pointer: u64,
+    /// no public access, indicated via cycle bit
+    /// The enqueue pointer; the spec refers to it as the EREP.
+    /// It is an internal variable of the XHCI controller.
+    /// The driver implicitly knows it reached the enqueue pointer (and thus
+    /// can conclude the ring is empty), when it detects a cycle-bit mismatch
+    /// at ERDP.
```
blitz, line 38: The formatting of these documentation paragraphs really throws my reading off. What about:

```
/// The event ring enqueue pointer (EREP).
///
/// EREP is internal state of the event ring. The driver implicitly
/// knows it reached the enqueue pointer (and thus can conclude the
/// ring is empty), when it detects a cycle-bit mismatch at ERDP.
```

There are established best practices regarding writing documentation in Rust: https://doc.rust-lang.org/rustdoc/how-to-write-documentation.html

We should follow it to make our codebase more accessible!

src/device/pci/rings.rs
```diff
@@ -0,0 +1,138 @@
...
+    /// equal to 1), so new TRBs can be written over the zero-initialized
+    /// memory. Later, the cycle_state has to flip after every full pass of the
+    /// event ring (i.e., in our case, when we move from the back of the
+    /// segment to the front of the single segment).
+    cycle_state: bool,
+}
+
+impl EventRing {
+    /// Configure the Event Ring.
+    /// Amongst setting the base address of the Event Ring Segment Table,
```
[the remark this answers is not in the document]
lbeierlieb: Thanks! I added the forgotten description.

src/device/pci/rings.rs
```diff
@@ -0,0 +1,138 @@
...
+            "retrieving TRB count of the first (and only) event ring segment from the segment table: {}",
+            self.trb_count
+        );
+    }
+
+    /// Handle writes to the Event Ring Dequeue Pointer (ERDP).
+    ///
+    /// # Parameters
+    ///
+    /// - `erdp`: value that the driven has written to the ERDP register.
```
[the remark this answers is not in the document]
lbeierlieb: fixed

src/device/pci/rings.rs
lbeierlieb: I wanted to explicitly see in the `XhciController` when DMA accesses occur. When the `EventRing` keeps its own reference, it becomes less obvious when guest memory is accessed.

I am not particular about the explicitness and can see the benefit in reducing the boilerplate for passing the `dma_bus` every time. I am open to redesign.

blitz: Should be easy to refactor later, so we just see how it goes.

Btw, if you don't want to keep ownership of the bus device, you can just accept a reference to `impl BusDevice` (or `dyn BusDevice`). Otherwise the caller always has to clone the Arc, which is inconvenient.

src/device/pci/rings.rs
```diff
@@ -0,0 +1,153 @@
...
+        }
+
+        dma_bus.write_bulk(self.enqueue_pointer, &trb.to_bytes(self.cycle_state));
+
+        let enqueue_address = self.enqueue_pointer;
+
+        self.enqueue_pointer += 16;
+        self.trb_count -= 1;
+
+        debug!(
```
blitz, line 141: nit: This seems more a `trace!`. I usually apply the rule "if the VM has the possibility to become unusably slow with these messages, use trace".

blitz (approved):
@lbeierlieb Feel free to merge whenever you want!

src/device/pci/rings.rs
lbeierlieb: I like the approach with the reference; unfortunately, I don't see a way to then directly call the `configure` function in `PciDevice::write_io`, because I need a mutable reference to the xhci's `event_ring` field and a reference to the `dma_bus` field. Adding another wrapper method in the XhciController can resolve this borrowing issue, but I'll just leave the cloning approach for now.

src/device/pci/trb.rs
```diff
@@ -0,0 +1,222 @@
...
+//! [here](https://www.intel.com/content/dam/www/public/us/en/documents/technical-specifications/extensible-host-controler-interface-usb-xhci.pdf).
+
+use super::constants::xhci::rings::trb_types::*;
+
+/// Represents a TRB that the XHCI controller can place on the event ring.
+#[derive(Debug)]
+pub enum EventTrb {
+    //TransferEvent,
+    CommandCompletionEvent(CommandCompletionEventTrbData),
+    PortStatusChangeEvent(PortStatusChangeEventData),
```
lbeierlieb, line 13: I noticed that I introduced inconsistent naming in the event-specific structs; I will all terminate all identifiers with `TrbData`.

PR #35, opened 2025-06-17
smoke-test: add USB device report for integration-smoke

ziyifu225:
This PR adds functionality to the integration-smoke test in `test.nix` that generates a report from the guest side.

The goal is to check USB device accessibility and permission information after boot.
Currently, this is implemented for **root login only**.

### What’s included:

1. I added `usbutils` on `nodes.machine` to make sure I can get USB device information from machine easily.
2. write a check to make sure `lsusb` can be used.
3. Dynamically detects the USB block device by scanning `/dev/sd*` entries using `udevadm`, instead of assuming a fixed path like /`dev/sda`, to account for cases where local disks occupy lower device names.
4. Extracts the ID_VENDOR_ID and ID_MODEL_ID of the detected USB block device via `udevadm`, enabling identification of the corresponding raw USB character device listed in `lsusb`.
5. Dump of current user identity via `whoami `and `id`.
6. get the permissions of these two USB devices by using `ls -l`.
7. All output is written to `/tmp/test_report.txt`.

### DIFF
--- a/nix/tests.nix
+++ b/nix/tests.nix
@@ -77,32 +77,25 @@ let
 
   # Report for target device path and access permissions
   usbDeviceInfoScript = pkgs.writeShellScript "usb-device-info" ''
-    {
-      echo "----- USB Block Device (/dev/sdX) -----"
-      if [ -f /tmp/usb_device ]; then
-        dev=$(cat /tmp/usb_device)
-        echo "Device path: $dev"
-        echo "--- USB Block Device (/dev/sdX) Permissions ---"
-        ls -l $dev
-      else
-        echo "No USB device found!" >&2
-        exit 1
-      fi
+    echo "----- USB Block Device (/dev/sdX) -----"
+    usb_device=$(lsblk -SJ | jq -r '.blockdevices[] | select(.tran == "usb") | .name')
+    dev="/dev/$usb_device"
+    echo "Device path: $dev"
+    echo "--- USB Block Device (/dev/sdX) Permissions ---"
+    ls -l $dev || echo "Could not stat block device $dev"
 
-      echo "----- USB Bus-Exposed Device -----"
-      if [ -f /tmp/bus_usb_device ]; then
-        usb_dev=$(cat /tmp/bus_usb_device)
-        echo "$usb_dev"
-        bus=$(echo "$usb_dev" | awk '{print $2}')
-        dev_num=$(awk '{ gsub(":",""); print $4 }' <<<"$usb_dev")
-        path="/dev/bus/usb/$bus/$dev_num"
-        echo "Character device path: $path"
-        echo "--- USB Bus-Exposed Device Permissions ---"
-        ls -l "$path"
-      else
-        echo "Character device path: <not found>"
-      fi
-    } &> /tmp/test_report.txt
+    echo "----- USB Bus-Exposed Device -----"
+    vendor=$(udevadm info --query=all --name=$dev | grep -oP 'ID_USB_VENDOR_ID=\K\w+' || true)
+    model=$(udevadm info --query=all --name=$dev | grep -oP 'ID_USB_MODEL_ID=\K\w+' || true)
+    bus_usb_device=$(lsusb -d "$vendor:$model" || true)
+    echo "$bus_usb_device"
+    bus=$(awk '{print $2}' <<<"$bus_usb_device")
+    dev_num=$(awk '{ gsub(":",""); print $4 }' <<<"$bus_usb_device")
+    path="/dev/bus/usb/$bus/$dev_num"
+    echo "Character device path: $path"
+    ls -l "$path" || echo "Could not stat character device $path"
+
+    exit 0
   '';
 in
 {
@@ -169,25 +162,9 @@ in
       # Read the diagnostic information after login.
       machine.wait_until_succeeds("grep -Eq '\s+1\s+PCI-MSIX.*xhci_hcd' ${cloudHypervisorLog}")
 
-      # find the USB Device
-      machine.execute("""
-      # get USB Block Device
-        usb_device=$(lsblk -SJ | jq -r '.blockdevices[] | select(.tran == "usb") | .name')
-        dev="/dev/$usb_device"
-        echo "$dev" > /tmp/usb_device
-
-        # get USB Bus-Exposed Device
-        vendor=$(udevadm info --query=all --name=$dev | grep -oP 'ID_USB_VENDOR_ID=\\K\\w+')
-        model=$(udevadm info --query=all --name=$dev | grep -oP 'ID_USB_MODEL_ID=\\K\\w+')
-        bus_usb_device=$(lsusb -d "$vendor:$model")
-        echo "$bus_usb_device" > /tmp/bus_usb_device
-      """)
-
       # Display device path and access permissions
-      machine.succeed("${usbDeviceInfoScript}")
-      
       print("-------- USB Device Information Report --------")
-      stdout = machine.execute("cat /tmp/test_report.txt")[1]
+      stdout = machine.execute("${usbDeviceInfoScript}")[1]
       print(stdout)
     '';
   };


nix/tests.nix
```diff
@@ -136,6 +140,83 @@ in
 
       # Check whether the virtual usb stick is available in the host.
       machine.succeed('grep -Fq "uninitialized drive" /dev/sda')
```
lbeierlieb, line 142: You mention that you check all block devices in case other disk occupy lower /dev/sd* names. This check assumes that the virtual USB stick will be available as `/dev/sda`, so there exists a bit of inconsistency in the assumptions of which devices might be available.

I suggest to either remove this check (you check later that a USB-backed drive was found), or to make use of the assertion here and just work with `/dev/sda`.

lbeierlieb (commented):
Nice work, this PR is a really well prepared first contribution!

I left some suggestions for minor improvements. Feel free to either implement the suggestion or argue against the changes.

nix/tests.nix
```diff
@@ -136,6 +140,83 @@ in
...
+          fi
+        done
+      """)
+
+      # show the information about path and permission
+      machine.succeed("""
+        {
+          echo "----- Current Identity -----"
+          whoami
+          id
```
blitz, line 169:
```suggestion
```

I suggest to remove this, because `machine.succeed` implies running the commands as root.

nix/tests.nix
```diff
@@ -136,6 +140,83 @@ in
...
+
+            # get Vendor/Model ID
+            vendor=$(udevadm info --query=all --name=$dev | grep -oP 'ID_USB_VENDOR_ID=\\K\\w+')
+            model=$(udevadm info --query=all --name=$dev | grep -oP 'ID_USB_MODEL_ID=\\K\\w+')
+
+            echo "$vendor" > /tmp/usb_vendor
+            echo "$model" > /tmp/usb_model
+            break
+          fi
+        done
```
blitz, line 161: Would `lsusb` do as well? It prints vendor/model ID for every device. You can also query inidividual devices with `lsusb -D /dev/bus/usb/001/001`.

nix/tests.nix
```diff
@@ -136,6 +140,83 @@ in
 
       # Check whether the virtual usb stick is available in the host.
       machine.succeed('grep -Fq "uninitialized drive" /dev/sda')
+
+      # Check whether lsusb can be used
+      machine.succeed("lsusb | grep -q 'ID'")
+
+      # find the USB Device
+      machine.succeed("""
```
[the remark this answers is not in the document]
ziyifu225: Thanks for the suggestion. I’ve changed both machine.succeed to machine.execute as you proposed.

nix/tests.nix
```diff
@@ -136,6 +140,83 @@ in
...
+            exit 1
+          fi
+          
+        } > /tmp/test_report.txt
+      """)
+
+      print("-------- Report from Guest --------")
+      stdout = machine.execute("cat /tmp/test_report.txt")[1]
+      print(stdout)
+
```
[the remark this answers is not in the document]
ziyifu225: The unnecessary blank line has been removed.

nix/tests.nix
ziyifu225: Good point — that was from my earlier testing, and I forgot to remove it after updating the code. It’s been deleted now.

nix/tests.nix
ziyifu225: I’ve removed it for now. I’ll add it back later when logging in as a non-root user.

nix/tests.nix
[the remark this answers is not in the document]
lbeierlieb: I only referred to the first usage, which will always succeed. The second `succeed` is vital, especially because you now removed the check for the `/dev/sda` device. The step will fail if the USB device is not present, because you `exit 1` when you the files in `/tmp` do not exist.

Now the smoke-test can pass even without the USB device present—we don't want that.
Please change the call back to `succeed` (currently line 157).

nix/tests.nix
lbeierlieb: The change is fine, I'm just keeping this conversation unresolved until the [other](https://github.com/cyberus-technology/usbvfiod/pull/35#discussion_r2156993986) `succeed` call is restored, covering the functionality of this step.

nix/tests.nix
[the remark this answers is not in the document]
ziyifu225: Got it — seems I misunderstood the difference between succeed and execute. I’ll fix the code.

nix/tests.nix
```diff
@@ -134,8 +139,54 @@ in
...
+        {
+          echo "----- USB Block Device (/dev/sdX) -----"
+          if [ -f /tmp/usb_device ]; then
+            dev=$(cat /tmp/usb_device) 
+            echo "Device path: $dev"
+            echo "--- USB Block Device (/dev/sdX) Permissions ---"
+            ls -l $dev
+          else
+            echo "No USB device found!" >&2
+            echo "No USB device found!"
```
snue, line 167: I think it is sufficient to just print this on `stderr` (remove the second line). I see you want it in the test log as well, but this is missing any other possible error output from called programs currently anyway. You want to redirect both file descriptors (stdout and stderr) to the test report log below instead.

nix/tests.nix
```diff
@@ -134,8 +139,54 @@ in
...
+      machine.execute("""
+      # get USB Block Device
+        usb_device=$(lsblk -SJ | jq -r '.blockdevices[] | select(.tran == "usb") | .name')
+        dev="/dev/$usb_device"
+        echo "$dev" > /tmp/usb_device
+
+        # get USB Bus-Exposed Device
+        vendor=$(udevadm info --query=all --name=$dev | grep -oP 'ID_USB_VENDOR_ID=\\K\\w+')
+        model=$(udevadm info --query=all --name=$dev | grep -oP 'ID_USB_MODEL_ID=\\K\\w+')
+        bus_usb_device=$(lsusb | grep "$vendor:$model")
```
snue, line 152: This does the same:
```suggestion
        bus_usb_device=$(lsusb -d "$vendor:$model")
```

nix/tests.nix
```diff
@@ -134,8 +139,54 @@ in
...
+            bus=$(echo "$usb_dev" | awk '{print $2}')
+            dev_num=$(echo "$usb_dev" | awk '{print $4}' | sed 's/://')
+            path="/dev/bus/usb/$bus/$dev_num"
+            echo "Character device path: $path"
+            echo "--- USB Bus-Exposed Device Permissions ---"
+            ls -l "$path"
+          else
+            echo "Character device path: <not found>"
+          fi        
+        } > /tmp/test_report.txt
```
snue, line 184: here, redirect also stderr:
```suggestion
        } &> /tmp/test_report.txt
```

nix/tests.nix
```diff
@@ -134,8 +139,54 @@ in
...
+            echo "No USB device found!"
+            exit 1
+          fi
+
+          echo "----- USB Bus-Exposed Device -----"
+          if [ -f /tmp/bus_usb_device ]; then
+            usb_dev=$(cat /tmp/bus_usb_device)
+            echo "$usb_dev"
+            bus=$(echo "$usb_dev" | awk '{print $2}')
+            dev_num=$(echo "$usb_dev" | awk '{print $4}' | sed 's/://')
```
snue, line 176: Please try to avoid `awk` AND `sed` in the same command :grin:  ... They are both great on their own. I suggest staying with `awk` and maybe also use a "here-string" instead of echo and pipe for input in this case:
```suggestion
            dev_num=$(awk '{ gsub(":",""); print $4 }' <<<"$usb_dev")
```

snue (approved):
Thanks! I think the logic breaks when we have more than one matching device exposed, but this is much better than a hardcoded path. It should work for quite a while and can stay like this. I left some other minor suggestions you can address. I'll already leave my approval here.

lbeierlieb (approved):

nix/tests.nix
ziyifu225: ✅ Done!

nix/tests.nix
ziyifu225: ✅ Done!

nix/tests.nix
ziyifu225: ✅ Done!

nix/tests.nix
ziyifu225: All requested changes have been made. Thanks!

snue:
We'll have to figure out what to do about @ziyifu225's access to the CI. I'll talk to @blitz.

I ran the smoke test locally and the output is a bit cluttered from the whole script.

```
vm-test-run-usbvfiod-Smoke-Test> machine: must succeed: 
vm-test-run-usbvfiod-Smoke-Test>   {
vm-test-run-usbvfiod-Smoke-Test>     echo "----- USB Block Device (/dev/sdX) -----"
vm-test-run-usbvfiod-Smoke-Test>     if [ -f /tmp/usb_device ]; then
vm-test-run-usbvfiod-Smoke-Test>       dev=$(cat /tmp/usb_device) 
vm-test-run-usbvfiod-Smoke-Test>       echo "Device path: $dev"
vm-test-run-usbvfiod-Smoke-Test>       echo "--- USB Block Device (/dev/sdX) Permissions ---"
vm-test-run-usbvfiod-Smoke-Test>       ls -l $dev
vm-test-run-usbvfiod-Smoke-Test>     else
vm-test-run-usbvfiod-Smoke-Test>       echo "No USB device found!" >&2
vm-test-run-usbvfiod-Smoke-Test>       exit 1
vm-test-run-usbvfiod-Smoke-Test>     fi
vm-test-run-usbvfiod-Smoke-Test> 
vm-test-run-usbvfiod-Smoke-Test>     echo "----- USB Bus-Exposed Device -----"
vm-test-run-usbvfiod-Smoke-Test>     if [ -f /tmp/bus_usb_device ]; then
vm-test-run-usbvfiod-Smoke-Test>       usb_dev=$(cat /tmp/bus_usb_device)
vm-test-run-usbvfiod-Smoke-Test>       echo "$usb_dev"
vm-test-run-usbvfiod-Smoke-Test>       bus=$(echo "$usb_dev" | awk '{print $2}')
vm-test-run-usbvfiod-Smoke-Test>       dev_num=$(awk '{ gsub(":",""); print $4 }' <<<"$usb_dev")
vm-test-run-usbvfiod-Smoke-Test>       path="/dev/bus/usb/$bus/$dev_num"
vm-test-run-usbvfiod-Smoke-Test>       echo "Character device path: $path"
vm-test-run-usbvfiod-Smoke-Test>       echo "--- USB Bus-Exposed Device Permissions ---"
vm-test-run-usbvfiod-Smoke-Test>       ls -l "$path"
vm-test-run-usbvfiod-Smoke-Test>     else
vm-test-run-usbvfiod-Smoke-Test>       echo "Character device path: <not found>"
vm-test-run-usbvfiod-Smoke-Test>     fi        
vm-test-run-usbvfiod-Smoke-Test>   } &> /tmp/test_report.txt
vm-test-run-usbvfiod-Smoke-Test> 
vm-test-run-usbvfiod-Smoke-Test> machine: (finished: must succeed: 
vm-test-run-usbvfiod-Smoke-Test>   {
vm-test-run-usbvfiod-Smoke-Test>     echo "----- USB Block Device (/dev/sdX) -----"
vm-test-run-usbvfiod-Smoke-Test>     if [ -f /tmp/usb_device ]; then
vm-test-run-usbvfiod-Smoke-Test>       dev=$(cat /tmp/usb_device) 
vm-test-run-usbvfiod-Smoke-Test>       echo "Device path: $dev"
vm-test-run-usbvfiod-Smoke-Test>       echo "--- USB Block Device (/dev/sdX) Permissions ---"
vm-test-run-usbvfiod-Smoke-Test>       ls -l $dev
vm-test-run-usbvfiod-Smoke-Test>     else
vm-test-run-usbvfiod-Smoke-Test>       echo "No USB device found!" >&2
vm-test-run-usbvfiod-Smoke-Test>       exit 1
vm-test-run-usbvfiod-Smoke-Test>     fi
vm-test-run-usbvfiod-Smoke-Test> 
vm-test-run-usbvfiod-Smoke-Test>     echo "----- USB Bus-Exposed Device -----"
vm-test-run-usbvfiod-Smoke-Test>     if [ -f /tmp/bus_usb_device ]; then
vm-test-run-usbvfiod-Smoke-Test>       usb_dev=$(cat /tmp/bus_usb_device)
vm-test-run-usbvfiod-Smoke-Test>       echo "$usb_dev"
vm-test-run-usbvfiod-Smoke-Test>       bus=$(echo "$usb_dev" | awk '{print $2}')
vm-test-run-usbvfiod-Smoke-Test>       dev_num=$(awk '{ gsub(":",""); print $4 }' <<<"$usb_dev")
vm-test-run-usbvfiod-Smoke-Test>       path="/dev/bus/usb/$bus/$dev_num"
vm-test-run-usbvfiod-Smoke-Test>       echo "Character device path: $path"
vm-test-run-usbvfiod-Smoke-Test>       echo "--- USB Bus-Exposed Device Permissions ---"
vm-test-run-usbvfiod-Smoke-Test>       ls -l "$path"
vm-test-run-usbvfiod-Smoke-Test>     else
vm-test-run-usbvfiod-Smoke-Test>       echo "Character device path: <not found>"
vm-test-run-usbvfiod-Smoke-Test>     fi        
vm-test-run-usbvfiod-Smoke-Test>   } &> /tmp/test_report.txt
vm-test-run-usbvfiod-Smoke-Test> , in 0.11 seconds)
vm-test-run-usbvfiod-Smoke-Test> -------- USB Device Information Report --------
vm-test-run-usbvfiod-Smoke-Test> ----- USB Block Device (/dev/sdX) -----
vm-test-run-usbvfiod-Smoke-Test> Device path: /dev/sda
vm-test-run-usbvfiod-Smoke-Test> --- USB Block Device (/dev/sdX) Permissions ---
vm-test-run-usbvfiod-Smoke-Test> brw-rw---- 1 root disk 8, 0 Jun 26 17:02 /dev/sda
vm-test-run-usbvfiod-Smoke-Test> ----- USB Bus-Exposed Device -----
vm-test-run-usbvfiod-Smoke-Test> Bus 003 Device 002: ID 46f4:0001 QEMU QEMU USB HARDDRIVE
vm-test-run-usbvfiod-Smoke-Test> Character device path: /dev/bus/usb/003/002
vm-test-run-usbvfiod-Smoke-Test> --- USB Bus-Exposed Device Permissions ---
vm-test-run-usbvfiod-Smoke-Test> crw-rw-r-- 1 root root 189, 257 Jun 26 17:02 /dev/bus/usb/003/002
vm-test-run-usbvfiod-Smoke-Test> 
vm-test-run-usbvfiod-Smoke-Test> (finished: run the VM test script, in 42.01 seconds)
```

Maybe we can build this into a script ([`pkgs.writeShellScript`](https://noogle.dev/f/pkgs/writeShellScript)) instead to hide the details.

snue:
I opened #39 with these changes for CI. I had to rebase again, so GitHub lets me create a PR. A `git fetch --all` followed by `git rebase -i origin/main` (and confirming the list of new commits by writing the file) worked for me.

nix/tests.nix
```diff
@@ -134,8 +169,26 @@ in
...
+        usb_device=$(lsblk -SJ | jq -r '.blockdevices[] | select(.tran == "usb") | .name')
+        dev="/dev/$usb_device"
+        echo "$dev" > /tmp/usb_device
+
+        # get USB Bus-Exposed Device
+        vendor=$(udevadm info --query=all --name=$dev | grep -oP 'ID_USB_VENDOR_ID=\\K\\w+')
+        model=$(udevadm info --query=all --name=$dev | grep -oP 'ID_USB_MODEL_ID=\\K\\w+')
+        bus_usb_device=$(lsusb -d "$vendor:$model")
+        echo "$bus_usb_device" > /tmp/bus_usb_device
+      """)
```
snue, line 184: You could do the same (`writeShellScript`) for this one :wink:

I think the pattern should be to have the script only generate the output (on stdout) and do the redirection to file in the `machine.succeed()` call.

nix/tests.nix
```diff
@@ -134,8 +169,26 @@ in
...
+
+      # Display device path and access permissions
+      machine.succeed("${usbDeviceInfoScript}")
```
snue, line 187: Please make the script only generate output, and do the file redirection in the call. This way, we see where the file is coming from that we read in the next `machine.execute()`:
```suggestion
      # Display device path and access permissions
      machine.succeed("${usbDeviceInfoScript} &> /tmp/test_report.txt")
```

nix/tests.nix
```diff
@@ -74,12 +74,47 @@ let
...
+        bus=$(echo "$usb_dev" | awk '{print $2}')
+        dev_num=$(awk '{ gsub(":",""); print $4 }' <<<"$usb_dev")
+        path="/dev/bus/usb/$bus/$dev_num"
+        echo "Character device path: $path"
+        echo "--- USB Bus-Exposed Device Permissions ---"
+        ls -l "$path"
+      else
+        echo "Character device path: <not found>"
+      fi
+    } &> /tmp/test_report.txt
```
snue, line 105: If you remove the redirection here, you also do not need the additional grouping (`{ ... }`) anymore.

nix/tests.nix
```diff
@@ -134,8 +169,26 @@ in
...
       machine.wait_until_succeeds("grep -Eq '\s+1\s+PCI-MSIX.*xhci_hcd' ${cloudHypervisorLog}")
 
-      # Check whether the virtual usb stick is available in the host.
-      machine.succeed('grep -Fq "uninitialized drive" /dev/sda')
+      # find the USB Device
+      machine.execute("""
+      # get USB Block Device
+        usb_device=$(lsblk -SJ | jq -r '.blockdevices[] | select(.tran == "usb") | .name')
+        dev="/dev/$usb_device"
+        echo "$dev" > /tmp/usb_device
```
snue, line 177: Here, you create two additional files (`usb_device` and `bus_usb_device`). The pattern of just producing output won't work for that. But I think you can just integrate the necessary parts into the other script and drop this whole script here (and the detour through the files, just use the variables).

snue (commented):
That already clears up the output quite a bit. I left some suggestions to clean it up a bit further.

PR #38, opened 2025-06-24
Add abstractions for the command ring

lbeierlieb:
This PR adds abstractions for command-ring handling itself as well as the Transfer Request Blocks (TRBs) that represent commands, which the driver can place on the command ring.

# TRBs

The controller uses DMA to read TRBs from the command ring; thus, it initially receives the raw byte presentation. When handling the commands, we do not want to be concerned with TRB memory layout and just use the encoded data. The type `CommandTrb` represents the possible commands, and offers parsing via the `try_from(bytes: &[u8])` method.

So far, the implementation parses `Link`, `EnableSlotCommand`, and `AddressDeviceCommand` TRBs completely. For the other command TRBs, it does not yet extract the command-specific data. You can add further parsing can easily on demand (e.g., we will at some point need to handle the `ConfigureEndpointCommand`.

A happy-path parsing test exists for each supported TRB.

# Command Ring

The Command Ring implementation offers two methods for handling reads/writes of the CRCR MMIO register (Command Ring Control Register): `control` for writes and `status` for reads.
Further, the function `next_command_trb` implements checking the command ring for a fresh TRB, parsing, and returning the parsing result together with the memory address of the command. We handle Link TRBs (which indicate ring wrap-around or at least a jump to a different ring segment) correctly.

We remove the old internal implementation of the command ring in the XHCI controller and replace it with the new abstraction. We now handle doorbell writes (to the first doorbell) by checking the command ring and handling potentially retrieved commands. 

# Limitations

The Command Ring Running bit (in the CRCR register) is not yet correctly handled, neither are writes to the CA and CS bits (Command Abort, Command Stop). The running state depends on external factors like the USBSTS register and the doorbell, so it's not trivial to implement correctly. However, we should get an USB device working with just assuming that the ring is always running.

### DIFF
--- a/src/device/pci/rings.rs
+++ b/src/device/pci/rings.rs
@@ -239,7 +239,7 @@ impl CommandRing {
     /// # Limitations
     ///
     /// The current implementation of this function is expecting to only be
-    /// called for initial setup. Any further writes (e.g., driver shopping the
+    /// called for initial setup. Any further writes (e.g., driver stopping the
     /// command ring because a command has timed out) are currently not handled
     /// properly.
     pub fn control(&mut self, value: u64) {
--- a/src/device/pci/trb.rs
+++ b/src/device/pci/trb.rs
@@ -292,7 +292,7 @@ impl LinkTrbData {
     /// the following assumptions to avoid duplicate checks:
     ///
     /// - `value` is a slice of size 16.
-    /// - The TRB type (upper 6 bit of byte 13) indicate a link TRB.
+    /// - The TRB type (upper 6 bit of byte 13) indicates a link TRB.
     ///
     /// # Limitations
     ///
@@ -311,7 +311,7 @@ impl LinkTrbData {
         let ring_segment_pointer = u64::from_le_bytes(rsp_bytes);
         let toggle_cycle = trb_bytes[12] & 0x2 != 0;
 
-        // the lowest for bit of the pointer are RsvdZ to ensure 16-byte
+        // the lowest four bit of the pointer are RsvdZ to ensure 16-byte
         // alignment.
         if ring_segment_pointer & 0xf != 0 {
             return Err(TrbParseError::RsvdZViolation);
@@ -342,7 +342,7 @@ impl AddressDeviceCommandTrbData {
     /// the following assumptions to avoid duplicate checks:
     ///
     /// - `value` is a slice of size 16.
-    /// - The TRB type (upper 6 bit of byte 13) indicate a link TRB.
+    /// - The TRB type (upper 6 bit of byte 13) indicates an address device TRB.
     ///
     /// # Limitations
     ///
@@ -360,7 +360,7 @@ impl AddressDeviceCommandTrbData {
         let icp_bytes: [u8; 8] = trb_bytes[0..8].try_into().unwrap();
         let input_context_pointer = u64::from_le_bytes(icp_bytes);
 
-        // the lowest for bit of the pointer are RsvdZ to ensure 16-byte
+        // the lowest four bit of the pointer are RsvdZ to ensure 16-byte
         // alignment.
         if input_context_pointer & 0xf != 0 {
             return Err(TrbParseError::RsvdZViolation);


src/device/pci/rings.rs
```diff
@@ -151,3 +151,278 @@ impl EventRing {
...
+    /// called for initial setup. Any further writes (e.g., driver shopping the
+    /// command ring because a command has timed out) are currently not handled
+    /// properly.
+    pub fn control(&mut self, value: u64) {
+        if self.running {
+            match value {
+                abort if abort & crcr::CA != 0 => todo!(),
+                stop if stop & crcr::CS != 0 => todo!(),
+                ignored => {
+                    warn!(
```
blitz, line 205: Looks more like `debug!` to me. Is there anything inherently dangerous for the guest to rewrite the register?

src/device/pci/trb.rs
```diff
@@ -187,10 +188,296 @@ pub enum CompletionCode {
...
+pub enum TrbParseError {
+    IncorrectSliceSize(usize),
+    UnsupportedOptionalCommand(u8, String),
+    UnknownTrbType(u8),
+    RsvdZViolation,
+}
+
+impl std::error::Error for TrbParseError {}
+
+impl fmt::Display for TrbParseError {
```
blitz, line 370: Can we use `thiserror` to minimize the boilerplate?

src/device/pci/rings.rs
lbeierlieb: Dangerous not sure, but there is certainly something wrong when the driver performs such a write.
The other fields are
- command ring running, read-only bit
- RsvdP bits
- dequeue pointer and cycle bit, which should not be modified while the command ring is running.

@snue wrote the code and I just moved the code from the xhci controller to the command ring, but I agree with Snu that the `warn!` is justified (notify about a code path that usually should not execute).

src/device/pci/xhci.rs
```diff
@@ -281,6 +281,7 @@ impl PciDevice for Mutex<XhciController> {
             offset::ERSTBA_HI => 0,
             offset::ERDP => self.lock().unwrap().event_ring.read_dequeue_pointer(),
             offset::ERDP_HI => 0,
+            0x2000 => 0, // kernel reads the doorbell after write
```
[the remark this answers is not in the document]
lbeierlieb: We are free to choose the doorbell offsets, so we used 0x2000 (only to report to the driver). Now that we listen to accesses to that address, it makes sense to stabilize this decision. I added a constant and replaced all occurrences of the hard-coded value.

src/device/pci/trb.rs
```diff
@@ -187,10 +188,296 @@ pub enum CompletionCode {
...
+    /// the following assumptions to avoid duplicate checks:
+    ///
+    /// - `value` is a slice of size 16.
+    /// - The TRB type (upper 6 bit of byte 13) indicate a link TRB.
+    ///
+    /// # Limitations
+    ///
+    /// The function currently does not check if the slice respects all RsvdZ
+    /// fields.
+    fn parse(trb_bytes: &[u8]) -> Result<Self, TrbParseError> {
```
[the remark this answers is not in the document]
lbeierlieb: Good point; I added TRB-type checks for the parsing functions.

src/device/pci/xhci.rs
```diff
@@ -237,6 +236,7 @@ impl PciDevice for Mutex<XhciController> {
                 .event_ring
                 .update_dequeue_pointer(value),
             offset::ERDP_HI => assert_eq!(value, 0, "no support for configuration above 4G"),
+            0x2000 => self.lock().unwrap().doorbell(),
```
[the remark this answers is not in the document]
lbeierlieb: already implemented after @phip1611 's suggestion

src/device/pci/rings.rs
```diff
@@ -151,3 +151,278 @@ impl EventRing {
...
+            crcr::CRR
+        } else {
+            0
+        }
+    }
+
+    /// Try to retrieve a new command from the command ring.
+    pub fn next_command_trb(
+        &mut self,
+        dma_bus: BusDeviceRef,
```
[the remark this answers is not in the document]
lbeierlieb: With more overview of the xhci controller functionality, I am now less concerned of being able to directly see in the xhci controller code where DMA requests happen. I took the other approach that you (or Snu) suggested in the last PR: Let the ring owns its own reference.
I changed the implementation of the command ring as well as the event ring.

src/device/pci/trb.rs
```diff
@@ -187,10 +192,286 @@ pub enum CompletionCode {
...
+            trb_type,
+            "LinkTrbData::parse called on TRB data with incorrect TRB type ({:#x})",
+            trb_type
+        );
+
+        let rsp_bytes: [u8; 8] = trb_bytes[0..8].try_into().unwrap();
+        let ring_segment_pointer = u64::from_le_bytes(rsp_bytes);
+        let toggle_cycle = trb_bytes[12] & 0x2 != 0;
+
+        // the lowest for bit of the pointer are RsvdZ to ensure 16-byte
```
snue, line 314:
```suggestion
        // the lowest four bit of the pointer are RsvdZ to ensure 16-byte
```

src/device/pci/trb.rs
```diff
@@ -187,10 +192,286 @@ pub enum CompletionCode {
...
+}
+
+impl AddressDeviceCommandTrbData {
+    /// Parse data of a Address Device Command TRB.
+    ///
+    /// Only `CommandTrb::try_from` should call this function. Thus, we make
+    /// the following assumptions to avoid duplicate checks:
+    ///
+    /// - `value` is a slice of size 16.
+    /// - The TRB type (upper 6 bit of byte 13) indicate a link TRB.
```
snue, line 345: copy and paste mismatch. This isn't for Link trb.

src/device/pci/trb.rs
```diff
@@ -187,10 +192,286 @@ pub enum CompletionCode {
...
+            trb_types::ADDRESS_DEVICE_COMMAND,
+            trb_type,
+            "AddressDeviceCommandTrbData::parse called on TRB data with incorrect TRB type ({:#x})",
+            trb_type
+        );
+
+        let icp_bytes: [u8; 8] = trb_bytes[0..8].try_into().unwrap();
+        let input_context_pointer = u64::from_le_bytes(icp_bytes);
+
+        // the lowest for bit of the pointer are RsvdZ to ensure 16-byte
```
snue, line 363:
```suggestion
        // the lowest four bit of the pointer are RsvdZ to ensure 16-byte
```

src/device/pci/trb.rs
```diff
@@ -187,10 +192,286 @@ pub enum CompletionCode {
...
+        // alignment.
+        if input_context_pointer & 0xf != 0 {
+            return Err(TrbParseError::RsvdZViolation);
```
snue, line 366: nit: maybe this is rather an `AlignmentViolation`? You could even remove the comment with that, as it would be self-explanatory to a degree.

snue (commented):
I like this very much! The unit tests are great! I only spotted a couple typos.

blitz (approved):

src/device/pci/trb.rs
lbeierlieb: When you look at [XHCI spec](https://www.intel.com/content/dam/www/public/us/en/documents/technical-specifications/extensible-host-controler-interface-usb-xhci.pdf) Figure 6-25, you can see that the authors specify the alignment by requiring the lowest bits of the pointer to be RsvdZ.

Introducing an `AlignmentViolation` does not hurt and can be useful, but considering that we do not expect the Linux driver to provide misaligned pointers, I keep it as is for now.

PR #39, opened 2025-06-26
Add a test report to the integration smoke test

snue:
This is here to run the Hercules CI on the branch for #35 (and merge with green CI). All discussion on the content should happen over there, so @ziyifu225 can address feedback.

> [!Note]
> This is a rebased branch so GitHub figures out how to open a new PR after the branches diverged.

### DIFF
--- a/nix/tests.nix
+++ b/nix/tests.nix
@@ -77,32 +77,25 @@ let
 
   # Report for target device path and access permissions
   usbDeviceInfoScript = pkgs.writeShellScript "usb-device-info" ''
-    {
-      echo "----- USB Block Device (/dev/sdX) -----"
-      if [ -f /tmp/usb_device ]; then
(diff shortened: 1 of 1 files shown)


nix/tests.nix
```diff
@@ -80,6 +80,11 @@ in
     name = "usbvfiod Smoke Test";
 
     nodes.machine = { pkgs, ... }: {
+      environment.systemPackages = [
```
phip1611, line 83: nit: ```suggestion
      environment.systemPackages = with pkgs; [
```

File: nix/tests.nix [hunk 2]
```diff
@@ -80,6 +80,11 @@ in
...
+        pkgs.usbutils
+        pkgs.jq
```
phip1611, line 85: nit: not sorted alphabetically

PR #40, opened 2025-06-27
Device discovery

lbeierlieb:
So far, we properly initialize the controller. We do not indicate to the driver that there is a device connected, so there is no further action from the driver.

This PR adds the necessary functionality to make the Linux kernel announce that it discovered a device:
- set and handle USBSTS and PORTSC registers to indicate that a device is connected to a port
- add abstractions to manage device slots
- handle Enable Slot Command by providing a Slot ID to the driver
- handle Address Device Command by initializing device context
- Transfer Ring abstraction
  - retrieve Transfer Ring from Device Context
  - read USB Requests from Transfer Ring (consisting of multiple TRBs)
- handle device doorbell by retrieving USB request
- missing: forward the request to the real device.

The last commit adds code to interact with the `nusb::Device`, but I ran out of time, so the last commit breaks the code and prevents it from compiling. See it as a potential starting point for exploring device interaction.

The other commits contain mostly code that I consider merge-worthy (there might be some shortcuts in places), but documentation and testing is lacking.

Without the last commit, the code runs. It responds to requests on the control transfer ring by sending Transfer Events, but no requested data (the descriptors) are copied to guest memory. The driver realizes that fact, retries to get the device descriptor a few times, before giving up the device discovery.
image

@snue Feel free to leave feedback or work yourself on this branch next week.

### DIFF
--- a/src/device/pci/xhci.rs
+++ b/src/device/pci/xhci.rs
@@ -182,7 +182,12 @@ impl XhciController {
                 let (completion_code, slot_id) = self.handle_enable_slot();
                 EventTrb::new_command_completion_event_trb(address, 0, completion_code, slot_id)
             }
-            CommandTrb::DisableSlot => todo!(),
+            CommandTrb::DisableSlot => {
+                // TODO this command probably requires more handling.
+                // Currently, we just acknowledge to not crash usbvfiod in the
+                // integration test.
+                EventTrb::new_command_completion_event_trb(address, 0, CompletionCode::Success, 1)
+            }
             CommandTrb::AddressDevice(data) => {
                 self.handle_address_device(&data);
                 EventTrb::new_command_completion_event_trb(
@@ -195,7 +200,12 @@ impl XhciController {
             CommandTrb::ConfigureEndpoint => todo!(),
             CommandTrb::EvaluateContext => todo!(),
             CommandTrb::ResetEndpoint => todo!(),
-            CommandTrb::StopEndpoint => todo!(),
+            CommandTrb::StopEndpoint => {
+                // TODO this command probably requires more handling.
+                // Currently, we just acknowledge to not crash usbvfiod in the
+                // integration test.
+                EventTrb::new_command_completion_event_trb(address, 0, CompletionCode::Success, 1)
+            }
             CommandTrb::SetTrDequeuePointer => todo!(),
             CommandTrb::ResetDevice => todo!(),
             CommandTrb::ForceHeader => todo!(),


src/device/pci/rings.rs
```diff
@@ -349,6 +356,197 @@ impl CommandRing {
...
+impl TransferRing {
+    pub fn new(endpoint_context: EndpointContext, dma_bus: BusDeviceRef) -> Self {
+        Self {
+            endpoint_context,
+            dma_bus,
+        }
+    }
+
+    // TODO rewrite this function to work like the CommandRing::next_command_trb
+    pub fn next_transfer_trb(&self) -> Option<(u64, Result<TransferTrb, TrbParseError>)> {
```
snue, line 374: I wonder if it would make sense to implement `Iterator` semantics for these rings. https://doc.rust-lang.org/rust-by-example/trait/iter.html

lbeierlieb:
Clippy, the smoke-test, and I are now happy with the current state of the PR. @blitz ready for review

src/device/pci/registers.rs
```diff
@@ -0,0 +1,42 @@
+/// A simple PORTSC register implementation supporting RW1C bits.
+///
+/// The PORTSC register requires us to initially set some bits and
+/// later react to 1-to-clear writes (RW1C) to get a device to show up.
+/// Perhaps later we need more fine-grained access to the bits or state
+/// handling, but we can use the simplistic implementation for now.
+#[derive(Debug, Clone)]
+pub struct PortscRegister {
```
blitz, line 8: This looks like a perfect opportunity for a happy path unit test.

src/device/pci/xhci.rs
```diff
@@ -91,6 +92,7 @@ impl XhciController {
             interrupt_management: 0,
             interrupt_moderation_interval: runtime::IMOD_DEFAULT,
             interrupt_line: Arc::new(DummyInterruptLine::default()),
+            portsc: PortscRegister::new(0x00260203),
```
blitz, line 96: Is it worth adding constants for these values?

src/device/pci/xhci.rs
```diff
@@ -106,7 +106,7 @@ impl XhciController {
     /// Obtain the current host controller status as defined for the `USBSTS` register.
     #[must_use]
     pub fn status(&self) -> u64 {
-        !u64::from(self.running) & 0x1u64
+        !u64::from(self.running) & 0x1u64 | 0x8 | 0x10
```
blitz, line 110: Is it worth adding constants for these values?

src/device/pci/device_slots.rs
```diff
@@ -0,0 +1,111 @@
...
+    ///
+    /// There should only exist one `DeviceSlotManager` per `XhciController`.
+    ///
+    /// # Parameters
+    ///
+    /// - num_slots: number of available slots. Use the same value as the
+    ///   controller reports in HCSPARAMS1.
+    /// - dma_bus: a reference to the guest's memory.
+    pub const fn new(num_slots: usize, dma_bus: BusDeviceRef) -> Self {
+        Self {
```
blitz, line 46: `assert!(num_slots > 0)` ?

src/device/pci/device_slots.rs
```diff
@@ -0,0 +1,111 @@
...
+        self.dcbaap = dcbaap;
+    }
+
+    /// Retrieve one of the available slot IDs.
+    ///
+    /// If a unused slot is available, this function returns the slot ID.
+    /// Otherwise, it returns `Option::None`.
+    pub fn reserve_slot(&mut self) -> Option<usize> {
+        let available_slot_id =
+            (1..=self.num_slots).find(|slot_id| !self.used_slots.contains(slot_id));
```
blitz, line 71: As we can only have 256 slots, this linear complexity seems ok. But I would add this as a comment somewhere so people don't complicate it.

src/device/pci/device_slots.rs
```diff
@@ -0,0 +1,111 @@
...
+///
+/// - index of the device context base address array (DCBAA), which points to
+///   the associated device context.
+/// - index of the doorbell register.
+/// - referenced in event and command TRBs
+///
+/// The XHCI controller reports the maximum number of device slots in the
+/// HCSPARAMS1 register. For device initialization, the driver requests a slot
+/// ID using the Enable Slot Command. The `DeviceSlotManager` is responsible
+/// of tracking which slot IDs are currently in use.
```
blitz, line 22: But why does it need DMA access to do that?

src/device/pci/xhci.rs
```diff
@@ -46,9 +47,8 @@ pub struct XhciController {
     /// Configured device slots.
     slots: Vec<()>,
 
-    /// Device Context Array
-    /// TODO: currently just the raw pointer configured by the OS
-    device_contexts: Vec<u64>,
+    /// Device Slot Management
+    device_slot_manager: DeviceSlotManager,
```
blitz, line 51: How does the vector of `<()>` relate to the device slot manager? Do we need the `slots` member?

src/device/pci/device_slots.rs
```diff
@@ -70,6 +70,188 @@ impl DeviceSlotManager {
...
+    ///
+    /// - slot_id: the slot ID for which the DeviceContext is requested.
+    pub fn get_device_context(&self, slot_id: u8) -> DeviceContext {
+        assert!(
+            self.used_slots.contains(&(slot_id as usize)),
+            "requested DeviceContext for unassigned slot_id"
+        );
+        // lookup address of device context in device context base address array
+        let device_context_address = self.dma_bus.read(Request::new(
+            self.dcbaap + slot_id as u64 * 8,
```
blitz, line 97: This needs to handle overflow. I would suggest using `self.dcbaap.overflowing_add` here to make it explicit that overflow is ok. (The driver wants to shoot itself in the foot probably, but it's ok from the device emulation point's of view.)

All computation with addresses provided by the guest should use overflowing arithmetic. (Just as the device would.)

blitz: This problem exists in other places as well.

src/device/pci/device_slots.rs
```diff
@@ -70,6 +70,188 @@ impl DeviceSlotManager {
...
+
+        // set slot state to addressed
+        let slot_state_addressed = 2;
+        self.dma_bus
+            .write_bulk(self.address + 15, &[slot_state_addressed << 3; 1]);
+
+        // set endpoint state to enabled
+        let ep_state_running = 1;
+        self.dma_bus
+            .write_bulk(self.address + 32, &[ep_state_running]);
```
blitz, line 184: This writes to the same address as above. The same happens for the write to `address + 15`. Is this intentional? Can you make the intention here more obvious? I guess more comments would be fine, if there is no way to restructure the code.

src/device/pci/device_slots.rs
```diff
@@ -70,6 +70,188 @@ impl DeviceSlotManager {
...
+    /// Give access to an endpoint context based on its index in the device
+    /// context.
+    ///
+    /// # Parameters
+    ///
+    /// - index: index in the device context. `1 <= index <= 31`.
+    fn get_endpoint_context_internal(&self, index: u64) -> EndpointContext {
+        assert!(
+            (1..31).contains(&index),
+            "index has to be between 1 and 31 (inclusive)"
```
blitz, line 196: nit: Obvious message is obvious. :) Remove?

src/device/pci/device_slots.rs
```diff
@@ -70,6 +70,188 @@ impl DeviceSlotManager {
...
+
+        EndpointContext::new(self.address + 32 * index, self.dma_bus.clone())
+    }
+
+    /// Give access to context of the default control endpoint.
+    /// direction.
+    ///
+    /// Endpoint 0 is a special endpoint. It always exists and it is bi-directional.
+    fn get_control_endpoint_context(&self) -> EndpointContext {
+        self.get_endpoint_context_internal(1)
```
blitz, line 207: Mmh. Somehow the indices change from 0-based to 1-based. This should be explained in the code.

blitz (commented):
I managed to get to commit 6 of 17. Will continue later!

src/device/pci/trb.rs
```diff
@@ -377,6 +378,138 @@ impl AddressDeviceCommandTrbData {
     }
 }
 
+/// Represents a TRB that the driver can place on a transfer ring.
+#[derive(Debug)]
```
blitz, line 443: We should agree on some standard traits to derive. `Debug` is already very good, I would also lobby for `Clone` (and `Copy` if it's just a couple of bytes big), and `PartialEq`/`Eq`.

src/device/pci/xhci.rs
```diff
@@ -296,6 +296,8 @@ impl PciDevice for Mutex<XhciController> {
             offset::DNCTL => 2,
             offset::CRCR => self.lock().unwrap().command_ring.status(),
             offset::CRCR_HI => 0,
+            offset::DCBAAP => self.lock().unwrap().device_slot_manager.get_dcbaap(),
+            offset::DCBAAP_HI => 0,
```
blitz, line 345: Shouldn't this be `dcbaap >> 32`?

src/device/pci/usbrequest.rs
```diff
@@ -0,0 +1,59 @@
...
+pub struct UsbRequest {
+    pub request_type: u8,
+    pub request: u8,
+    pub value: u16,
+    pub index: u16,
+    pub length: u16,
+    pub data: Option<u64>,
+}
+
+impl UsbRequest {
```
blitz, line 11: These constructors don't seem very helpful. Is there any advantage over constructing the `UsbRequest` directly?

They might even harm readability, because `index: 1; length: 4` become `1, 4` when using the constructors. It also makes it easier to mix up parameters.

src/device/pci/rings.rs
```diff
@@ -450,6 +452,133 @@ impl TransferRing {
 
         Some(trb_result)
     }
+
+    /// take setup+data+status TRBs or setup+status TRBs from transfer ring
+    /// and extract the request information
+    pub fn next_request(&self) -> Option<Result<(u64, UsbRequest), RequestParseError>> {
```
blitz, line 458: I would somehow expect `Result<Option<UsbRequest>, SomeError>` or maybe `Option<Result<UsbRequest, SomeError>>`. From reading the code, I still don't understand why we need to hand the address around.

src/device/pci/rings.rs
```diff
@@ -450,6 +452,133 @@ impl TransferRing {
...
+                    trb,
+                )));
+            }
+        };
+
+        let second_trb = self.next_transfer_trb();
+        let data_trb_or_address = match second_trb {
+            None => {
+                // there should follow either Data or Status Stage
+                return Some(Err(RequestParseError::MissingTrb));
```
blitz, line 486: What does the spec say here? Is it an error to have "incomplete" requests in the ring? I would assume that the controller just waits for more TRBs to become available in this case. This would imply a state machine for handling TRBs.

For now the code is ok, I'm just trying to understand what we would like to have eventually.

src/device/pci/rings.rs
```diff
@@ -450,6 +452,133 @@ impl TransferRing {
...
+    pub fn next_request(&self) -> Option<Result<(u64, UsbRequest), RequestParseError>> {
+        let first_trb = self.next_transfer_trb();
+        let setup_trb_data = match first_trb {
+            None => {
+                // no TRB available -> no request available
+                return None;
+            }
+            Some((_, Err(err))) => {
+                // got a erroneous TRB, pass error on
+                return Some(Err(RequestParseError::TrbParseError(err)));
```
blitz, line 467: Is there a downside to `return Err(...)`? (i.e. flipping `Option` and `Result`?)

src/device/pci/rings.rs
```diff
@@ -450,6 +452,133 @@ impl TransferRing {
...
+        Some(Ok((address, request)))
+    }
+}
+
+#[derive(Error, Debug)]
+pub enum RequestParseError {
+    #[error("Encountered invalid TRB")]
+    TrbParseError(TrbParseError),
+    #[error("Encountered unexpected TRB type. Expected type(s) {0:?}, got TRB {1:?}")]
+    UnexpectedTrbType(Vec<u8>, TransferTrb),
```
blitz, line 579: Aren't they all `[u8; 16]`?

src/device/pci/xhci.rs
```diff
@@ -225,6 +225,48 @@ impl XhciController {
...
+            .get_control_transfer_ring();
+
+        let (address, request) = match transfer_ring.next_request() {
+            None => panic!(
+                "Device doorbell was rang, but there is no request on the control transfer ring"
+            ),
+            Some(Err(err)) => panic!(
+                "Failed to retrieve request from control transfer ring: {:?}",
+                err
+            ),
```
blitz, line 248: You can simplify the code by dying where the error comes up. At the moment, there is a lot of code to bubble up the error.

src/device/pci/device_slots.rs
lbeierlieb: There are more guest address calculations in the code base already. I'm not addressing the issue in this PR, but I created an internal issue for a follow-up PR which corrects the calculations in all places (and updates the developer guidelines).

src/device/pci/trb.rs
lbeierlieb: I created an internal issue to address this problem in the whole code base (after this PR).

src/device/pci/trb.rs
```diff
@@ -377,6 +378,138 @@ impl AddressDeviceCommandTrbData {
...
+    ///
+    /// While this function can parse all available Transfer TRB types, it does
+    /// not parse all of them in full detail. If the function returns only the
+    /// enum variant without an associated struct, the parsing for the
+    /// particular command is not yet implemented.
+    fn try_from(bytes: &[u8]) -> Result<Self, Self::Error> {
+        let slice_size = bytes.len();
+        if slice_size != 16 {
+            return Err(TrbParseError::IncorrectSliceSize(slice_size));
+        }
```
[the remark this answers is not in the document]
lbeierlieb: Also affects the existing command TRBs. I created a separate issue; will address after this PR.

src/device/pci/xhci.rs
lbeierlieb: yes, done.

src/device/pci/device_slots.rs
lbeierlieb: added a note to the function documentation about the linear time complexity.

src/device/pci/device_slots.rs
lbeierlieb: device context entries: 
   - entry 0: slot context
   - entry 1: endpoint context for endpoint 0
   - entry 2: endpoint context for endpoint 1 OUT
   - entry 3: endpoint context for endpoint 1 IN
   - entry 4: endpoint context for endpoint 2 OUT
   - ...
   - entry 31: endpoint context for endpoint 15 IN

I added the explanation in the code.

src/device/pci/trb.rs
```diff
@@ -500,4 +633,87 @@ mod tests {
...
+            assert!(
+                link_data.toggle_cycle,
+                "toggle_cycle bit was parsed incorrectly."
+            );
+        } else {
+            panic!(
+                "A TRB with TRB type 6 should result in a TransferTrb::Link. Got instead: {:?}",
+                trb
+            );
+        }
```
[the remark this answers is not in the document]
lbeierlieb: makes sense, but currently `PartialEq` is missing. I already created an issue for implementing default traits; shortening the tests is also part of the issue's scope.

src/device/pci/trb.rs
```diff
@@ -149,6 +151,65 @@ impl PortStatusChangeEventTrbData {
...
+    /// # Parameters
+    ///
+    /// - `trb_pointer`: Pointer to the transfer even that generated the event.
+    /// - `trb_transfer_length`: Residual number of bytes not transferred.
+    /// - `completion_code`: Encodes the completion status of the associated
+    ///   transfer.
+    /// - `event_data`: Whether this event was generated by an Event Data TRB.
+    /// - `endpoint_id`: On which endpoint the transfer happened.
+    /// - `slot_id`: On which slot the transfer happened.
+    pub const fn new_transfer_event_trb(
```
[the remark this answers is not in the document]
lbeierlieb: Yes, I would like to see all possible event types when I start type `EventTrb::new_`, that's also why this constructor is a function of `EventTrb` and not `TransferEventTrbData` and returns the `TransferEventTrbData` inside of the `EventTrb`. I don't want other code to directly use `TransferEventTrbData` anywhere.

Does that intentions make sense? @blitz

src/device/pci/rings.rs
lbeierlieb: yes, separate issue exists to fix across all trb implementations. Not addressing in this PR

src/device/pci/trb.rs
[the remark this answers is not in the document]
blitz: Sounds good.

src/device/pci/device_slots.rs
```diff
@@ -0,0 +1,314 @@
...
+    /// - entry 3: endpoint context for endpoint 1 IN
+    /// - entry 4: endpoint context for endpoint 2 OUT
+    /// - ...
+    /// - entry 31: endpoint context for endpoint 15 IN
+    ///
+    /// # Parameters
+    ///
+    /// - index: index in the device context. `1 <= index <= 31`.
+    fn get_endpoint_context_internal(&self, index: u64) -> EndpointContext {
+        assert!((1..31).contains(&index));
```
snue, line 204:
```suggestion
        assert!((1..=31).contains(&index));
```

src/device/pci/xhci.rs
snue: Yes, but we do not announce 64 bit support for the controller currently, so the guest places all the context structures within the lower 32 bits for us. We assert this is currently the case on the MMIO write above.

src/device/pci/xhci.rs
lbeierlieb: `slots` was part of dummy implementation of the device slots in the XHCI. I missed to remove/replace the existing dummy code. Now, everything device slots in handled with the `device_slot_manager`.

src/device/pci/xhci.rs
```diff
@@ -266,7 +266,9 @@ impl PciDevice for Mutex<XhciController> {
                 .update_dequeue_pointer(value),
             offset::ERDP_HI => assert_eq!(value, 0, "no support for configuration above 4G"),
             offset::DOORBELL_CONTROLLER => self.lock().unwrap().doorbell_controller(),
-            _ => todo!(),
+            addr => {
+                panic!("unknown write {}", addr);
```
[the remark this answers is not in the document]
lbeierlieb: yes, done

src/device/pci/rings.rs
lbeierlieb: will address in follow-up by adding addresses to trb and request structs

PR #44, opened 2025-07-15
xhci_backend: add memory barrier before interrupt

snue:
Ensure all guest memory modifications are realized before raising the interrupt to the guest.

This is most likely not needed, as the read/write system calls on the `EventFd` are supposed to act as a sufficient barrier. Guest kernel drivers will expect "DMA IO semantics" from the device, and will employ sufficient barriers (`dma_rmb()` / `dma_wmb()`). Let's add the explicit synchronization and see if it has measurable impact later on.

FYI: this inserts an `mfence` instruction in the `x86_64` release build.

### DIFF
--- a/src/xhci_backend.rs
+++ b/src/xhci_backend.rs
@@ -41,6 +41,10 @@ struct InterruptEventFd {
 
 impl InterruptLine for InterruptEventFd {
     fn interrupt(&self) {
+        // Ensure all guest memory modifications can be observed by another process.
+        // The sequentially consistent fence emits a hardware memory barrier operation.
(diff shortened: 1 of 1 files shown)


lbeierlieb:
Right, I (and probably @snue as well) considered only the scenario where the driver reacts to events once it receives the interrupt. That assumption might hold true for the Linux driver, but the XHCI spec mentions "Interrupts are optional. The xHC may be managed by polling Event Rings.".

snue:
> So it's important to be aware of these issues. If we have/introduce a function to write completion events, this is also a great opportunity to drop a mfence.

Indeed, we will have to touch additional places to be sure. I'll leave that for follow-up MRs and merge this as is.

PR #47, opened 2025-07-17
Expose real device to the guest

snue:
This PR adds a basic abstraction that allows to forward control transfers (not endpoint transfers) from the virtual xhci controller to an `nusb` device handle. It also acknowledges the basic commands required to survive the initial device detection.

I added another check to the `smoke-test` to see the device popping up in the Guest `lsusb` output.

Marking this as draft, as there is some cleanup left (mostly around the `RealDevice` abstraction).

TODO:
- [X] Fix/improve the clone semantics -> just removed the Clone requirement and simplified
- [ ] return success/errors from  `control_transfer()`
    - Decided not to do this now. There are more details to be discussed for the abstraction. I just raised the possible errors from debug to more visible log warnings. 
- [X] more/better default trait implementations? (`lolusb` :grin:)
    -  Just printing the active configuration in the Debug trait now. This does not incur any I/O according to the manual. 
- [ ] support more than one device? Tie the device to a port?
   - we leave this for later when we also provide a second (or more) port/slot

### DIFF
--- a/nix/tests.nix
+++ b/nix/tests.nix
@@ -38,6 +38,8 @@ let
               ExecStart = pkgs.writeShellScript "diagnostic-dump" ''
                 echo Dumping Diagnostics
                 cat /proc/interrupts
+                echo
+                ${pkgs.usbutils}/bin/lsusb
               '';
               StandardOutput = "journal+console";
               StandardError = "journal+console";
@@ -183,6 +185,7 @@ in
 
       # Read the diagnostic information after login.
       machine.wait_until_succeeds("grep -Eq '\s+[1-9][0-9]*\s+PCI-MSIX.*xhci_hcd' ${cloudHypervisorLog}")
+      machine.wait_until_succeeds("grep -q 'ID ${vendorId}:${productId} QEMU QEMU USB HARDDRIVE' ${cloudHypervisorLog}")
     '';
   };
 }


src/device/pci/xhci.rs
```diff
@@ -217,7 +235,18 @@ impl XhciController {
                 )
             }
             CommandTrbVariant::SetTrDequeuePointer => todo!(),
-            CommandTrbVariant::ResetDevice => todo!(),
+            CommandTrbVariant::ResetDevice => {
+                // TODO this command probably requires more handling.
+                // Currently, we just acknowledge to not crash usbvfiod in the
+                // integration test.
```
snue, line 241: The comment isn't quite accurate. It crashes when testing/developing locally with arbitrary devices. The device from the integration test does not trigger this path.

lbeierlieb (approved):
I added some minor suggestions, which you could address before merging. The overview of the changes in this PR look fine.

What bother me a bit is the organization of the commits. I think "[xhci: remove the derived Clone trait](https://github.com/cyberus-technology/usbvfiod/pull/47/commits/891a6f664d340e8f5c7a6d3f25e53a4a6134038c)" should not also change the Debug implementation, which could very well be in a separate atomic commit.

Staying with the same commit: it removes code that the first or second commit added, and there are multiple example where one commit introduces something and a later commit in the same PR removes or changes parts. I see the difficulty of working off existing (prototyping) commits and wanting to keep them for authorship tracking, but I would rather see a series of atomic commits that add or change different things without my name anywhere instead of the current back-and-forth changes. I tried to go through and review commit by commit, but it didn't make much sense because so many things I looked at were changed by later commits.

Could you split/squash/reorganize the commits, such that there are not "reverts" of previously added changes before merging?

src/device/pci/nusb.rs
```diff
@@ -0,0 +1,89 @@
...
+impl Debug for NusbDeviceWrapper {
+    fn fmt(&self, f: &mut std::fmt::Formatter<'_>) -> std::fmt::Result {
+        // The active configuration is either cached or not available
+        // for unconfigured devices. There is no I/O for this.
+        match self.device.active_configuration() {
+            Ok(active_configuration) => f
+                .debug_struct("NusbDeviceWrapper")
+                .field("device", &active_configuration)
+                .finish(),
+            Err(_) => f.write_str("Unconfigured USB device"),
```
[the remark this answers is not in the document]
snue: I was probably just lazy 😁. I'll go for `debug_struct` for the error case as well.

src/device/pci/xhci.rs
snue: And yes, this was a result of copy&paste from the other command handling. :sweat_smile:

PR #48, opened 2025-07-18
Add unit test for retrieving requests from the transfer ring

lbeierlieb:
The tests checks whether `TransferRing::next_request` can

- correctly parse a request out of three TRBs (setup stage + data stage + status stage)
- correctly parse a request out of two TRBs (setup stage + status stage)
- correctly handles ring wrap-arounds/Link TRBs

The test indirectly also tests the underlying `TransferRing::next_transfer_trb` function.

A few changes outside the test were necessary:

- added `derive(PartialEq, Eq)` to a number of types to make the return value of `TransferRing::next_request` usable with `assert_eq!`
- `EndpointContext::new` needs to be public so that the test can construct an endpoint context without going the whole `DeviceContext` setup process.
- Added non-bulk reads and writes to the testing RAM device, because the endpoint context retrieves and sets the the dequeue pointer of the transfer ring using non-bulk reads and writes.

### DIFF
--- a/src/device/pci/device_slots.rs
+++ b/src/device/pci/device_slots.rs
@@ -243,7 +243,7 @@ impl EndpointContext {
     ///
     /// - address: the address of the endpoint context in guest memory.
     /// - dma_bus: reference to the guest memory.
-    const fn new(address: u64, dma_bus: BusDeviceRef) -> Self {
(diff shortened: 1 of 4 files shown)


snue (approved):
LGTM! Thanks.

PR #49, opened 2025-07-21
Use wrapping_add to calculate DMA addresses

lbeierlieb:
Previously, when using normal additions to calculate DMA addresses, behavior differed between builds:

- debug builds crash when an overflow occurs due to inserted overflow
checks
- release build perform wrap-around on overflow

This PR introduces `u64::wrapping_add` for wrap-around behavior, just as a real controller most likely has, and adds a note for future contributions in the developer guidelines. 

Note: wrapping of addresses should not occur during normal operation, only when the USB driver in the guest wants to shoot itself in the foot.

### DIFF
--- a/docs/developers/guidelines.md
+++ b/docs/developers/guidelines.md
@@ -13,9 +13,4 @@ operation, only when the USB driver in the guest wants to shoot
 itself in the foot. In such cases, we want to behave like a physical
 XHCI controller and wrap around.
 
-If we used normal calculations, the behavior in overflow situations
-would differ between builds:
-
-- debug builds crash when an overflow occurs due to inserted overflow
-checks
-- release build perform wrap-around on overflow
+Explicitly treating the calculations as wrapping aligns the behavior between debug and release builds.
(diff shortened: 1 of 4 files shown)


docs/developers/guidelines.md
```diff
@@ -0,0 +1,21 @@
...
+operation, only when the USB driver in the guest wants to shoot
+itself in the foot. In such cases, we want to behave like a physical
+XHCI controller and wrap around.
+
+If we used normal calculations, the behavior in overflow situations
+would differ between builds:
+
+- debug builds crash when an overflow occurs due to inserted overflow
+checks
+- release build perform wrap-around on overflow
```
snue, line 21: This is correct, but with the listing at a glance it looks like that is the "wanted" behavior. Maybe just:


```suggestion
Explicitly treating the calculations as wrapping aligns the behavior between debug and release builds.
```

snue (approved):
LGTM. Just suggested different wording in the guidelines.

PR #50, opened 2025-07-23
Add required release barriers before event-ring writes

lbeierlieb:
I found two occasions where we we should introduce memory barriers.

First, right now, we handle our Address Device Command as follows.

```
copy_data_from_input_context_to_device_context()
write_completion_event_to_event_ring()
fence(SeqCst)
interrupt()
```

When the driver starts reading memory only once it received the interrupt, everything is in place and there is no problem. However, the driver is allowed to poll the event ring. Thus, when the completion event write completes before the context copy, the driver might right invalid data when accessing the device context.

The same situation happens with our real-device communication on device-to-host transfers.

```
copy_device_response_to_guest();
write_transfer_event_to_event_ring();
// interrupt as before...
```

The driver might poll the event ring, see the transfer event, and read garbage from the data buffer, because the copy has not yet completed.

The solution is to place a `fence(Ordering::Release)` before the event writes, preventing the reordering of stores and ensuring that the device context/transfer data stores complete before the event stores.

x86 does not reorder stores, so the fence is a no-op on that architecture. But on other architectures (such as ARM), stores can get reordered. Further, the barrier also prevents the compiler from reordering stores, which is also crucial on x86.


src/device/pci/xhci.rs
```diff
@@ -204,6 +207,10 @@ impl XhciController {
             }
             CommandTrbVariant::AddressDevice(data) => {
                 self.handle_address_device(&data);
+                // handle_address_device copies 64 bytes to the device context.
+                // The stores have to be finished before the command completion
+                // event is written.
+                fence(Ordering::Release);
```
blitz, line 213: Shouldn't this move to `event_ring.enqueue` at the bottom of this function (or maybe even into this function)? Otherwise, we have to remember to add it for each individual completion event.

snue: Fully agree, this should just be part of writing any completion events. With this being only a relatively cheap Release barrier, I also don't see an issue with adding it for all event ring updates.

PR #51, opened 2025-07-24
Derive standard traits in all custom types and create note in developer guidelines

ziyifu225:
### Reason for this PR
Streamline the codebase by keeping only the trait derives that are truly needed—reducing boilerplate, preventing unintended copies of stateful structs, and clarifying our guidelines for future contributors.
### Changes in this PR

- Removed redundant derives

    Eliminated several unused Clone annotations from state‑bearing structs (e.g., memory segments, rings, registers) to avoid accidental copying.

-  Added essential derives

    Supplied `Debug`, `PartialEq`, `Eq`, or `Copy `+` Clone` only where testing, logging, or lightweight value semantics genuinely require them.

- Documentation update

    Expanded `guidelines.md` with a concise “Trait Derivations” section that clarifies when each common derive should or shouldn’t be used.
### Test
All changes compile and unit tests pass locally (`cargo check`, `cargo test`).

### DIFF
--- a/docs/developers/guidelines.md
+++ b/docs/developers/guidelines.md
@@ -40,3 +40,4 @@ For consistency and readability, follow this trait order when deriving multiple 
 
 ```rust
 #[derive(Debug, Clone, Copy, PartialEq, Eq, Default)]
+```
--- a/docs/developers/guidelines.md
+++ b/docs/developers/guidelines.md
@@ -14,3 +14,29 @@ itself in the foot. In such cases, we want to behave like a physical
 XHCI controller and wrap around.
 
 Explicitly treating the calculations as wrapping aligns the behavior between debug and release builds.
+
+## Trait Derivations
+
+Deriving standard traits improves debuggability, testability, and code clarity. While not strictly mandatory, developers are encouraged to apply them where safe and semantically appropriate.
+
+- **Always derive `Debug`**  
+  Include `#[derive(Debug)]` for all structs and enums to enable diagnostic output and logging.
+
+- **Derive `Clone` or `Copy`**  
+  Use `Copy` only for small types with trivial duplication semantics (e.g., numeric types). Use `Clone` for safe but potentially more complex duplication.  
+  _Avoid both_ if the type encapsulates ownership-sensitive resources (e.g., file handles, raw pointers, custom lifetimes).
+
+- **Derive `PartialEq`, `Eq`**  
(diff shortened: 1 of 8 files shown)


lbeierlieb (changes requested):
Thanks for the thorough check of the codebase, I especially appreciate the explanations in your commit messages!
The changes look good to me, I only left a very minor suggestion for an addition.

However, I would like to see `PartialEq` and `Eq` implemented for `CommandTrb` and `TransferTrb`, as well as have the tests in `trb.rs` and `rings.rs` make use of the equality. Right now, there is a lot of matching and comparing individual fields happening to check whether TRBs were parsed correctly.

docs/developers/guidelines.md
```diff
@@ -14,3 +14,29 @@ itself in the foot. In such cases, we want to behave like a physical
...
+  Provide `Default` when a clear and unambiguous zero-configuration or initial state exists.
+
+Derives may be added proactively for likely future needs (e.g., test support or logging), provided they don’t introduce ambiguity or misuse.
+
+### Recommended Derive Order
+
+For consistency and readability, follow this trait order when deriving multiple traits:
+
+```rust
+#[derive(Debug, Clone, Copy, PartialEq, Eq, Default)]
```
snue, line 42: You need to close this code block with backticks.

src/cli.rs
```diff
@@ -47,6 +47,7 @@ pub struct Cli {
 }
 
 /// The location of the server socket for the vfio-user client connection.
+#[derive(Debug, Clone, Copy)]
```
snue, line 50: Your guidelines mention file descriptors as problematic with Copy/Clone for the ownership semantics. This enum encapsulates a `RawFd`. Should we skip `Copy`/`Clone` here for now?

docs/developers/guidelines.md
lbeierlieb: I was confused here as well, but the github rendering displayed the code block correctly, so I thought the next guideline update will have to close this block. I agree that ideally the backticks should be added now.

ziyifu225: Good catch — I indeed forgot to close the code block. I've added the missing backticks in the latest update. Thanks!

src/cli.rs
ziyifu225: You are right! I've remove Copy/Clone from `ServerSocket` to avoid `fd` ownership issues.

lbeierlieb (approved):
Very nice!

PR #53, opened 2025-07-29
Parse Stop Endpoint Command

lbeierlieb:
Fully parse the Stop Endpoint Command.

I currently (during my endpoint configuration experiments) would like the information about the `endpoint_id` encoded in a Stop Endpoint Command. Because the functionality can stand on its own and will be needed later, I put it into this smaller PR.

Additionally, the PR aligns the names of the tests in `trb.rs` with the test names in the rest of the project.

### DIFF
--- a/src/device/pci/trb.rs
+++ b/src/device/pci/trb.rs
@@ -696,7 +696,7 @@ mod tests {
     use super::*;
 
     #[test]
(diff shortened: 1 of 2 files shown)


snue (approved):
Looks good!

PR #55, opened 2025-08-04
Add logic for detecting event ring wraparound

ziyifu225:
## Why this PR
The current implementation of the Event Ring does not support wraparound behavior. Once the TRB entries in the ring’s single memory segment are exhausted, `usbvfiod` will crash due to a lack of space for new Transfer Event TRBs.

This limitation has gone unnoticed because our tests so far have not triggered a full ring—256 entries have been sufficient. However, with real data flowing through transfer rings, the Event Ring will eventually overflow unless wraparound is properly handled.

This PR updates the enqueue function of the Event Ring to correctly wrap around to the start of its single memory segment when reaching the end, ensuring that the software can continue processing new events without failure.

The change is based on Section 4.9.4 of [the xHCI specification](https://www.intel.com/content/dam/www/public/us/en/documents/technical-specifications/extensible-host-controler-interface-usb-xhci.pdf), adapted to our simplified setup with a single entry in the segment table.

## What does this PR do?
This PR adds support for wraparound behavior in the Event Ring, allowing it to reuse its only memory segment once the end is reached.
Specifically, it introduces the following logic:

- When the enqueue_pointer reaches the end of the segment, it wraps around to the start of the segment.
    At this point, the cycle_state is toggled, as required by the xHCI specification (section 4.9.4), so that hardware and software can correctly distinguish new entries from old ones.

- If the ring is full—meaning all TRB entries have been enqueued and not yet consumed (i.e., enqueue_pointer has wrapped and caught up with dequeue_pointer)—the following happens:

1. The enqueue function logs a warning indicating that no space is available.
2. The TRB is not written.
3. The function returns without advancing the enqueue pointer, effectively back-pressuring the caller.

- This implementation maintains correct tracking of:

1. enqueue_pointer position;
2. trb_count (number of entries in the ring);
3. cycle_state, which toggles on each full traversal of the segment.

### Test
A unit test event_ring_wraps_correctly was added to verify the new wraparound logic. It checks that:

- The ring correctly accepts TRBs up to its capacity.

- When full, further enqueue attempts are rejected with a warning.

- After advancing the dequeue pointer to simulate consumption, enqueue can resume from the start of the segment.
- The cycle_state toggles correctly upon wraparound.

## TODOs

While this PR implements basic wraparound support for the Event Ring, full handling of the ring full condition is not yet in place.
Currently, when the ring is full, a warning is logged but no actual flow control or error signaling is performed.
The following TODOs remain unimplemented and are marked in the code for future work:

- **Blocking or waiting logic on full ring**
In enqueue_or_wait, the software is expected to advance the ERDP and wait until space becomes available. This logic is not yet implemented.

- **Pausing Transfer and Command Ring processing**
In enqueue_at_tail, the system should ideally suspend further TRB processing until space is reclaimed. This coordination is currently missing.

- **Emitting an Event Ring Full Error Event TRB**
In enqueue_at_tail, the controller may emit a dedicated error TRB to report that the Event Ring has run out of space. This behavior is noted but not yet implemented.

### DIFF
--- a/src/device/pci/rings.rs
+++ b/src/device/pci/rings.rs
@@ -77,6 +77,19 @@ pub struct EventRing {
     /// event ring (i.e., in our case, when we move from the back of the
     /// segment to the front of the single segment).
     cycle_state: bool,
+    /// The base address of the first (and only) Event Ring segment.
+    ///
+    /// This value is derived from the Event Ring Segment Table during
+    /// controller initialization (ERSTBA write).
+    /// We use this address to wrap around the enqueue pointer once the
+    /// segment is full.
+    segment_base: u64,
+    /// The number of TRBs that the first (and only) segment can hold.
+    ///
+    /// The size is parsed from the Event Ring Segment Table at setup time.
+    /// When the TRB count reaches 0, the ring wraps to `segment_base`
+    /// and this value is used to reset `trb_count`.
+    segment_size: u32,
 }
 
 impl EventRing {
@@ -93,6 +106,8 @@ impl EventRing {
             enqueue_pointer: 0,
             trb_count: 0,
             cycle_state: false,
+            segment_base: 0,
+            segment_size: 0,
         }
     }
 
@@ -111,15 +126,17 @@ impl EventRing {
         assert_eq!(erstba & 0x3f, 0, "unaligned event ring base address");
 
         self.base_address = erstba;
-        self.enqueue_pointer = self.dma_bus.read(Request::new(
+        self.segment_base = self.dma_bus.read(Request::new(
             erstba.wrapping_add(BASE_ADDR),
             RequestSize::Size8,
         ));
-        self.trb_count = self
+        self.segment_size = self
             .dma_bus
             .read(Request::new(erstba.wrapping_add(SIZE), RequestSize::Size4))
             as u32;
         self.cycle_state = true;
+        self.enqueue_pointer = self.segment_base;
+        self.trb_count = self.segment_size;
 
         debug!("event ring segment table is at {:#x}", erstba);
         debug!(
@@ -152,40 +169,72 @@ impl EventRing {
         self.dequeue_pointer
     }
 
-    /// Enqueue an Event TRB to the ring.
-    ///
-    /// # Current Limitations
-    ///
-    /// The method is not capable of wrapping around to the start of the single
-    /// segment. We fail once the first segment is full
+    /// Enqueue a new Event TRB into the Ring.
     ///
     /// # Parameters
-    ///
     /// - `trb`: the TRB to enqueue.
+    ///
+    /// # Limitations
+    /// The current implementation does not handle ring-full recovery and will panic (`todo!()`) in that case.
     pub fn enqueue(&mut self, trb: &EventTrb) {
+        // TODO: Proper handling of full Event Ring
+        // According to xHCI §4.9.4, the xHC must:
+        //
+        // 1. Stop fetching new TRBs from the Transfer and Command Rings.
+        // 2. Emit an Event Ring Full Error Event TRB to the Event Ring (if supported).
+        // 3. Advance the Event Ring Enqueue Pointer (EREP) accordingly.
+        // 4. Wait for software (the host driver) to advance the Event Ring Dequeue Pointer (ERDP),
+        //    at which point normal event generation can resume.
         if self.check_event_ring_full() {
-            todo!();
+            todo!("The Event Ring is full!");
         }
 
         self.dma_bus
             .write_bulk(self.enqueue_pointer, &trb.to_bytes(self.cycle_state));
 
-        let enqueue_address = self.enqueue_pointer;
-
-        self.enqueue_pointer = self.enqueue_pointer.wrapping_add(TRB_SIZE as u64);
         self.trb_count -= 1;
 
         trace!(
-            "enqueued TRB in first segment of event ring at address {:#x}. Space for {} more TRBs left (TRB: {:?})",
-            enqueue_address, self.trb_count, trb
+            "enqueued TRB in first segment of event ring at address {:#x}. Space for {} more TRBs left in segment (TRB: {:?})",
+            self.enqueue_pointer, self.trb_count, trb
         );
+
+        self.advance_enqueue_pointer();
     }
 
-    // The method is currently not capable of dealing with wrapping around to
-    // the start of the single segment and just reports full once the segment
-    // is filled up.
+    /// Advances the enqueue pointer to the next slot in the event ring,
+    /// wrapping to the start when the end of the segment is reached.
+    fn advance_enqueue_pointer(&mut self) {
+        if self.trb_count == 0 {
+            self.wraparound();
+        } else {
+            self.enqueue_pointer = self.enqueue_pointer.wrapping_add(TRB_SIZE as u64);
+        }
+    }
+
+    /// Checks whether the Event Ring is full, based on xHCI §4.9.4.
+    ///
+    /// # Return
+    /// - `true` if the Event Ring is full and an Event Ring Full Error Event should be enqueued at the current position.
+    /// - `false` if there is at least one more slot available.
     const fn check_event_ring_full(&self) -> bool {
-        self.trb_count == 0
+        self.dequeue_pointer
+            == match self.trb_count {
+                1 => self.segment_base,
+                _ => self.enqueue_pointer.wrapping_add(TRB_SIZE as u64),
+            }
+    }
+
+    /// Wraps the Event Ring back to the segment base to start a new cycle.
+    fn wraparound(&mut self) {
+        self.enqueue_pointer = self.segment_base;
+        self.trb_count = self.segment_size;
+        self.cycle_state = !self.cycle_state;
+
+        trace!(
+            "Wrapped around event ring to base {:#x}, cycle_state flipped",
+            self.segment_base
+        );
     }
 }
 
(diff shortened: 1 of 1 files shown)


snue (changes requested):
Thanks for advancing our event ring handling. I left a bunch of comments. And I must look into the spec once more to double-check the special cases. I think we can make this more concise and hide the edge-cases from the main flow.

src/device/pci/rings.rs
```diff
@@ -152,16 +169,29 @@ impl EventRing {
...
+    ///
+    /// 1. Stop fetching new TRBs from the Transfer and Command Rings.
+    /// 2. Emit an Event Ring Full Error Event TRB to the Event Ring (if supported).
+    /// 3. Advance the Event Ring Enqueue Pointer (EREP) accordingly.
+    /// 4. Wait for software (the host driver) to advance the Event Ring Dequeue Pointer (ERDP),
+    ///    at which point normal event generation can resume.
     ///
-    /// - `trb`: the TRB to enqueue.
+    /// These behaviors are currently unimplemented. The logic should ensure that the
+    /// ring is not overwritten and that the xHC stalls until space is made available.
```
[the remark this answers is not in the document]
lbeierlieb: Helpful guidelines: https://doc.rust-lang.org/rustdoc/how-to-write-documentation.html

src/device/pci/rings.rs
```diff
@@ -179,13 +209,36 @@ impl EventRing {
             "enqueued TRB in first segment of event ring at address {:#x}. Space for {} more TRBs left (TRB: {:?})",
             enqueue_address, self.trb_count, trb
         );
+
+        if self.trb_count == 0 {
+            self.enqueue_pointer = self.segment_base;
+            self.trb_count = self.segment_size;
+            self.cycle_state = !self.cycle_state;
```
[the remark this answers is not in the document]
ziyifu225: Thanks for the suggestion! I gave it some thought. In our current logic, the `self.trb_count == 0` check must happen after the TRB has been written and the count decremented, so if we use an if/else structure, it would need to be phrased as `if self.trb_count != 0`.
```
    pub fn enqueue(&mut self, trb: &EventTrb) {
        if self.check_event_ring_full() {
            todo!("The Event Ring is full!");
        }

        if self.trb_count != 0 {
            self.dma_bus
                .write_bulk(self.enqueue_pointer, &trb.to_bytes(self.cycle_state));

            let enqueue_address = self.enqueue_pointer;

            self.enqueue_pointer = self.enqueue_pointer.wrapping_add(TRB_SIZE as u64);
            self.trb_count -= 1;

            trace!(
                "enqueued TRB in first segment of event ring at address {:#x}. Space for {} more TRBs left in segment (TRB: {:?})",
                enqueue_address, self.trb_count, trb
            );
        } else {
            self.wraparound();
        }
    }

    fn wraparound(&mut self) {
        self.enqueue_pointer = self.segment_base;
        self.trb_count = self.segment_size;
        self.cycle_state = !self.cycle_state;
    
        trace!("Wrapped around event ring to base {:#x}, cycle_state flipped", self.segment_base);
    }
```
While that does work, I find it a bit less intuitive to read—especially compared to the explicit `if self.trb_count == 0` branch, which clearly marks the wraparound logic. I’m leaning toward keeping the current structure for clarity.

Happy to adjust if you think the `if/else` structure brings a clearer flow.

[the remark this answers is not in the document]
snue: Yes, that's what I had in mind. With the if-else moving to an "advance_enqueue_pointer" function (possibly later) that would eventually handle jumping from segment to segment (and not full wrap around) as well.

You can remove the `enqueue_address` with the logic above (and use `self.enqueue_pointer` in the `trace` as it has not moved yet). In the end there does not need to be any branch left directly in the enqueue function.

I prefer this to touching the enqueue pointer twice, as with the previous logic you have an interim invalid (out of bounds) enqueue pointer value when you should have wrapped or moved to the next segment.

src/device/pci/rings.rs
```diff
@@ -652,6 +704,90 @@ mod tests {
...
+    #[test]
+    fn event_ring_start_empty_enqueue_fill_then_wraparound_after_dequeue_pointer_move() {
+        let erste = [
+            // segment_base = 0x1000
+            // trb_count = 4
+            0x00, 0x10, 0x00, 0x00, 0x00, 0x00, 0x00, 0x00, 0x04, 0x00, 0x00, 0x00, 0x00, 0x00,
+            0x00, 0x00,
+        ];
+
+        let ram = Arc::new(TestingRamDevice::new(&vec![0; 0x2000]));
```
lbeierlieb, line 733: What's the reason for using "so much" memory? It's not so much that it is expensive, but a lot more than would be necessary, so I am wondering what the reason is.

src/device/pci/rings.rs
```diff
@@ -652,6 +704,90 @@ mod tests {
...
+        ram.write_bulk(0x0, &erste);
+        let mut ring = EventRing::new(ram.clone());
+        ring.configure(0x0);
+        ring.update_dequeue_pointer(ring.segment_base);
```
lbeierlieb, line 771: the initialization is exactly the same as is in the test above. Please move the code to a function like
```
fn init_ram_and_ring() -> (Arc<TestingRamDevice>, EventRing)
```

src/device/pci/rings.rs
```diff
@@ -652,6 +704,90 @@ mod tests {
...
+            1,                       // endpoint_id
+            1,                       // slot_id
+        )
+    }
+
+    fn assert_trb_written(ram: &TestingRamDevice, addr: u64) {
+        let mut buf = [0u8; 16];
+        ram.read_bulk(addr, &mut buf);
+        assert_ne!(buf, [0u8; 16], "TRB not written at address {:#x}", addr);
+    }
```
lbeierlieb, line 722: You should check the cycle bit instead of the non-zero check. Your approach works fine with zero-initialized memory and no wrap-around. After the wrap-around, you enqueue the dummy TRB, which overwrites the old dummy TRB: The only effect is that the cycle bit changes from 1 to 0. Your assertion succeeds because you see the completion code and ids, but it would also succeed without the wrap-around write. The only way to check is to confirm that after the wrap-around write the cycle bit was set to 0. 
```suggestion
    fn assert_trb_written(ram: &TestingRamDevice, addr: u64, cycle_state: bool) {
        let mut buf = [0u8; 16];
        ram.read_bulk(addr, &mut buf);
        let cycle_bit = buf[12] & 0x1 != 0;
        assert_eq(cycle_bit, cycle_state, "TRB not written at address {:#x}", addr);
    }
```

src/device/pci/rings.rs
```diff
@@ -652,6 +704,90 @@ mod tests {
...
+        ram.write_bulk(0x0, &erste);
+        let mut ring = EventRing::new(ram.clone());
+        ring.configure(0x0);
+        ring.update_dequeue_pointer(ring.segment_base);
+
+        ring.enqueue(&dummy_trb()); // TRB 1
+        ring.enqueue(&dummy_trb()); // TRB 2
+        ring.enqueue(&dummy_trb()); // TRB 3
+
+        assert_trb_written(&ram, ring.segment_base);
```
lbeierlieb, line 743: Your reference value should not come from the instance that you want to test. `EventRing` could set the value differently then you expect and pass the assertion with incorrect behavior. Please you `0x1000` or introduce a constant for the value at the start of the function instead of using `ring.segment_base`.

src/device/pci/rings.rs
lbeierlieb: Just saw below that you don't actually check if an TRB was written after wrap-around. I think you should, so this change is still relevant.

src/device/pci/rings.rs
```diff
@@ -652,6 +704,90 @@ mod tests {
...
+        assert_trb_written(&ram, ring.segment_base + 32);
+
+        ring.update_dequeue_pointer(ring.segment_base + 16);
+
+        ring.enqueue(&dummy_trb()); // TRB 4
+        assert_trb_written(&ram, ring.segment_base + 48);
+
+        assert_eq!(ring.enqueue_pointer, ring.segment_base);
+        assert_eq!(ring.trb_count, ring.segment_size);
+        assert!(!ring.cycle_state);
```
lbeierlieb, line 754: Those struct fields are implementation details that we don't really want to test. Instead, check the behavior of the implementation by enqueuing another TRB and look for its correct appearance at the start of the segment (see also suggested changes for the `assert_trb_written` function.

Your current checks have to be true to make the enqueue succeed, and the behavior-based check allows changes to the implementation without adapting the test.

lbeierlieb (changes requested):
I added a few comments and suggestions to improve the tests.

src/device/pci/rings.rs
ziyifu225: I used 0x2000 here just to have a safe amount of memory.
But since it only really needs 0x1040, would it be better to change it to that?

snue: 0x2000 is two pages of memory (8KiB). You place a 16 byte segment table entry and 4 * 16 byte TRBs. That's 80 bytes actually used (a factor of 100x). Lukas was asking why you need so much padding (zeroes/unused memory) between segment table and ring structure.

The test should probably try to only use as much memory as is actually required/touched. This makes it more likely that issues with the implementation will cause an invalid memory access that the test can catch. You could fit everything within 0x50 bytes.

ziyifu225: Reduce the memory as suggested: segment_base at 0x10 and everything within 0x50 bytes.

snue (approved):
LGTM. Thanks. You can squash the two commits before merge, but the code/tests look good now.

PR #78, opened 2025-09-17
Allow enabling of any bulk endpoint

lbeierlieb:
So far, we only supported endpoint 3 (EP1 IN) and 4 (EP2 OUT), because these two were used by all USB 3 flashdrives that we tested.
This PR implements more generic handling of the endpoints of a nusb device, because we already observed a USB 2 flashdrive using endpoints 2 (EP1 OUT) and 5 (EP2 IN).

The scenario in the CI test runs unchanged, but we should support any kind of USB 3, bulk-only device now (not tested; I cherry-picked the two commits from this PR on my prototype branch for the USB 2 flashdrive, and the implementation handled the endpoints 2 and 5 well).

I am leaving some review comments myself where I am unsure if another implementation would be nicer.

### DIFF
--- a/src/device/pci/nusb.rs
+++ b/src/device/pci/nusb.rs
@@ -14,11 +14,15 @@ use std::{
     time::Duration,
 };
 
+enum EndpointWrapper {
+    BulkIn(nusb::Endpoint<Bulk, In>),
+    BulkOut(nusb::Endpoint<Bulk, Out>),
+}
+
 pub struct NusbDeviceWrapper {
     device: nusb::Device,
     interface: nusb::Interface,
-    ep_in: Option<nusb::Endpoint<Bulk, In>>,
-    ep_out: Option<nusb::Endpoint<Bulk, Out>>,
+    endpoints: [Option<EndpointWrapper>; 30],
 }
 
 impl Debug for NusbDeviceWrapper {
@@ -38,8 +42,7 @@ impl NusbDeviceWrapper {
         Self {
             device,
             interface,
-            ep_in: None,
-            ep_out: None,
+            endpoints: std::array::from_fn(|_| None),
         }
     }
 
@@ -115,7 +118,7 @@ impl RealDevice for NusbDeviceWrapper {
 
     fn transfer_out(
         &mut self,
-        _endpoint_id: u8,
+        endpoint_id: u8,
         trb: &TransferTrb,
         dma_bus: &BusDeviceRef,
     ) -> (CompletionCode, u32) {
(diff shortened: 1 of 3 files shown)


src/device/pci/nusb.rs
```diff
@@ -14,11 +14,15 @@ use std::{
...
+    BulkIn(nusb::Endpoint<Bulk, In>),
+    BulkOut(nusb::Endpoint<Bulk, Out>),
+}
+
 pub struct NusbDeviceWrapper {
     device: nusb::Device,
     interface: nusb::Interface,
-    ep_in: Option<nusb::Endpoint<Bulk, In>>,
-    ep_out: Option<nusb::Endpoint<Bulk, Out>>,
+    endpoints: [Option<EndpointWrapper>; 30],
```
lbeierlieb, line 25: We have to handle up to 30 different endpoints, with the endpoint ids ranging from 2 to 31 (id 1 is the control endpoint, we don't need to track it). The current implementation tracks exactly the 30 possible endpoints, but due to the starting-at-2 id there are always offset calculations involved (at the few locations that access the endpoints).

Alternative ideas:
- use array with size of 32. No more offset calculations, but two array indices that could hold information while they should not (going against "make impossible state unrepresentable")
- use a map collection. again going against "make impossible state unrepresentable", slower than array access.

src/device/pci/nusb.rs
```diff
@@ -191,24 +212,42 @@ impl RealDevice for NusbDeviceWrapper {
...
                 }
-                // Program requires USB bulk out endpoint 0x2
-                self.ep_out = Some(self.interface.endpoint::<Bulk, Out>(0x2).unwrap());
-                debug!("enabled EP4 on real device");
+                self.endpoints[id as usize - 2] = Some(EndpointWrapper::BulkOut(
+                    self.interface.endpoint::<Bulk, Out>(id / 2).unwrap(),
+                ));
+                debug!("enabled EP{} on real device", id);
+            }
+            _ => {
```
lbeierlieb, line 246: My intention here was to have each case clearly identified. Because of the modulo operation, the compiler cannot tell that all cases are covered.

Alternativ: make `id if id % 2 == 0` the default case and perhaps highlight with a comment that this case covers all the even numbers between 2 and 31.

snue: I find it a bit too much duplication with the separate cases. You only determine whether to add an In or Out endpoint wrapper based on the last bit. Can you just do that within one match arm for the "valid id" case?

lbeierlieb: I stuck too much with adapting the previous fixed-endpoint version (and did not give too much thought because we will touch this function soon in the context of interrupt endpoints).

I rewrote the function. What do you think?

snue (approved):

PR #79, opened 2025-09-17
Move us forward in the roadmap

lbeierlieb:
The readme has been out of date for a while. Indicate that we have working USB storage and are now working on broadening the device support.


snue:
Thanks! We also have an outdated note in there for attaching devices that I would like to see removed:

> [!NOTE]
> Attached USB devices do not yet appear in the guest. The relevant plumbing
> is not implemented yet.

PR #89, opened 2025-09-25
Identify endpoint types

lbeierlieb:
So far, we assumed that everything except the control endpoint EP0 is a bulk endpoint. We want to support InterruptIn endpoints soon, so we need to be able to distinguish the configured endpoint type, so we can access the endpoints on the real device correctly. This PR implements the functionality to extract the endpoint-type information from the input context in the Configure Endpoint Command and to provide the information to the `enable_endpoint` function of the `realdevice` trait, where it will be needed to enable the correct endpoints on the real USB device.

This PR does not change the nusb implementation of `realdevice`, because the `enable_endpoint` function will receive substantial rework to make the existing bulk-endpoint handling asynchronous; making changes to handle interrupt endpoints does not make sense yet.

### DIFF
--- a/src/device/pci/device_slots.rs
+++ b/src/device/pci/device_slots.rs
@@ -9,7 +9,9 @@ use crate::device::{
     pci::constants::xhci::device_slots::{endpoint_state, slot_state},
 };
 
-use super::{constants::xhci::device_slots::endpoint_state::*, rings::TransferRing};
+use super::{
+    constants::xhci::device_slots::endpoint_state::*, realdevice::EndpointType, rings::TransferRing,
+};
 
 /// Abstraction for Device Slots.
 ///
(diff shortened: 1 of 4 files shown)


src/device/pci/device_slots.rs
```diff
@@ -236,16 +238,27 @@ impl DeviceContext {
...
                 self.address.wrapping_add(ep_context_offset as u64),
                 &input_context[ep_context_offset..ep_context_offset + 32],
             );
+
+            let ep_type = match (input_context[ep_context_offset + 4] >> 3) & 0x7 {
+                2 => EndpointType::BulkOut,
+                6 => EndpointType::BulkIn,
+                4 => EndpointType::Control,
+                7 => EndpointType::InterruptIn,
+                val => todo!("encountered unsupported endpoint type: {}", val),
```
snue, line 255: Nice! This will mean we crash loudly with unsupported devices (endpoints), but that's the expected behavior for now. I just hope it doesn't surprise us in testing eventually.

snue (approved):

PR #91, opened 2025-10-08
Add ssh access when using interactive vm tests driver

Noi0103:
This will add openssh service configurations to the qemu host system and the cloud-hypervisor guest system that allow to connect over ssh for interactive debugging of vm tests.

### DIFF
--- a/docs/developers/vmTest.md
+++ b/docs/developers/vmTest.md
@@ -1,0 +1,34 @@
+# Overview
+We provide an integration test that runs in CI and in an interactive local session.
+
+> [!NOTE]
+> An overview about nix vm tests is helpful and can be found here: https://nix.dev/tutorials/nixos/integration-testing-using-virtual-machines.html
+
+The integration test includes a static port forward of SSH to the developers machine port 2000. This provides a `root` login to the virtual machine with empty password.
+```mermaid
+graph LR;
+  id1[Developer Machine]
+  id2[QEMU Host]
+  id3[Cloud Hypervisor Guest]
+  id1--ssh root@localhost:2000--->id2
+  id2--ssh root@192.168.100.2--->id3
+```
+
+
+## Quick Start
+To build and run the interactive vm test (build logs flag is optional):
+```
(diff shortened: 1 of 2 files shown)


docs/developers/vmTest.md
```diff
@@ -0,0 +1,19 @@
+# Quick Start
+Note: A overview about nix vm test is helpful and can be found here: https://nix.dev/tutorials/nixos/integration-testing-using-virtual-machines.html
```
lbeierlieb, line 2:
```suggestion
[!NOTE]
A overview about nix vm test is helpful and can be found here: https://nix.dev/tutorials/nixos/integration-testing-using-virtual-machines.html
```

nix/tests.nix
```diff
@@ -160,12 +186,26 @@ in
                 --kernel ${netboot.kernel} \
                 --cmdline ${lib.escapeShellArg netboot.cmdline} \
                 --initramfs ${netboot.initrd} \
-                --user-device socket=${usbvfiodSocket}
+                --user-device socket=${usbvfiodSocket} \
+                --net "tap=tap0,mac=,ip=192.168.100.1,mask=255.255.255.0"
```
lbeierlieb, line 190: We should mention that we use the 192.168.100.x subnet in the documention, to avoid confusion when by chance somebody uses the same subnet on a different interface and packets get misrouted.

nix/tests.nix
```diff
@@ -160,12 +186,26 @@ in
...
+      services.openssh = {
+        enable = true;
+        settings = {
+          PermitRootLogin = "yes";
+          PermitEmptyPasswords = "yes";
+        };
+      };
+      security.pam.services.sshd.allowNullPassword = true;
+      virtualisation.forwardPorts = [
+        { from = "host"; host.port = 2000; guest.port = 22; }
```
lbeierlieb, line 206: Same for the port, we should make it easy to see from the documentation which port is used to avoid potential port collisions and user confusion.

PR #176, opened 2026-01-29
Add PCAP logging option

ziyifu225:
As the project has grown more complex, it has become increasingly hard to pinpoint issues by manually scanning logs line by line. Adding PCAP capture support allows us to record USB transfer traffic in a parseable format and use tools like Wireshark to filter and isolate the exact signals we care about, making it much faster to locate relevant events and diagnose problems.
All PCAP record structures follow the official USB PCAP specification.([PCAP Capture File Format](https://datatracker.ietf.org/doc/id/draft-gharris-opsawg-pcap-00.html#name-packet-record) and [Packet structure](https://www.tcpdump.org/linktypes/LINKTYPE_USB_LINUX.html))

# Design
## Structure
The implementation of PCAP packets' structure is split into three parts:

- **meta.rs** defines the USB metadata attached to each recorded event and keeps call sites independent from PCAP layout details.

- **packet.rs** builds PCAP records and Linux USB link-layer headers from metadata and payload bytes, and handles writing the resulting bytes to the output file.

- **mod.rs** provides the public entry points that connect transport-side USB events to the serialization layer without exposing packet-format internals upstream.

## Bus number and Device address in packet
The capture output is meant to describe USB traffic as seen by this virtual xHCI path, not to reproduce host-side topology identifiers. If host-side bus and device numbering is needed for debugging, that can be captured separately on the host with usbmon.

The recorded bus number now represents the USB generation of the attached device, using 2 for USB 2.x and 3 for USB 3.x. The recorded device number now uses the slot id assigned by the virtual xHCI controller. This makes the dump better match the virtual device context carried through this stack.

## Error mapping
In the current implementation, transfer-side failures are grouped into three categories: Disconnect, Stall, and TransactionError. The PCAP status mapping follows Linux USB error semantics so the recorded error values stay close to the meaning that host-side tooling and developers would normally expect.

Disconnect is mapped to -ENODEV because it indicates that the target device is no longer present while the transfer is being processed. Stall is mapped to -EPIPE because a USB STALL reflects an endpoint-side reject or halt condition, which Linux conventionally reports as EPIPE for USB transfers. TransactionError is mapped to -EPROTO because it represents a protocol- or transaction-level transfer failure rather than an explicit device removal or endpoint stall.

This mapping keeps the PCAP error records readable and consistent with common Linux USB debugging conventions, while still reflecting the transfer error model used internally by the project.

# Usage
Add PCAP option `--pcap-path` when startup `usbvfiod`. The `PCAP` file will be stored in target path.

An example Wireshark view of a captured PCAP is shown in the figure below.
image

# Limitation
Isochronous transfers are not implemented in the project yet, so PCAP currently captures only control, bulk, and interrupt traffic.

### DIFF
--- a/src/device/pcap/packet.rs
+++ b/src/device/pcap/packet.rs
@@ -1,0 +1,434 @@
+use std::fs::{self, File};
+use std::io::{BufWriter, Write};
+use std::path::PathBuf;
+use std::sync::{Mutex, OnceLock};
+use std::time::SystemTime;
+
+use crate::device::xhci::usbrequest::UsbRequest;
+use tracing::warn;
+
+const LINKTYPE_USB_LINUX: u32 = 189;
+const PCAP_MAGIC: u32 = 0xa1b2c3d4;
+const PCAP_MAJOR: u16 = 2;
+const PCAP_MINOR: u16 = 4;
+const SNAPLEN: u32 = 65_535;
+
+/// USB event type stored as an ASCII code in the packet header.
+#[derive(Clone, Copy)]
+pub enum UsbEventType {
+    Submission,
+    Completion,
+    Error,
+}
+
+impl UsbEventType {
+    const fn code(self) -> u8 {
+        match self {
+            Self::Submission => b'S',
+            Self::Completion => b'C',
+            Self::Error => b'E',
+        }
+    }
+}
+
+/// USB transfer type stored in the linktype header.
+#[derive(Clone, Copy, Debug)]
+pub enum UsbTransferType {
+    // TODO: implement isochronous transfer logging
+    // Isochronous,
+    Control,
+    Bulk,
+    Interrupt,
+}
+
+impl UsbTransferType {
+    const fn code(self) -> u8 {
+        match self {
+            // TODO: implement isochronous transfer logging
+            // Self::Isochronous => 0,
+            Self::Interrupt => 1,
+            Self::Control => 2,
+            Self::Bulk => 3,
+        }
+    }
+}
+
+/// USB direction used to set the endpoint address IN/OUT bit.
+#[derive(Clone, Copy, Debug)]
+pub enum UsbDirection {
+    HostToDevice,
+    DeviceToHost,
+}
+
+impl UsbDirection {
+    const fn endpoint_address(self, xhci_endpoint_id: u8) -> u8 {
+        let endpoint_number = xhci_endpoint_id / 2;
+        match self {
+            Self::HostToDevice => endpoint_number,
+            Self::DeviceToHost => endpoint_number | 0x80,
+        }
+    }
+}
+
+/// Packet timestamp in seconds and microseconds.
+#[derive(Clone, Copy, Debug, Default, PartialEq, Eq)]
+pub struct Timestamp {
+    pub seconds: u32,
+    pub microseconds: u32,
+}
+
+impl From<SystemTime> for Timestamp {
+    fn from(value: SystemTime) -> Self {
+        let duration = value
+            .duration_since(SystemTime::UNIX_EPOCH)
+            .unwrap_or_default();
+        Self {
+            seconds: duration.as_secs() as u32,
+            microseconds: duration.subsec_micros(),
+        }
+    }
+}
+
+/// Linux USB per-packet header for PCAP linktype 189.
+///
+/// `header_bytes` writes the fields in little-endian order.
+pub struct UsbPacketLinktypeHeader {
+    pub id: u64,
+    pub event_type: u8,
+    pub transfer_type: u8,
+    pub endpoint_address: u8,
+    pub device_address: u8,
+    pub bus_number: u16,
+    pub setup_flag: u8,
+    pub data_flag: u8,
+    pub status: i32,
+    pub data_length: u32,
+    pub delivered_data_length: u32,
+    pub setup: [u8; 8],
+}
+
+impl UsbPacketLinktypeHeader {
+    pub fn header_bytes(&self, event_timestamp: Timestamp) -> [u8; 48] {
+        let mut header = [0u8; 48];
+        header[0..8].copy_from_slice(&self.id.to_le_bytes());
+        header[8] = self.event_type;
+        header[9] = self.transfer_type;
+        header[10] = self.endpoint_address;
+        header[11] = self.device_address;
+        header[12..14].copy_from_slice(&self.bus_number.to_le_bytes());
+        header[14] = self.setup_flag;
+        header[15] = self.data_flag;
+        header[16..24].copy_from_slice(&(event_timestamp.seconds as i64).to_le_bytes());
+        header[24..28].copy_from_slice(&(event_timestamp.microseconds as i32).to_le_bytes());
+        header[28..32].copy_from_slice(&self.status.to_le_bytes());
+        header[32..36].copy_from_slice(&self.data_length.to_le_bytes());
+        header[36..40].copy_from_slice(&self.delivered_data_length.to_le_bytes());
+        header[40..48].copy_from_slice(&self.setup);
+        header
+    }
+}
+
+/// Builds the fixed PCAP file header.
+pub fn pcap_global_header_bytes() -> [u8; 24] {
+    let mut header = [0u8; 24];
+    header[0..4].copy_from_slice(&PCAP_MAGIC.to_le_bytes());
+    header[4..6].copy_from_slice(&PCAP_MAJOR.to_le_bytes());
+    header[6..8].copy_from_slice(&PCAP_MINOR.to_le_bytes());
+    header[8..12].copy_from_slice(&0u32.to_le_bytes());
+    header[12..16].copy_from_slice(&0u32.to_le_bytes());
+    header[16..20].copy_from_slice(&SNAPLEN.to_le_bytes());
+    header[20..24].copy_from_slice(&LINKTYPE_USB_LINUX.to_le_bytes());
+    header
+}
+
+/// Builds one PCAP record from the record header, linktype header, and payload.
+pub fn pcap_record_bytes(
+    capture_timestamp: Timestamp,
+    event_timestamp: Timestamp,
+    meta: &UsbPacketLinktypeHeader,
+    payload: &[u8],
+) -> Vec<u8> {
+    let link_header = meta.header_bytes(event_timestamp);
+    let original_len = link_header.len() + payload.len();
+    let captured_len = original_len.min(SNAPLEN as usize);
+    let mut record = Vec::with_capacity(16 + captured_len);
+    record.extend_from_slice(&capture_timestamp.seconds.to_le_bytes());
+    record.extend_from_slice(&capture_timestamp.microseconds.to_le_bytes());
+    record.extend_from_slice(&(captured_len as u32).to_le_bytes());
+    record.extend_from_slice(&(original_len as u32).to_le_bytes());
+
+    if captured_len <= link_header.len() {
+        record.extend_from_slice(&link_header[..captured_len]);
+        return record;
+    }
+
+    record.extend_from_slice(&link_header);
+    let remaining = captured_len - link_header.len();
+    record.extend_from_slice(&payload[..remaining]);
+    record
+}
+
+/// Lazily opens the PCAP file and writes the global header once.
+///
+/// I/O failures disable PCAP logging and emit a warning.
+pub struct PcapManager {
+    path: Option<PathBuf>,
+    writer: Option<BufWriter<File>>,
+    warned: bool,
+}
+
+impl PcapManager {
+    pub const fn new(path: Option<PathBuf>) -> Self {
+        Self {
+            path,
+            writer: None,
+            warned: false,
+        }
+    }
+
+    fn ensure_writer(&mut self) -> Option<&mut BufWriter<File>> {
+        let file_path = self.path.clone()?;
+
+        if self.writer.is_some() {
+            return self.writer.as_mut();
+        }
+
+        if let Some(parent) = file_path.parent() {
+            if let Err(error) = fs::create_dir_all(parent) {
+                if !self.warned {
+                    warn!(
+                        "Disabling USB PCAP logging after failing to create {}: {}",
+                        parent.display(),
+                        error
+                    );
+                    self.warned = true;
+                }
+                self.path = None;
+                return None;
+            }
+        }
+
+        let mut writer = match File::create(&file_path).map(BufWriter::new) {
+            Ok(writer) => writer,
+            Err(error) => {
+                if !self.warned {
+                    warn!(
+                        "Disabling USB PCAP logging after failing to open {}: {}",
+                        file_path.display(),
+                        error
+                    );
+                    self.warned = true;
+                }
+                self.path = None;
+                return None;
+            }
+        };
+
+        if let Err(error) = writer.write_all(&pcap_global_header_bytes()) {
+            if !self.warned {
+                warn!(
+                    "Disabling USB PCAP logging after failing to write header to {}: {}",
+                    file_path.display(),
+                    error
+                );
+                self.warned = true;
+            }
+            self.path = None;
+            return None;
+        }
+
+        self.writer = Some(writer);
+        self.writer.as_mut()
+    }
+
+    pub fn write_record(&mut self, record: &[u8]) {
+        let writer = match self.ensure_writer() {
+            Some(writer) => writer,
+            None => return,
+        };
+
+        if let Err(error) = writer.write_all(record).and_then(|_| writer.flush()) {
+            if !self.warned {
+                warn!("Failed to write USB PCAP record: {}", error);
+                self.warned = true;
+            }
+            self.path = None;
+            self.writer = None;
+        }
+    }
+}
+
+static MANAGER: OnceLock<Mutex<PcapManager>> = OnceLock::new();
+
+/// Process-wide entry point for optional USB PCAP logging.
+pub struct UsbPcapManager;
+
+impl UsbPcapManager {
+    pub fn init(path: Option<PathBuf>) {
+        if path.is_some() {
+            let _ = MANAGER.set(Mutex::new(PcapManager::new(path)));
+        }
+    }
+
+    pub fn write_record(record: &[u8]) {
+        if let Some(manager) = MANAGER.get() {
+            manager.lock().unwrap().write_record(record);
+        }
+    }
+}
+
+/// Emit a PCAP record for a control transfer submission event.
+pub fn log_control_submission(
+    meta: super::meta::EndpointPcapMeta,
+    request: &UsbRequest,
+    payload: &[u8],
+) {
+    log_submission(
+        meta,
+        request.address,
+        u32::from(request.length),
+        Some(build_setup_bytes(request)),
+        payload,
+    );
+}
+
+/// Emit a PCAP record for a control transfer completion event
+pub fn log_control_completion(
+    meta: super::meta::EndpointPcapMeta,
+    urb_id: u64,
+    status: i32,
+    actual_length: u32,
+    payload: &[u8],
+) {
+    log_completion(meta, urb_id, status, actual_length, payload);
+}
+
+/// Emit a PCAP record for an error-related event with optional setup data.
+pub fn log_error(
+    meta: super::meta::EndpointPcapMeta,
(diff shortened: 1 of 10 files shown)


src/device/pci/error_map.rs
```diff
@@ -0,0 +1,31 @@
+use std::error::Error;
+
+/// Map an error chain into a USB PCAP status value.
```
snue: As this is strictly for the PCAP error status and only used for feeding it into the `pcap::log_*` calls, I think moving it into the `pcap.rs` module is better.

src/device/pci/pcap.rs
```diff
@@ -0,0 +1,488 @@
...
+        header[40..48].copy_from_slice(&self.setup);
+        header
+    }
+}
+
+/// Build the PCAP global header bytes.
+///
+/// This is the fixed header written once at the start of every PCAP file.
+/// The global header layout follows [the official PCAP spec](https://datatracker.ietf.org/doc/id/draft-gharris-opsawg-pcap-00.html#name-file-header);
+/// detailed field descriptions are not repeated here.
```
Noi0103: nit: I would like to have this "we refer to this official documentation" in a similar pattern like we already do.
> /// For documentation of the fields other than `address`, see Section "9.3 USB
> /// Device Requests" in the USB 2.0 specification.

snue (commented):
Thanks for the draft. Of course some rework will be required once the worker state machine is done. I also left suggestions and some questions. For some things we can likely find a more idiomatic rust expression.

src/device/xhci/endpoint_handle.rs
```diff
@@ -408,6 +448,14 @@ impl<ROEH: RealOutEndpointHandle> EndpointHandle for OutEndpointHandle<ROEH> {
             TransferTrbVariant::Normal(normal_data) => {
                 let mut data = vec![0; normal_data.transfer_length as usize];
                 self.dma_bus.read_bulk(normal_data.data_pointer, &mut data);
+                let event_timestamp = Timestamp::from(SystemTime::now());
```
snue: Does it make sense to take a timestamp in the various functions instead of passing it in? Is there a case where you want to use a different timestamp than "now" (when the pcap logging function is called)?

ziyifu225: You're right. I was grabbing the timestamp at each call site to keep it as close as possible to when the event happens, but that does make the code pretty repetitive. Moving it into the helper functions makes things cleaner and is good for how the logging works.

src/device/pcap/packet.rs
```diff
@@ -0,0 +1,488 @@
...
+impl PcapManager {
+    pub const fn new(path: Option<PathBuf>) -> Self {
+        Self {
+            path,
+            writer: None,
+            warned: false,
+        }
+    }
+
+    fn ensure_writer(&mut self) -> Option<&mut BufWriter<File>> {
```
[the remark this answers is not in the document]
ziyifu225: I wrote it this way mainly to keep PCAP from impacting the main process. Right now I’m treating it as best-effort debug logging, so if PCAP isn’t enabled, or if writer setup fails, I don’t want that to break USB handling.

That said, I’m still not completely sure whether PCAP failures should become fatal once PCAP is explicitly enabled. If we want enabled PCAP to be a hard requirement, returning a Result would be cleaner. If we want it to stay auxiliary, then keeping the failure local and letting the process continue seems more appropriate.

src/device/xhci/endpoint_launcher.rs
```diff
@@ -178,22 +244,34 @@ impl<CRD: CompleteRealDevice> EndpointLauncher<CRD> {
...
 
+    fn pcap_usb_type(device: &CRD) -> u16 {
+        match device.realdevice_ref().speed() {
+            Some(speed) if speed.is_usb2_speed() => 2,
+            Some(_) => 3,
+            None => 0,
+        }
+    }
+
+    #[allow(clippy::too_many_arguments)]
```
lbeierlieb: many of the arguments are the same for every invocation. Perhaps it makes sense to use a struct for the launch arguments. Initialize the always-the-same fields (such as the slot and endpoint id) once, later set the endpoint-type-dependent fields (such as the real endpoint) and invoke the function.

Not necessary for this PR, though.

src/device/xhci/endpoint_launcher.rs
```diff
@@ -178,22 +244,34 @@ impl<CRD: CompleteRealDevice> EndpointLauncher<CRD> {
             .ok_or_else(|| anyhow!("channel should never close"))
     }
 
+    fn pcap_usb_type(device: &CRD) -> u16 {
+        match device.realdevice_ref().speed() {
+            Some(speed) if speed.is_usb2_speed() => 2,
+            Some(_) => 3,
+            None => 0,
+        }
+    }
```
lbeierlieb: This function feels out of place here. You call it multiple types, but actually calling once is enough (before the match (you call the function in every match case right now); so I suggest inlining this function into the run_loop before the match statement.

src/device/xhci/endpoint_handle.rs
```diff
@@ -476,17 +490,45 @@ impl<ROEH: RealOutEndpointHandle> EndpointHandle for OutEndpointHandle<ROEH> {
...
+                                )
+                            }
                             OutTrbProcessingResult::Stall => {
+                                let event_timestamp = Timestamp::from(SystemTime::now());
+                                pcap::out_error(
+                                    self.pcap_meta,
+                                    transfer_trb.address,
+                                    &OutTrbProcessingResult::Stall,
+                                    &[],
+                                    event_timestamp,
```
lbeierlieb: I don't like how noisy and repeating the code is here. It should be fine for now and I see little benefit to improving the code structure, considering @Noi0103 is working to replace the endpoint handles (so that they can handle event data TRBs).

src/device/pcap/packet.rs
```diff
@@ -0,0 +1,488 @@
...
+///
+/// This provides a single synchronized entry point for PCAP output so callers
+/// do not need to store writers themselves. The manager is optional to allow
+/// USB PCAP logging to be enabled or disabled at runtime.
+pub struct UsbPcapManager;
+
+impl UsbPcapManager {
+    pub fn init(path: Option<PathBuf>) {
+        *MANAGER.lock().unwrap() = Some(PcapManager::new(path));
+    }
```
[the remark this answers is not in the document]
ziyifu225: The main reason I kept this structure is that the PCAP state is decided from the CLI arguments during startup, and main already has a dedicated initialization step for that. main reads the PCAP path once and passes it into UsbPcapManager::init(path) at process startup, so using OnceLock fits that “set it once during initialization, then only read it later” flow quite naturally.

The file itself is still created only when the first record is actually written, but the output path and enabled/disabled state are both already determined at startup.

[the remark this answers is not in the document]
snue: Yes, `OnceLock` does solve my initial complaints. It's a more generic `LazyLock` for when you need to pass arguments (like the `path`) to the constructor. Not initializing it should also naturally avoid the lock taking overhead when no pcap dumps are created.

src/device/pcap/packet.rs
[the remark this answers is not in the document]
snue: Good question. We should do what is the least surprising/impacting for the user later. I feel pcap logging issues should never kill the controller operation. So handling all errors gracefully and just logging the issue is probably the way to go. But we should also avoid spamming the logs with errors for every packet once the logging fails. Is there a convenient `warn_once` wrapper in Rust?

So, using `expect()` on the result should be avoided, but using an explicit `Result<>` type to better relay information about the encountered error is useful if the function does not log errors itself. We wouldn't want to add explicit error logging in multiple call sites.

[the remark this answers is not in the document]
snue: Looking at the code, your current handling matches what I had in mind :smile:

src/device/pcap/packet.rs
```diff
@@ -0,0 +1,461 @@
...
+const fn data_flag_value(event: UsbEventType, payload_len: usize) -> u8 {
+    if payload_len != 0 {
+        0
+    } else {
+        match event {
+            UsbEventType::Submission => b'<',
+            UsbEventType::Completion => b'>',
+            UsbEventType::Error => b'E',
+        }
+    }
```
snue: Can you give a reference for this usage? Does this match `usbmon` capturing? I don't fully understand why the flag gets so abused. Also, shouldn't the "direction" (`<`/`>`) rather reflect whether this is an IN or OUT request than whether it is a submission/completion?

ziyifu225: In Usbmon, this part of data flag implementation is here: https://github.com/torvalds/linux/blob/master/drivers/usb/mon/mon_bin.c#L538C2-L551C3. I was also shocked to find that it was based on the event type “submission” and ‘completion’ rather than direction “IN/OUT.”(and then it writes the defined data tag to the data flag [here](https://github.com/torvalds/linux/blob/master/drivers/usb/mon/mon_bin.c#L607C1-L618C3))

And for the error handling, is https://github.com/torvalds/linux/blob/master/drivers/usb/mon/mon_bin.c#L637C1-L676C2. Usbmon treats error handling as a separate data flag.

In addition, in usbmon, flag_data distinguishes [several payload-related situations](https://github.com/torvalds/linux/blob/master/drivers/usb/mon/mon_bin.c#L408C1-L447C1). The normal case is 0, meaning the payload exists and was successfully copied. Additionally, usbmon defines Z (transfer_buffer == NULL) and D (payload not safely accessible, e.g. DMA related).
These cases are needed because usbmon observes some buffers. Our project does not use such a buffer model, since the payload is already owned and processed directly in userspace. Therefore, the usbmon-specific Z and D cases are not applicable here.

snue: Thanks for the pointers! The way I read the code, the `<`/`>` are just replaced for `IN Submission` and `OUT Completion`, where they also set `length=0`. The actual data of the transfer with data-flag `0` and actual length is in the corresponding IN Completion/OUT Submission likewise. See the `usb_urb_dir_in(...)` usage. And they also track direction for the control endpoint explicitly.

We will see if we can encounter DMA issues and zero buffers. Right now we at least don't handle such situations explicitly, and don't need them in the PCAP files.

ziyifu225: Thanks for the hint — I indeed had not noticed the usb_urb_dir_in() usage before. I took another look at the code, and my current understanding is as follows:
- IN Submission -> captured length = 0, no payload written, flag_data = '<'
- OUT Completion -> captured length = 0, no payload written, flag_data = '>'
- IN Completion -> captured length = actual payload length, flag_data = 0
- OUT Submission -> captured length = actual payload length, flag_data = 0
- Error -> flag_data = 'E'

Is that correct?

snue: Yes, that's how I interpret it currently at least. There is this extra data flag of `=` used in some cases, and figuring out the direction for the control endpoint seemed a bit weird (it also depends on the setup flag). Our "endpoint number" computation may be wrong for the control endpoint, as it does not encode the direction in the even/odd last bit. We may not be able to use the simple `rotate_right` there after all if we have to track direction explicitly.

snue: The `rotate_right` works for the `nusb` code, where we only use it for normal endpoints. The control endpoints are handled differently, so we do not need to special-case that. But in your pcap code, tracking the direction explicitly to also handle the control EP correctly will be required. Sorry, you had that correct before already :wink: (edit: and it is still correct with your changes, you handle control separately)

We should probably add that as comment in the endpoint address calculation.

ziyifu225: I’ve pushed a new commit and updated the logic according to the latest interpretation of the `<`/`>` data flag semantics.

For the `=` data flag, I also did some further research. I could not find a corresponding definition in usbmon itself, so it seems that `=` may no longer be part of the usbmon semantics and is currently mainly used by QEMU.

In QEMU, the meaning of [`<`/`>` is effectively the same as in usbmon](https://github.com/qemu/qemu/blob/master/hw/usb/pcap.c#L175C1-L184C6), except that QEMU uses the setup parameter to distinguish submission vs. completion. Here, setup does not refer to the USB control SETUP stage itself. Instead, QEMU uses _flag_setup_ to indicate whether the packet belongs to the USB control setup stage.

The `=` flag in QEMU is then used for the remaining cases with payload that are neither `<` nor `>`. However, because control packets contain setup information, and that data doesn't blong to URB length, there are situations where the data flag is shown as `=` while the URB length is still 0. This is why, during some control transfers, Wireshark indicates that data exists even though the URB length is zero.

src/device/xhci/nusb.rs
```diff
@@ -82,13 +82,11 @@ impl NusbDeviceWrapper {
...
-
-    match is_out_endpoint {
-        true => endpoint_number,
-        false => 0x80 | endpoint_number,
-    }
+    // `endpoint_id / 2` yields the USB endpoint number, while the low bit of the
+    // xHCI endpoint id distinguishes OUT (even) from IN (odd). Rotating right by
+    // one moves that direction bit into the USB IN position (0x80) and leaves the
+    // endpoint number in the low bits.
+    endpoint_id.rotate_right(1)
```
snue, line 90: I like the comment, but would add the note here as well, so we won't forget 😄 

```suggestion
    // NOTE: This only applies to normal endpoints!
    // `endpoint_id / 2` yields the USB endpoint number, while the low bit of the
    // xHCI endpoint id distinguishes OUT (even) from IN (odd). Rotating right by
    // one moves that direction bit into the USB IN position (0x80) and leaves the
    // endpoint number in the low bits.
    endpoint_id.rotate_right(1)
```

snue (approved):
I think this can be moved out of draft now. Merging this may mean some more work for @Noi0103, but I currently don't see a reason to block on the upcoming PRs.

src/device/pcap/packet.rs
```diff
@@ -0,0 +1,433 @@
...
+impl From<SystemTime> for Timestamp {
+    fn from(value: SystemTime) -> Self {
+        let duration = value
+            .duration_since(SystemTime::UNIX_EPOCH)
+            .unwrap_or_default();
+        Self {
+            seconds: duration.as_secs() as u32,
+            microseconds: duration.subsec_micros(),
+        }
+    }
```
lbeierlieb: What is the point of the `Timestamp` struct? Using `SystemTime` and calling `.as_secs()` instead of `.seconds` and `subsec_micros()` instead of `.microseconds` seems seems equivalent to me, especially because you still include type casts on `seconds` and `microseconds` before passing the values to the record. We could get rid of this struct and the `from` calls around each `SystemTime::now()` call. Or am I missing something?

src/device/xhci/endpoint_launcher.rs
```diff
@@ -116,46 +117,121 @@ impl<CRD: CompleteRealDevice> EndpointLauncher<CRD> {
             debug!("endpoint context specifies endpoint type {endpoint_type:?}");
 
             let endpoint_sender = match device {
-                Some(device) => match endpoint_type {
+                Some(device) => {
+                    let pcap_usb_type = match device.realdevice_ref().speed() {
```
lbeierlieb: why `_type`? I think it should be called something with "bus".

src/device/xhci/endpoint_launcher.rs
```diff
@@ -48,6 +48,14 @@ pub struct LaunchRequester {
     msg_send: mpsc::UnboundedSender<LaunchRequest>,
 }
 
+#[derive(Debug, Clone)]
```
lbeierlieb: do you need the `Clone` for anything? `EndpointContext` is supposed to provide exclusive access to an endpoint context in the guest memory, it should not be cloneable.

src/device/xhci/slot_manager.rs
```diff
@@ -648,7 +648,7 @@ impl Slot {
 /// An endpoint context has a size of 32 bytes, lies in guest memory, and
 /// contains information about an endpoint, most importantly for us the dequeue
 /// pointer and cycle state of the associated transfer ring.
-#[derive(Debug)]
+#[derive(Debug, Clone)]
```
lbeierlieb: See previous comment, `EndpointContext` should not be cloneable.

src/device/pcap/packet.rs
```diff
@@ -61,10 +61,11 @@ pub enum UsbDirection {
 }
 
 impl UsbDirection {
-    const fn endpoint_address(self, endpoint: u8) -> u8 {
+    const fn endpoint_address(self, xhci_endpoint_id: u8) -> u8 {
+        let endpoint_number = xhci_endpoint_id / 2;
         match self {
-            Self::HostToDevice => endpoint & 0x7f,
-            Self::DeviceToHost => endpoint | 0x80,
+            Self::HostToDevice => endpoint_number,
```
lbeierlieb: commit "packet: use USB endpoint addresses in packet headers" should be squashed into the commit that introduced `endpoint_address` function.

src/device/pcap/packet.rs
snue: We have two timestamps in each packet, the `event_timestamp` and the `capture_timestamp`. I don't see how these would differ much in our case. I would be fine with removing the struct and all timestamp mentions and only record the same `now()` for both timestamps in `pcap_record_bytes`.

Alternatively, we record an event timestamp early (before handling/logging the event) and pass it into the various `log_*()` functions, but I don't think that adds meaningful detail currently, and the functions have enough arguments already. I'd prefer simplifying by hiding the timestamp handling further.

lbeierlieb (changes requested):
Some comments--most importantly, `EndpointContext` should not be `Clone`.
There is a bit more opportunity to squash commits that I have not mentioned in comments (where commits change code added by previous commits), but I think we can also leave those commits as is.

src/device/xhci/endpoint_launcher.rs
ziyifu225: Because the data being received here is of the USB type. However, it does occupy the bus number field. How about calling it `pcap_usb_bus_number`? Or do you have a better suggestion?

lbeierlieb: with USB type you mean USB version? As far as I know, we used "version" everywhere to distinguish usb-3 and usb-2 devices. `pcap_usb_bus_number` sounds good.

src/device/pcap/packet.rs
```diff
@@ -152,18 +133,16 @@ pub fn pcap_global_header_bytes() -> [u8; 24] {
...
-    capture_timestamp: Timestamp,
-    event_timestamp: Timestamp,
-    meta: &UsbPacketLinktypeHeader,
-    payload: &[u8],
-) -> Vec<u8> {
-    let link_header = meta.header_bytes(event_timestamp);
+pub fn pcap_record_bytes(meta: &UsbPacketLinktypeHeader, payload: &[u8]) -> Vec<u8> {
+    let now = SystemTime::now()
+        .duration_since(SystemTime::UNIX_EPOCH)
+        .unwrap_or_default();
```
lbeierlieb: Why the conversion to a duration? 🤔 passing the `SystemTime` instance to `header_bytes` transmits the same information.

src/device/xhci/endpoint_launcher.rs
ziyifu225: Yes, I mean "version" here.😂 OK, I'll use `pcap_usb_bus_number` here.

src/device/pcap/packet.rs
ziyifu225: The conversion is needed because the PCAP USB format does not store a timestamp as a `SystemTime` value. The spec defines the timestamp fields as:

> The timestamp fields contain a time stamp for the event, in seconds and microseconds since 1970-01-01 00:00:00 UTC; this MUST be the same as the timestamp for the packet.

So we need to turn the `SystemTime` time point into an elapsed duration since `SystemTime::UNIX_EPOCH`, then split that duration into `as_secs()` and `subsec_micros()` for the PCAP fields.

lbeierlieb: Right, you do the same conversion as before, I missed the detail when it was wrapped in `Timestamp` and the approach makes sense.
Nit: don't use `unwrap_or_default`, which suggests that the conversion could fail. Use `unwrap` with a Safety comment or `expect` and mention that failure is not possible (unless time runs backwards ;D).

lbeierlieb (approved):
Please consider addressing the open nit. Otherwise, all comments addressed and resolved. Thanks for the great work and your endurance on this large PR!

PR #228, opened 2026-04-13
Add inital user documentation

snue:
This is a starting point for our user documentation. Now that there is a `0.1.0` tag, we should not claim "this project is not yet usable" anymore.

* Move goalpost in README.
* Add basic user documentation to obtain and run the software.
* Add security considerations and threat model description.

It is lacking more complete descriptions of service setup and system integration, but these can come later.

### DIFF
--- a/README.md
+++ b/README.md
@@ -9,7 +9,7 @@ virtual machines using the [vfio-user
 protocol](https://www.qemu.org/docs/master/interop/vfio-user.html). Other
 VMMs might also work, but are currently not the main target.
 
-This project is still under active development. We are planning to work on this
+This project is under active development. We are planning to work on this
 project in the following order:
 
 1. **Validating our Assumptions** (Done)
@@ -24,14 +24,18 @@ project in the following order:
      endpoints.
 1. **Hotplug Devices** (Done)
    - We enable attach and detach of exposed devices during runtime.
-1. **Stability & Error Recovery** (🚧 **Ongoing** 🚧)
+1. **Stability & Error Recovery** (Done)
    - We seek out points where we currently panic but should not have to panic.
+1. **Windows Guest Support** (🚧 **Ongoing** 🚧)
+   - Enable all features required for the Windows xHCI driver to interact with
+     devices through the controller.
 1. **Everything Beyond**
    - Many topics remain open. We stay flexible regarding upcoming features.
+   - Missing features include isochronous endpoint support, non-Linux Host
+     support, and support for other VMMs than Cloud Hypervisor.
 
-If you want to use this code, please check back later or [get in
-touch](https://cyberus-technology.de/en/contact), if you need
-professional support.
+If you want to use this code in production and need professional support,
+please [get in touch](https://cyberus-technology.de/en/contact).
 
 ## Documentation
 
--- a/docs/overview.md
+++ b/docs/overview.md
@@ -1,10 +1,16 @@
 # Documentation
 
+This project offers a vfio-user server that allows you to pass USB devices
+(such as storage drives, keyboards, etc.) from a host system to a Cloud
+Hypervisor virtual machine. For a basic overview of the goals, see the
+[Architecture Overview](./developers/architecture.md).
+
 ## For Users
 
-As the project is **not usable yet**, there is no user
-documentation. Please check back later or [get in
-touch](https://cyberus-technology.de/en/contact), if you need
+- [Setup and Invocation](./users/basic.md)
+- [Security Considerations](./users/security.md)
+
+Please [get in touch](https://cyberus-technology.de/en/contact), if you need
 professional support.
 
 ## For Developers
--- a/docs/users/basic.md
+++ b/docs/users/basic.md
@@ -1,0 +1,91 @@
+# Usage
+
+The package currently offers two binaries:
+* `usbvfiod`: the vfio-user server emulating an xHCI controller
+* `remote`: a utility to control USB host device attachment
+
+The server communicates via two Unix sockets:
+* a `vfio-user` socket for communication between `usbvfiod` and the VMM (Cloud Hypervisor)
+* a `hotplug` socket to attach/detach/list devices exposed by `usbvfiod` from the host (optional)
+
+> [!TODO]
+> Currently, the server creates both sockets. The `vfio-user` standard
+> requires that a server can also accept the `vfio-user` socket as a file
+> descriptor, but we have not yet implemented this functionality.
+
+## Obtaining the Package
+
+The software is available through the [GitHub
+website](https://github.com/cyberus-technology/usbvfiod) as a `nix
+flake`, or can be built from source as `cargo` project. We also
+release versions to [crates.io](https://crates.io/crates/usbvfiod)
+and maintain [a `NixOS`
+package](https://search.nixos.org/packages?channel=unstable&query=usbvfiod).
+
+> [!NOTE]
+> There are currently no other officially maintained distribution packages
+> available.
+
+## Invocation
+
+You can run the server through `nix` directly from GitHub, or install the
+binaries in your system through `cargo install usbvfiod`.
+
+```console
+nix run github:cyberus-technology/usbvfiod#usbvfiod -- \
+  --socket-path /path/to/usbvfiod.sock                 \
+  --hotplug-socket-path /path/to/usb-hotplug.sock
+2026-04-10T07:00:16.353894Z  INFO usbvfiod: We're up!
+```
+
+> [!NOTE]
+> Instead of (or additionally to) providing a `hotplug` socket, you can
+> specify devices to expose on the controller directly with
+> `--device /dev/bus/usb/<BUS>/<DEV>`
+
+Connect the virtual xHCI controller to a Cloud Hypervisor instance through the
+`vfio-user` socket, by adding `--user-device socket=/path/to/usbvfiod.sock` to
+the Cloud Hypervisor command line invocation. The socket must exist (`usbvfiod`
+must be running) before Cloud Hypervisor can attach to it.
+
+Use the `remote` binary to list, attach, and detach devices through the
+`hotplug` socket.
+
+```console
+nix run github:cyberus-technology/usbvfiod#remote -- \
+  --socket /path/to/usb-hotplug.sock                 \
+  --list
+No attached devices
+```
+
+Attach a USB device to the controller through the `hotplug` socket. The 
+example uses a keyboard recognized at `/dev/bus/usb/009/003`.
+
+```
+nix run github:cyberus-technology/usbvfiod#remote -- \
+  --socket /tmp/usb-hotplug.sock                     \
+  --attach /dev/bus/usb/009/003
+Requesting attachment of device 009:003
+2026-04-10T07:01:16.353894Z  INFO usbvfiod::device::xhci::port: Attached Full Speed (12 Mbps) device (9, 3) to port 5 (USB2 port)
+    SuccessfulOperation
+```
+
+Detach the device using the recorded device ID also shown on `--list`.
+
+```console
+nix run github:cyberus-technology/usbvfiod#remote -- \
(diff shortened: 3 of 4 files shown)


docs/users/basic.md
```diff
@@ -0,0 +1,89 @@
...
+
+## Obtaining the Package
+
+The software is available through the [GitHub
+website](https://github.com/cyberus-technology/usbvfiod) as a `nix flake`,
+or can be built from source as `cargo` project. We may release versions to
+[crates.io](https://crates.io/crates/usbvfiod) eventually.
+
+> [!NOTE]
+> We also maintain [a `NixOS` package](https://search.nixos.org/packages?channel=unstable&query=usbvfiod).
```
snue: Let's just hope it shows up there soon: https://github.com/NixOS/nixpkgs/pull/509586

README.md
```diff
@@ -24,14 +24,19 @@ project in the following order:
...
    - Many topics remain open. We stay flexible regarding upcoming features.
+   - Missing features include `Isochronous` endpoint support, non-Linux Host
+     support, and support for other VMMs than Cloud-Hypervisor.
 
-If you want to use this code, please check back later or [get in
-touch](https://cyberus-technology.de/en/contact), if you need
-professional support.
+If you want to use this code in production, please [get in
+touch](https://cyberus-technology.de/en/contact), if you need professional
+support.
```
lbeierlieb: The changes are fine, but I think the sentence is a bit clunky in general. Can we remove the first or the second if clause, or combine them ("If you want to use this code in production and need professional support")?

README.md
```diff
@@ -24,14 +24,19 @@ project in the following order:
...
    - We enable attach and detach of exposed devices during runtime.
-1. **Stability & Error Recovery** (🚧 **Ongoing** 🚧)
+1. **Stability & Error Recovery** (Done)
    - We seek out points where we currently panic but should not have to panic.
+1. **Windows Guest Support** (🚧 **Ongoing** 🚧)
+   - Enable all features required for the Windows xHCI driver to interact with
+     devices through the controller.
 1. **Everything Beyond**
    - Many topics remain open. We stay flexible regarding upcoming features.
+   - Missing features include `Isochronous` endpoint support, non-Linux Host
```
lbeierlieb: nit: I find the highlighting (and capitalization) of "isochronous" unnecessary.

README.md
```diff
@@ -24,14 +24,19 @@ project in the following order:
...
+     support, and support for other VMMs than Cloud-Hypervisor.
```
lbeierlieb: "cloud-hypervisor" (name of the binary, name of the repository) and "Cloud Hypervisor" (how the readme of the cloud-hypervisor repository refers to the project) are fine for me, but "Cloud-Hypervisor" is a mixture of the two that I have not seen before anywhere and I suggest we do not use this spelling variant.

docs/users/basic.md
```diff
@@ -0,0 +1,89 @@
+# Usage
+
+The package currently offers two binaries:
+* `usbvfiod` - the vfio-user server emulating an xHCI controller
+* `remote` - a utility to control USB host device attachment
```
lbeierlieb: nit: I would prefer colons (or alternatively en dashes) instead of the hyphens.

docs/users/basic.md
```diff
@@ -0,0 +1,89 @@
...
+
+Two Unix sockets are provided for operation:
```
lbeierlieb: "are provided" is confusing here. Who provides the sockets? The documentation just above talked about the package. Maybe "The usbvfiod binary provides two Unix sockets." or probably more what we want to express here: "The usbvfiod binary communicates via two Unix sockets."

docs/users/basic.md
```diff
@@ -0,0 +1,89 @@
...
+* a `vfio-user` socket for communication between `usbvfiod` and the VMM (Cloud Hypervisor)
+* a `hotplug` socket to attach/detach/list devices exposed by `usbvfiod` from the host
```
lbeierlieb: nit: mention this socket is optional?

docs/users/basic.md
```diff
@@ -0,0 +1,89 @@
...
+
+> [!TODO]
+> Both sockets are currently created by the server. It must also be possible
+> to provide the `vfio-user` socket as file descriptor, but this is not yet
+> implemented.
```
lbeierlieb: Suggestion (because the review tool doesn't let me create one and I don't know why): "Currently, the server creates both sockets. The vfio-user standard requires that a server can also accept the vfio-user socket as a file descriptor, but we have not yet implemented this functionality."

docs/users/basic.md
lbeierlieb: Why is this NixOS package mention a separate note and not just part of the previous paragraph? I don't see the distinction.

docs/users/basic.md
```diff
@@ -0,0 +1,89 @@
...
+
+```console
+nix run github:cyberus-technology/usbvfiod#remote -- \
+  --socket /path/to/usb-hotplug.sock                 \
+  --list
+No attached devices
+```
+
+Attach a USB device to the controller through the `hotplug` socket. The 
+example uses a Keyboard recognized at `/dev/bus/usb/009/003`.
```
lbeierlieb: I see no reason to capitalize keyboard.

docs/users/basic.md
```diff
@@ -0,0 +1,89 @@
...
+```
+
+## Device Access Restrictions
+
+The `usbvfiod` server and `remote` binary do not require elevated privileges.
+Only access to the USB character device nodes is required for the component
+opening the device. The permissions for specific devices can be managed through
+`udev` rules (`TAG+="uaccess"`), or the `remote` binary can be invoked with
+elevated privileges (e.g. `sudo`). The `hotplug` socket must be accessible to
+the `remote` binary.
```
lbeierlieb, line 91: nit: use "you can manage"/"you can invoke" instead of passive voice.

docs/users/security.md
```diff
@@ -0,0 +1,100 @@
...
+those guarantees.
+
+### Host and Guest Memory Access
+
+The server has full access to the Guest VM's physical backing memory.
+Guest-controlled DMA (Direct Memory Access) operations of the virtual
+controller are only supported from/to this memory. The Guest VM is generally
+allowed arbitrary modifications to its memory, or to "shoot itself in the
+foot" through the virtual controller. Access to Guest memory regions outside
+the backing memory (i.e. "P2P DMA") is deliberately not supported through the
```
lbeierlieb: needs to be "i.e.,"

docs/users/security.md
```diff
@@ -0,0 +1,100 @@
...
+`vfio-user` server.
+
+All Host-DMA is gated through the Kernel user-space USB API. No direct control
+over the host USB controller is granted to the server.
+
+### Privilege Separation
+
+The server accepts handles to open devices through the `hotplug` socket. This
```
lbeierlieb: "handles to open devices" could be "handles to devices that are open" or "handles in order to open devices". Did you mean the former? And why "handle" instead of "file descriptor"?

docs/users/security.md
```diff
@@ -0,0 +1,100 @@
...
+ensures the server itself never requires the privileges to open device nodes,
```
lbeierlieb: "never" might be confusing. The server does need privileges if it wants to open devices supplied with `--device`. You mentioned the option above, so the "never" statement seems conflicting.

docs/users/security.md
```diff
@@ -0,0 +1,100 @@
...
+## Hardening Options
+
+The design allows for various orthogonal hardening options.
+
+### Service Sandboxing
+
+The vfio-user server can be run with strict sandboxing to further limit its capabilities:
+* **Process isolation**: Ensure the server runs as a non-root user with minimal group memberships.
+* **User namespace isolation**: Run the server in a dedicated user namespace where it has minimal privileges.
+* **System call filtering**: The `vfio-user` server allowed syscalls can be further restricted, e.g. through [systemd](https://www.freedesktop.org/software/systemd/man/latest/systemd.exec.html#System%20Call%20Filtering).
```
lbeierlieb: e.g.,

lbeierlieb (commented):
Thanks for writing up the documentation! ❤️  I did not make it all the way through the changes, but I accumulated a number of comments and submit these comments now, because I'm not sure when I come around to continue the review.

docs/users/basic.md
snue: I took your suggestion.

docs/users/basic.md
snue: I marked it "(optional)" in this listing.

docs/users/basic.md
snue: Reworded with your suggestion.

docs/users/basic.md
snue: Colons it is.

README.md
snue: Yes. I fixed the spelling for consistency.

README.md
snue: Indeed. I took your suggestion.

docs/users/basic.md
snue: Removed the passive voice.

docs/users/security.md
snue: Yes, the former: "already opened" ... I thought "file descriptor" is a bit too much of an implementation detail, but it's not going to be anything else ever. So I now went with that.

docs/users/security.md
snue: Ok, I removed the "never". The idea was that it is never mandatory, given the hotplug socket. But I see how this can be confusing.

docs/users/basic.md
snue: The note only mentioned there are no official distribution packages initially. Then you created the nixpkgs PR and I fixed the note. I restructured and now just list the package as one of the official options above.

docs/users/basic.md
```diff
@@ -0,0 +1,90 @@
+# Usage
+
+The package currently offers two binaries:
+* `usbvfiod`: the vfio-user server emulating an xHCI controller
+* `remote`: a utility to control USB host device attachment
```
blitz, line 5: Maye `usbvfiodctrl` would be a nice name.

snue: Or `usbvfioctl` (to roughly align with `systemd`/`systemctl` naming) or `cyberusb` :wink:  I'd leave the rename for a different PR than the documentation, though.

docs/overview.md
```diff
@@ -1,15 +1,21 @@
 # Documentation
 
+This project offers a vfio-user server that allows you to pass USB devices
+(like storage drives, keyboards, etc.) from a host system to a Cloud
```
lbeierlieb: "like" and "etc" both indicate that there can be more. Storage drivers and keyboards are part of the group of "USB devices", so you should use "such as" instead of "like".


```suggestion
(such as storage drives, keyboards) from a host system to a Cloud
```

lbeierlieb (approved):

docs/overview.md
snue: Nice suggestion to use `such as` here, but I would prefer to keep the `, etc.` to indicate there can be more. Storage and keyboard are tested in CI, but I would not prematurely put the restriction in the documentation.

snue:
CI [hit something in the attach/detach test](https://hercules-ci.com/accounts/github/cyberus-technology/derivations/%2Fnix%2Fstore%2Fcsxv39pdi9dhlx9776yvdxzyigmsj0nk-vm-test-run-attach-detach-usb-1.1.drv/log?via-job=e4028321-a6a3-4faa-811b-a93099eb5e0d) we have not seen so far:

```
[   80.756652] usbvfiod[666]: thread 'tokio-rt-worker' (706) panicked at src/device/xhci/nusb.rs:336:92:
[   80.757188] usbvfiod[666]: Failed to open endpoint on nusb device. We could handle this error and always return transaction errors. Panic is fine for now.: endpoint already in use
```

Obviously not related to documentation. Still something we don't need to panic on necessarily.

PR #236, opened 2026-05-05
Add partial portpmsc implementation

Noi0103:
Windows will attempt to access this register so we should enable some basic functionality for reading and writing.

### DIFF
--- a/src/device/pci/xhci.rs
+++ b/src/device/pci/xhci.rs
@@ -16,7 +16,7 @@ use crate::device::{
     },
     xhci::{
         endpoint_launcher::EndpointLauncher,
-        port::{get_portli_index, get_portsc_index, HotplugControl, PortArray},
+        port::{get_portli_index, get_portpmsc_index, get_portsc_index, HotplugControl, PortArray},
         real_device::CompleteRealDevice,
         registers::{UsbcmdRegister, UsbstsRegister},
         slot_manager::SlotManager,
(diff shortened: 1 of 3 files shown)


src/device/xhci/port.rs
```diff
@@ -314,6 +330,10 @@ pub const fn get_portsc_index(addr: u64) -> Option<usize> {
     get_port_index_from_addr(addr, offset::PORTSC, MAX_PORTS, 0)
 }
 
+pub const fn get_portpmsc_index(addr: u64) -> Option<usize> {
+    get_port_index_from_addr(addr, offset::PORTSC, MAX_PORTS, 0x4)
+}
```
snue, line 335: These helpers are only used from the MMIO access side, and there they want the port ID instead. We should make the `get_port_index_from_addr` return the `get_port_id_from_addr` (i.e. `+1`) instead, and get rid of all the extra `port_index + 1` in the MMIO read/write for portsc/portpmsc. And maybe rename all functions to `_id` instead of `_index`, but that's not important as the `id` will be the correct index for the `OneIndexed` array.

lbeierlieb (approved):

snue (commented):
Nothing wrong with that. I just spot an opportunity to get rid of some redundant offset handling. You wouldn't strictly need `OneIndexed` for the current use of `portpmsc` (only through the MMIO path). But we can keep it consistent and improve for `portsc` as well.

src/device/xhci/port.rs
lbeierlieb: The suggestion is great, I suggest to merge as is and replace the index-based handling with id. In a follow up, @Noi0103 (who currently can use the time better with further Windows testing) or @ziyifu225 (perhaps if this task is still open when PCAP is done; she also fittingly wrote the original portsc-indexing code) can rework these functions for portsc, portpmsc, portli.

src/device/pci/xhci.rs
```diff
@@ -163,6 +163,16 @@ impl<CRD: CompleteRealDevice> PciDevice for XhciController<CRD> {
                     .write_portsc(port_id, value)
                     .expect("event_sender should be alive");
             }
+
+            addr if get_portpmsc_index(addr).is_some() => {
+                // SAFETY: unwrap() is safe because we already checked is_some() in the match guard above
+                let port_index = get_portpmsc_index(addr).unwrap();
```
snue, line 169: This will be so much nicer with `if let` support in match arms. But it's only a stable feature with rust 1.95.

PR #237, opened 2026-05-05
Upgrade USBCMD register from bool to u32

Noi0103:
We previously never had to remember any values of USBCMD except the bit 0 with Run/Stop state. With windows as guest after an error occurs we did not properly provide the USBCMD state and the driver did not clean up as expected (instead assumed a stopped xHC and acted according to that).
This is, as of right now, likely only relevant for debugging, when something else already did not work and system software attempts some recovery.

### DIFF
--- a/src/device/pci/xhci.rs
+++ b/src/device/pci/xhci.rs
@@ -43,7 +43,7 @@ pub struct XhciController<CRD: CompleteRealDevice> {
 impl<CRD: CompleteRealDevice> XhciController<CRD> {
     pub fn new(dma_bus: BusDeviceRef, async_runtime: runtime::Handle) -> Self {
         let usbcmd = UsbcmdRegister::new();
-        let usbsts = UsbstsRegister::new(usbcmd.running_bit());
+        let usbsts = UsbstsRegister::new(usbcmd.value_reference());
         let interrupter = Interrupter::new(dma_bus.clone(), &async_runtime);
         let port_array = PortArray::new(interrupter.create_event_sender(), async_runtime.clone());
         let ep_launch_requester = EndpointLauncher::start(
@@ -58,7 +58,7 @@ impl<CRD: CompleteRealDevice> XhciController<CRD> {
             &async_runtime,
             interrupter.create_event_sender(),
             slot_manager.create_slot_worker_handle(),
-            usbcmd.running_bit(),
+            usbcmd.value_reference(),
         );
 
         Self {
@@ -202,7 +202,7 @@ impl<CRD: CompleteRealDevice> PciDevice for XhciController<CRD> {
             offset::SUPPORTED_PROTOCOLS_USB2_CONFIG_RESERVED => 0,
 
             // xHC Operational Registers
-            offset::USBCMD => 0,
+            offset::USBCMD => self.usbcmd.read(),
             offset::USBSTS => self.usbsts.read(),
             offset::DNCTL => 2,
             offset::CRCR => self.command_ring.status(),
--- a/src/device/xhci/command_ring.rs
+++ b/src/device/xhci/command_ring.rs
@@ -1,7 +1,7 @@
 //! Implements a XHCI command ring and a worker task that services the ring.
 
 use std::sync::{
-    atomic::{AtomicBool, Ordering},
+    atomic::{AtomicBool, AtomicU32, Ordering},
     Arc,
 };
 
@@ -14,7 +14,7 @@ use tracing::{debug, info, trace, warn};
 
 use crate::device::{
     bus::BusDeviceRef,
-    pci::constants::xhci::operational::crcr,
+    pci::constants::xhci::operational::{crcr, usbcmd},
     xhci::{
         interrupter::EventSender,
         linked_ring::LinkedRing,
@@ -34,7 +34,7 @@ struct CommandWorker {
     state: WorkerState,
     receiver: mpsc::UnboundedReceiver<WorkerMessage>,
     commandring_running: Arc<AtomicBool>,
-    controller_running: Arc<AtomicBool>,
+    usbcmd: Arc<AtomicU32>,
     event_sender: EventSender,
     ring: LinkedRing,
     slot_handle: SlotWorkerHandle,
@@ -71,7 +71,7 @@ impl CommandRing {
         async_runtime: &runtime::Handle,
         event_sender: EventSender,
         slot_handle: SlotWorkerHandle,
-        controller_running: Arc<AtomicBool>,
+        usbcmd: Arc<AtomicU32>,
     ) -> Self {
         let (sender_to_worker, receiver) = mpsc::unbounded_channel();
         let running = Arc::new(AtomicBool::new(false));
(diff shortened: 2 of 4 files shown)


src/device/xhci/registers.rs
```diff
@@ -162,41 +162,53 @@ impl DcbaapRegister {
...
-        Self { running }
+    pub fn read(&self) -> u32 {
+        self.value.load(std::sync::atomic::Ordering::Relaxed)
     }
 
-    pub fn write(&self, value: u64) {
-        self.running.store(value & 0x1 != 0, Ordering::Relaxed);
+    pub fn write(&self, value: u32) {
+        // this mask is incomplete due to:
+        // - bit 7 being defined as RW or RO depending on the HCCPARAMS1 LHRC bit
```
snue: But we know the contents of `HCCPARAMS1`? It's hardcoded and only contains the capability pointer.

src/device/xhci/registers.rs
```diff
@@ -162,41 +162,53 @@ impl DcbaapRegister {
...
+        // - bit 14
+        // - bit 15
+        const BITMASK_RW: u32 = 0b100101111100001111;
+
+        // an actual controller reset is impossible to do
+        const BITMASK_CANT_IMPLEMENT: u32 = 0b10;
```
snue: We can emulate a controller reset quite well. Just all devices will disappear. And we probably never need to signal a reset "in progress" as long as the driver ensures the controller halted (as "recommended" in the spec). We currently don't do anything on reset, though.

We can just exclude the bit from the "writable" (storable) bits right away.

src/device/xhci/registers.rs
```diff
@@ -162,41 +162,53 @@ impl DcbaapRegister {
...
     }
 
-    pub fn write(&self, value: u64) {
-        self.running.store(value & 0x1 != 0, Ordering::Relaxed);
+    pub fn write(&self, value: u32) {
+        // this mask is incomplete due to:
+        // - bit 7 being defined as RW or RO depending on the HCCPARAMS1 LHRC bit
+        // - bit 14
+        // - bit 15
+        const BITMASK_RW: u32 = 0b100101111100001111;
```
snue: Can we instead just mention what we support here (in the comment)? The mask doesn't look spec compliant. At least bits 8 and 9 should always read 0.

Noi0103: I wasn't sure what exactly to add in the mask, so I took all writable bits that are non-optional according to the spec and did not think about usbvfiod specific things like the HCCPARAMS1 LHRC being known.

We could remove 8 and 9 and maybe at some point in the future catch them in `write()` when we need to.

src/device/pci/xhci.rs
```diff
@@ -114,7 +114,8 @@ impl<CRD: CompleteRealDevice> PciDevice for XhciController<CRD> {
 
         match req.addr {
             // xHC Operational Registers
-            offset::USBCMD => self.usbcmd.write(value),
+            // SAFETY: The USBCMD is defined as 32 bit
+            offset::USBCMD => self.usbcmd.write(value as u32),
```
snue: Checking the supplied value can be left to the called function if necessary. Just make the signature take u64 again and remove the mental load from the call site.

src/device/xhci/command_ring.rs
```diff
@@ -206,7 +207,8 @@ impl CommandWorker {
                             msg => warn!("Unexpected message: msg={msg:?}, state={:?}", self.state),
                         }
                     }
-                    if !self.controller_running.load(Ordering::Relaxed) {
+                    let controller_running = (self.usbcmd.load(Ordering::Relaxed) & 0x1) == 1;
```
snue: You defined the RUN/STOP bit in the constants now. You can use it here.

```suggestion
                    let controller_running = (self.usbcmd.load(Ordering::Relaxed) & usbcmd::RS) == usbcmd::RS;
```

It's a bit unfortunate we have to be so explicit here and don't have a nice convenience wrapper to check if a bit is set.

You can also call it `controller_stopped`, check for `== 0`, and remove the inversion on the `if` below.

src/device/xhci/registers.rs
```diff
@@ -181,41 +181,54 @@ impl DcbaapRegister {
...
+        Self { usbcmd }
     }
 
     pub fn read(&self) -> u64 {
-        let hch = if self.running.load(Ordering::Relaxed) {
-            0
-        } else {
-            usbsts::HCH
-        };
+        let is_running = (self.usbcmd.load(Ordering::Relaxed) & 0x1) == 1;
```
snue: Same here, we can use the declared bit constants.

src/device/xhci/registers.rs
```diff
@@ -181,41 +181,54 @@ impl DcbaapRegister {
...
+    pub fn read(&self) -> u64 {
+        self.value.load(std::sync::atomic::Ordering::Relaxed).into()
     }
 
-    pub fn write(&self, value: u64) {
-        self.running.store(value & 0x1 != 0, Ordering::Relaxed);
+    /// Simple write with a very limited list of allowed bits (RsvdP is also not written).
+    pub fn write(&self, value: u32) {
+        // Currently writable bits ...
+        const BITMASK_RW: u32 = usbcmd::RS as u32 | usbcmd::INTE as u32;
```
snue: nit: We can call them `BITMASK_PRESERVED`. That's what we do with those bits.

src/device/xhci/registers.rs
```diff
@@ -181,41 +181,54 @@ impl DcbaapRegister {
...
-    pub fn write(&self, value: u64) {
-        self.running.store(value & 0x1 != 0, Ordering::Relaxed);
+    /// A simple write with a very limited list of allowed bits (RsvdP is also not written).
+    pub fn write(&self, value: u32) {
+        // Currently writable bits ...
+        const BITMASK_RW: u32 = usbcmd::RS as u32 | usbcmd::INTE as u32;
+        // ... ignoring any other bits and printing a warning when attempted.
+        if value & !BITMASK_RW > 0 {
+            warn!(
+                "received at least one bit that is ignored for USBCMB: {}",
```
snue: typo:
```suggestion
                "received at least one bit that is ignored for USBCMD: {}",
```

src/device/pci/xhci.rs
```diff
@@ -115,7 +115,7 @@ impl<CRD: CompleteRealDevice> PciDevice for XhciController<CRD> {
         match req.addr {
             // xHC Operational Registers
             // SAFETY: The USBCMD is defined as 32 bit
```
snue: Now you should remove the "SAFETY" statement as well 😉

src/device/xhci/registers.rs
```diff
@@ -196,18 +197,21 @@ impl UsbcmdRegister {
...
             warn!(
-                "received at least one bit that is ignored for USBCMB: {}",
-                value & !BITMASK_RW
+                "received at least one bit that is ignored for USBCMD: {}",
+                value & !BITMASK_PRESERVED
             );
         }
 
-        let value = value & BITMASK_RW;
+        let value = value & BITMASK_PRESERVED;
```
snue: If you do the conversion before writing the value, you don't need conversions on the bit masks. It's also safe because we control the bits that can be stored, and they are all within the lower 32 bits.

snue (approved):

PR #238, opened 2026-05-06
Implement event data trb handling

Noi0103:
Windows is never (as far as I saw in logs at least) utilizing the interrupt on completion bit. Windows will instead use event data trb.

The event data appeared on the Control Endpoint and the handler is initially written for the Control Endpoint.

When NormalIn and NormalOut Endpoints also needed event data handling I vendored the handle functions into the respective NormalEndpoint implementations because it seemed easiest. Combining this so that any endpoint can call the same handle_event_data seems better/less error prone.
Should that be added to this PR?

### DIFF
--- a/src/device/pci/constants.rs
+++ b/src/device/pci/constants.rs
@@ -291,12 +291,13 @@ pub mod xhci {
 
     /// Constants for the capability register.
     pub mod capability {
+
         /// We only emulate version 1.0.0 of the XHCI spec for simplicity.
         pub const HCIVERSION: u64 = 0x100;
         pub const HCSPARAMS1: u64 =
             (super::MAX_PORTS << 24) | (super::MAX_INTRS << 8) | super::MAX_SLOTS;
         pub const HCSPARAMS2: u64 = super::MAX_ERST_SIZE_EXP << 4;
-        pub const HCCPARAMS1: u64 = super::offset::SUPPORTED_PROTOCOLS << 14;
+        pub const HCCPARAMS1: u64 = hccparams1::PAE | super::offset::SUPPORTED_PROTOCOLS << 14;
 
         pub const USB_STRING: u64 = 0x20425355;
 
@@ -319,6 +320,12 @@ pub mod xhci {
             pub const CAP_INFO: u64 = ID | (MAJOR << 24) | (MINOR << 16) | (NEXT << 8);
             pub const CONFIG: u64 =
                 (super::super::NUM_USB3_PORTS + 1) | (super::super::NUM_USB2_PORTS << 8);
+        }
+
+        // See xhci specification chapter 5.3.6
+        pub mod hccparams1 {
+            // Parse all Event Data TRB while advancing to the next TD after a Short Packet.
+            pub const PAE: u64 = 0x100;
         }
     }
 
--- a/src/device/xhci/endpoint_handle.rs
+++ b/src/device/xhci/endpoint_handle.rs
@@ -1,14 +1,14 @@
 use std::{
+    cmp::min,
     fmt::Debug,
     future::Future,
     mem::{self},
-    ops::ControlFlow,
     pin::Pin,
 };
 
 use anyhow::anyhow;
 use replace_with::replace_with_or_abort;
-use tracing::{debug, warn};
+use tracing::{debug, info, trace, warn};
 
 use crate::device::{
     bus::BusDeviceRef,
@@ -22,8 +22,8 @@ use crate::device::{
             RealOutEndpointHandle,
         },
         trb::{
-            CompletionCode, EventDataTrb, EventTrb, NormalTrb, RawTrb, TransferTrb,
-            TransferTrbVariant,
+            CompletionCode, DataStageTrb, EventDataTrb, EventTrb, NormalTrb, RawTrb, SetupStageTrb,
+            StatusStageTrb, TransferTrb, TransferTrbVariant, TrbDmaInfo,
         },
         usbrequest::UsbRequest,
     },
@@ -79,16 +79,77 @@ impl BaseEndpointHandle for DummyEndpointHandle {
     }
 }
 
+#[derive(Debug, PartialEq, Eq)]
+pub struct ControlTransferState {
+    /// upcoming or current stage/TD of a control transfer to be handled
+    pub state: ControlTransferStage,
+    /// holding the UsbRequest and all associated data
+    pub data: ControlTransferData,
+}
+impl ControlTransferState {
+    const fn new(data: ControlTransferData) -> Self {
+        Self {
+            state: ControlTransferStage::ExpectSetupStageTrb,
+            data,
+        }
+    }
+}
+
+fn interrupt_on_completion(
+    address: u64,
+    completion_code: CompletionCode,
+    event_data: bool,
+    endpoint_id: u8,
+    slot_id: u8,
+    event_sender: &EventSender,
+) -> anyhow::Result<()> {
+    trace!("interrupt_on_completion triggered for address {}", address);
+    let event = EventTrb::new_transfer_event_trb(
+        address,
+        0,
+        completion_code,
+        event_data,
+        endpoint_id,
+        slot_id,
+    );
+
+    event_sender.send(event)?;
+    Ok(())
+}
+
+// Track how far we are with parsing the Control Transfer (chain of TRB).
+#[derive(Debug, PartialEq, Eq)]
+pub enum ControlTransferStage {
+    /// Nothing happened yet. Awaiting a Setup Stage Trb and dropping any other
+    /// Trb (they will not reach the hardware device).
+    ExpectSetupStageTrb,
+    /// Either collect data if a Data Stage Trb is received or skip the Data
+    /// Stage TD altogether if a Status Stage Trb is received.
+    MaybeDataStageTrb,
+    MoreData,
+    /// Finished processing the Data Stage if there was one.
+    ExpectStatusStageTrb,
+}
+/// The state machine provides the information partially as ControlSubmissionState::AwaitingControlIn(TransferTrb).
+/// Track state between us and the guest for building the current control request.
+#[derive(Debug, PartialEq, Eq)]
+pub enum ControlTransferData {
+    In(UsbRequest),
+    Out(UsbRequest),
+}
+
 #[derive(Debug)]
 pub struct ControlEndpointHandle<RCEH: RealControlEndpointHandle> {
     slot_id: u8,
     endpoint_id: u8,
     pcap_meta: EndpointPcapMeta,
     real_ep: RCEH,
-    trb_parser: ControlRequestParser,
     dma_bus: BusDeviceRef,
     event_sender: EventSender,
+    /// referring to usbvfiod to hardware communication
     submission_state: ControlSubmissionState,
+    /// referring to usbvfiod to guest communication
+    transfer_state: ControlTransferState,
 }
 
 impl<RCEH: RealControlEndpointHandle> ControlEndpointHandle<RCEH> {
@@ -105,24 +166,303 @@ impl<RCEH: RealControlEndpointHandle> ControlEndpointHandle<RCEH> {
             endpoint_id,
             pcap_meta,
             real_ep,
-            trb_parser: ControlRequestParser::new(dma_bus.clone()),
             dma_bus,
             event_sender,
             submission_state: ControlSubmissionState::NoTrbSubmitted,
+            transfer_state: ControlTransferState::new(ControlTransferData::In(
+                UsbRequest::default(),
+            )),
+        }
+    }
+
+    fn handle_setup_stage_trb(&mut self, address: u64, trb: SetupStageTrb) -> anyhow::Result<()> {
+        let usb_request = UsbRequest {
+            address,
+            request_type: trb.request_type,
+            request: trb.request,
+            value: trb.value,
+            index: trb.index,
+            length: trb.length,
+            data_pointer: None,
+            data: vec![],
+        };
+
+        if trb.request_type & 0x80 != 0 {
+            trace!("SetupStage TRB with ControlIn");
+
+            self.transfer_state =
+                ControlTransferState::new(ControlTransferData::In(usb_request.clone()));
+
+            self.real_ep.submit_control_request(usb_request.clone())?;
+            pcap::control_submission(self.pcap_meta, &usb_request);
+
+            self.submission_state = ControlSubmissionState::AwaitingControlIn(
+                address,
+                TransferTrbVariant::SetupStage(trb),
+            );
+        } else {
+            trace!("SetupStage TRB with ControlOut");
+
+            self.transfer_state = ControlTransferState::new(ControlTransferData::Out(usb_request));
+
+            // actual hardware request happens in status stage after consuming the data stage td
+
+            if trb.interrupt_on_completion {
+                interrupt_on_completion(
+                    address,
+                    CompletionCode::Success,
+                    false,
+                    self.endpoint_id,
+                    self.slot_id,
+                    &self.event_sender,
+                )?;
+            }
+
+            self.transfer_state.state = ControlTransferStage::MaybeDataStageTrb;
+            self.submission_state = ControlSubmissionState::ParserConsumedTrb(
+                address,
+                TransferTrbVariant::SetupStage(trb),
+            );
+        }
+
+        Ok(())
+    }
+
+    fn handle_setup_stage_hardware_response(
+        &mut self,
+        address: u64,
+        trb: SetupStageTrb,
+        hardware_data: &mut Vec<u8>,
+    ) -> anyhow::Result<()> {
+        match &mut self.transfer_state.data {
+            // collect hardware data
+            ControlTransferData::In(request) => {
+                trace!("control in data {:?}", hardware_data);
+
+                pcap::control_completion_in(self.pcap_meta, request.address, hardware_data);
+
+                request.data.append(hardware_data);
+
+                request.data.resize(trb.length as usize, 0);
+
+                if trb.interrupt_on_completion {
+                    interrupt_on_completion(
+                        address,
+                        CompletionCode::Success,
+                        false,
+                        self.endpoint_id,
+                        self.slot_id,
+                        &self.event_sender,
+                    )?;
+                }
+
+                self.transfer_state.state = ControlTransferStage::MaybeDataStageTrb;
+            }
+            ControlTransferData::Out(_) => {
+                unreachable!("internal error: ControlOut SetupTrb have insufficient information to do the Hardware request; a submission state to arrive here should never be used");
+            }
+        }
+        Ok(())
+    }
+
+    fn transfer_data_slices<T: TrbDmaInfo>(&mut self, address: u64, trb: &T) {
+        let data_pointer = trb.data_pointer();
+        let immediate_data = trb.has_immediate_data();
+
+        let transfer_length = if immediate_data {
+            min(8, trb.transfer_length())
+        } else {
+            trb.transfer_length()
+        };
+
+        match &mut self.transfer_state.data {
+            ControlTransferData::In(usb_request) => {
+                trace!("DMA for ControlIn");
+
+                // From xhci specification chapter 4.11.2.2:
+                //
+                // System software is responsible for ensuring that the total data length defined by a
+                // Data Stage TD (i.e. the sum of the Length fields of the Data Stage TRB and all Normal
+                // TRBs) is equal to wLength. Note that communicating with some non-compliant
+                // devices may require violating this rule.
+                if usb_request.data.len() < transfer_length as usize {
+                    self.transfer_state.state = ControlTransferStage::ExpectStatusStageTrb;
+                    self.submission_state = ControlSubmissionState::ParserError(address);
+                    return;
+                }
+
+                let byte_slice: Vec<u8> = usb_request
+                    .data
+                    .drain(0..transfer_length as usize)
+                    .collect();
+
+                trace!(
+                    "DataStage TRB len: {} slice: {:?}",
+                    byte_slice.len(),
+                    byte_slice
+                );
+                self.dma_bus.write_bulk(data_pointer, &byte_slice);
+            }
+            ControlTransferData::Out(usb_request) => {
+                trace!("DMA for ControlOut");
+
+                if immediate_data {
+                    // Only event data should follow when immediate data is used here
+                    // but we do not check for that and allow multiple immediate data
+                    // TRB in the data stage TD.
+
+                    usb_request.data.append(
+                        &mut data_pointer.to_le_bytes()[..transfer_length as usize].to_vec(),
+                    );
+                } else {
+                    let mut tmp = vec![0u8; transfer_length as usize];
+                    self.dma_bus.read_bulk(data_pointer, &mut tmp);
+
+                    usb_request.data.append(&mut tmp);
+                }
+            }
+        }
+    }
+
+    fn handle_data_stage_trb(&mut self, address: u64, trb: DataStageTrb) -> anyhow::Result<()> {
+        trace!("DataStage TRB");
+
+        self.transfer_data_slices(address, &trb);
+
+        if trb.interrupt_on_completion {
+            interrupt_on_completion(
+                address,
+                CompletionCode::Success,
+                false,
+                self.endpoint_id,
+                self.slot_id,
+                &self.event_sender,
+            )?;
+        }
+
+        if trb.chain {
+            self.transfer_state.state = ControlTransferStage::MoreData;
+        } else {
+            self.transfer_state.state = ControlTransferStage::ExpectStatusStageTrb;
+        }
+
+        self.submission_state =
+            ControlSubmissionState::ParserConsumedTrb(address, TransferTrbVariant::DataStage(trb));
+        Ok(())
+    }
+
+    fn handle_normal_trb(&mut self, address: u64, trb: NormalTrb) -> anyhow::Result<()> {
+        trace!("Normal TRB");
+
+        self.transfer_data_slices(address, &trb);
+
+        if trb.interrupt_on_completion {
+            interrupt_on_completion(
+                address,
+                CompletionCode::Success,
+                false,
+                self.endpoint_id,
+                self.slot_id,
+                &self.event_sender,
+            )?;
+        }
+
+        if !trb.chain {
+            self.transfer_state.state = ControlTransferStage::ExpectStatusStageTrb;
+        }
+
+        self.submission_state =
+            ControlSubmissionState::ParserConsumedTrb(address, TransferTrbVariant::Normal(trb));
+        Ok(())
+    }
+
+    fn handle_status_stage_trb(&mut self, address: u64, trb: StatusStageTrb) -> anyhow::Result<()> {
+        match &mut self.transfer_state.data {
+            ControlTransferData::In(_) => {
+                trace!("StatusStage TRB with ControlIn");
+
+                if trb.interrupt_on_completion {
+                    interrupt_on_completion(
+                        address,
+                        CompletionCode::Success,
+                        false,
+                        self.endpoint_id,
+                        self.slot_id,
+                        &self.event_sender,
+                    )?;
+                }
+
+                if !trb.chain {
+                    self.transfer_state.state = ControlTransferStage::ExpectSetupStageTrb;
+                }
+                self.submission_state = ControlSubmissionState::ParserConsumedTrb(
+                    address,
+                    TransferTrbVariant::StatusStage(trb),
+                );
+            }
+            ControlTransferData::Out(usb_request_out) => {
+                trace!("StatusStage TRB with ControlOut");
+
+                self.real_ep
+                    .submit_control_request(usb_request_out.clone())?;
+                pcap::control_submission(self.pcap_meta, usb_request_out);
+
+                self.submission_state = ControlSubmissionState::AwaitingControlOut(
+                    address,
+                    TransferTrbVariant::StatusStage(trb),
+                );
+            }
+        }
+        Ok(())
+    }
+
+    fn handle_status_stage_hardware_response(
+        &mut self,
+        address: u64,
+        trb: StatusStageTrb,
+    ) -> anyhow::Result<()> {
+        match &mut self.transfer_state.data {
+            ControlTransferData::In(_) => {
+                unreachable!("internal error: ControlIn requests do the Hardware request in the SetupStage; a submission state to arrive here should never be used");
+            }
+            ControlTransferData::Out(usb_request) => {
+                trace!("StatusStage TRB with ControlOut");
+
+                pcap::control_completion_out(
+                    self.pcap_meta,
+                    usb_request.address,
+                    u32::from(usb_request.length),
+                );
+
+                if !trb.chain {
+                    self.transfer_state.state = ControlTransferStage::ExpectSetupStageTrb;
+                }
+
+                if trb.interrupt_on_completion {
+                    interrupt_on_completion(
+                        address,
+                        CompletionCode::Success,
+                        false,
+                        self.endpoint_id,
+                        self.slot_id,
+                        &self.event_sender,
+                    )?;
+                }
+                Ok(())
+            }
         }
     }
 }
 
-#[derive(Debug, Default)]
+/// Track communication between us and the host hardware.
+#[derive(Debug, Default, Clone)]
 enum ControlSubmissionState {
     #[default]
     NoTrbSubmitted,
-    ParserConsumedTrb,
-    // store address of trb that failed to parse.
-    // needs to be specified inside the transfer event indicating the error.
+    ParserConsumedTrb(u64, TransferTrbVariant),
     ParserError(u64),
-    AwaitingControlIn(UsbRequest),
-    AwaitingControlOut(UsbRequest),
+    AwaitingControlIn(u64, TransferTrbVariant),
+    AwaitingControlOut(u64, TransferTrbVariant),
 }
 
 impl<RCEH: RealControlEndpointHandle> EndpointHandle for ControlEndpointHandle<RCEH> {
@@ -130,28 +470,91 @@ impl<RCEH: RealControlEndpointHandle> EndpointHandle for ControlEndpointHandle<R
         Pin<Box<dyn Future<Output = anyhow::Result<TrbProcessingResult>> + Send + 'a>>;
 
     fn submit_trb(&mut self, trb: RawTrb) -> anyhow::Result<()> {
-        let trb_address = trb.address;
-        if let ControlFlow::Break(res) = self.trb_parser.trb(trb) {
-            match res {
-                Ok(request) => {
-                    let request_copy = request.clone_without_data();
-                    let is_out_request = request.request_type & 0x80 == 0;
+        let variant = TransferTrbVariant::parse(trb.buffer);
 
-                    pcap::control_submission(self.pcap_meta, &request);
+        if let TransferTrbVariant::Unrecognized(_, _) = &variant {
+            // logging this is happening in next_completion()
+            self.submission_state = ControlSubmissionState::ParserError(trb.address);
+        }
 
-                    self.real_ep.submit_control_request(request)?;
-
-                    self.submission_state = match is_out_request {
-                        true => ControlSubmissionState::AwaitingControlOut(request_copy),
-                        false => ControlSubmissionState::AwaitingControlIn(request_copy),
-                    };
+        match &self.transfer_state.state {
+            ControlTransferStage::ExpectSetupStageTrb => match variant {
+                TransferTrbVariant::SetupStage(setup_stage) => {
+                    self.handle_setup_stage_trb(trb.address, setup_stage)?;
                 }
-                Err(_) => {
-                    self.submission_state = ControlSubmissionState::ParserError(trb_address);
+                other_trb => {
+                    info!(
+                        "invalid control transfer sequence; expected Setup Stage Trb, got: {:?}",
+                        other_trb
+                    );
+                    self.submission_state =
+                        ControlSubmissionState::ParserConsumedTrb(trb.address, other_trb);
                 }
-            }
-        } else {
-            self.submission_state = ControlSubmissionState::ParserConsumedTrb;
+            },
+            ControlTransferStage::MaybeDataStageTrb => match variant {
+                TransferTrbVariant::SetupStage(setup_stage) => {
+                    info!(
+                        "received Setup Stage TRB abort ongoing control transfer in favour of this new one"
+                    );
+                    self.handle_setup_stage_trb(trb.address, setup_stage)?;
+                }
+                TransferTrbVariant::DataStage(data_stage) => {
+                    self.handle_data_stage_trb(trb.address, data_stage)?;
+                }
+                TransferTrbVariant::StatusStage(status_stage) => {
+                    self.handle_status_stage_trb(trb.address, status_stage)?;
+                }
+                other_trb => {
+                    info!(
+                        "invalid control transfer sequence; expected Setup, Data or Status Stage Trb, got: {:?}",
+                        other_trb
+                    );
+                    self.transfer_state.state = ControlTransferStage::ExpectSetupStageTrb;
+                    self.submission_state =
+                        ControlSubmissionState::ParserConsumedTrb(trb.address, other_trb);
+                }
+            },
+            ControlTransferStage::MoreData => match variant {
+                TransferTrbVariant::SetupStage(setup_stage) => {
+                    info!(
+                        "received Setup Stage TRB abort ongoing control transfer in favour of this new one"
+                    );
+                    self.handle_setup_stage_trb(trb.address, setup_stage)?;
+                }
+                TransferTrbVariant::Normal(normal) => {
+                    self.handle_normal_trb(trb.address, normal)?;
+                }
+                other_trb => {
+                    info!(
+                        "invalid control transfer sequence; expected Setup Stage, Normal or Event Data Trb, got: {:?}",
+                        other_trb
+                    );
+                    self.transfer_state.state = ControlTransferStage::ExpectSetupStageTrb;
+                    self.submission_state =
+                        ControlSubmissionState::ParserConsumedTrb(trb.address, other_trb);
+                }
+            },
+
+            ControlTransferStage::ExpectStatusStageTrb => match variant {
+                TransferTrbVariant::SetupStage(setup_stage) => {
+                    info!(
+                        "received Setup Stage TRB abort ongoing control transfer in favour of this new one"
+                    );
+                    self.handle_setup_stage_trb(trb.address, setup_stage)?;
+                }
+                TransferTrbVariant::StatusStage(status_stage) => {
+                    self.handle_status_stage_trb(trb.address, status_stage)?;
+                }
+                other_trb => {
+                    info!(
+                        "invalid control transfer sequence; expected Setup or Status Stage Trb, got: {:?}",
+                        other_trb
+                    );
+                    self.transfer_state.state = ControlTransferStage::ExpectSetupStageTrb;
+                    self.submission_state =
+                        ControlSubmissionState::ParserConsumedTrb(trb.address, other_trb);
+                }
+            },
         }
 
         Ok(())
@@ -159,12 +562,19 @@ impl<RCEH: RealControlEndpointHandle> EndpointHandle for ControlEndpointHandle<R
 
     fn next_completion(&mut self) -> Self::TrbCompletionFuture<'_> {
         Box::pin(async {
-            let result = match self.submission_state {
-                ControlSubmissionState::ParserConsumedTrb => TrbProcessingResult::Ok,
-                ControlSubmissionState::ParserError(trb_address) => {
-                    pcap::trb_error(self.pcap_meta, trb_address);
+            let result = match self.submission_state.clone() {
+                ControlSubmissionState::ParserConsumedTrb(address, variant) => {
+                    trace!("consumed trb from address: {} as: {:?}", address, variant);
+                    TrbProcessingResult::Ok
+                }
+                ControlSubmissionState::ParserError(address) => {
+                    info!(
+                        "Failed to parse Transfer Trb on Control Endpoint. slot {}",
+                        self.slot_id
+                    );
+                    pcap::trb_error(self.pcap_meta, address);
                     let event = EventTrb::new_transfer_event_trb(
-                        trb_address,
+                        address,
                         0,
                         CompletionCode::TrbError,
                         false,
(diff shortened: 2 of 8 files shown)


src/device/xhci/endpoint_handle.rs
```diff
@@ -745,3 +1321,1039 @@ impl<RIEH: RealInEndpointHandle> BaseEndpointHandle for InEndpointHandle<RIEH> {
...
+    const BM_REQUEST_TYPE_OUT: u8 = 0;
+
+    const TRB_TYPE_NORMAL: u8 = 0x1;
+    const TRB_TYPE_SETUP_STAGE: u8 = 0x2;
+    const TRB_TYPE_DATA_STAGE: u8 = 0x3;
+    const TRB_TYPE_STATUS_STAGE: u8 = 0x4;
+    const TRB_TYPE_EVENT_DATA: u8 = 0x7;
+
+    const TRT_OUT_DATA: u8 = 0x2;
+    const TRT_IN_DATA: u8 = 0x3;
```
snue: nit: what is "TRT"?

src/device/xhci/endpoint_handle.rs
```diff
@@ -745,3 +1321,1039 @@ impl<RIEH: RealInEndpointHandle> BaseEndpointHandle for InEndpointHandle<RIEH> {
...
+            .with_dir()
+            .build();
+
+        control_endpoint
+            .submit_trb(setup_stage)
+            .expect("this dummy hardware request should never fail");
+        control_endpoint
+            .next_completion()
+            .await
+            .expect("this dummy hardware request should never fail");
```
snue: Why do you get the `next_completion` only after the setup stage? I rather like the approach where you define the vector of TRBs and just submit them in a loop (as you do for the bulk_in transfer test with event data).

src/device/xhci/trb.rs
```diff
@@ -1088,10 +1088,201 @@ pub enum TrbParseError {
     UnsupportedOptionalCommand(u8, String),
     #[error("TRB type {0} does not refer to any command.")]
     UnknownTrbType(u8),
-    #[error("Detected a non-zero value in a RsvdZ field")]
+    #[error("Detected a non-zero value in a RsvdZ field.")]
```
snue: We didn't really spell it out anywhere, but we should follow what either the standard library does (lowercase with no punctuation), or what cloud-hypervisor does (capitalized first letter, then lowercase with no punctuation). We currently have a pretty wild mix, but adding punctuation is not the way to fix it.
 
https://github.com/rust-lang/api-guidelines/issues/79

src/device/xhci/interrupter.rs
```diff
@@ -55,7 +55,7 @@ struct EventWorker {
 }
 
 #[derive(Debug)]
-enum InterrupterMessage {
+pub enum InterrupterMessage {
```
snue: What exactly is the required scope here? The message was an internal detail so far and isn't even documented. Can we limit this to `pub(crate)` maybe?

src/device/xhci/trb.rs
```diff
@@ -1088,10 +1088,201 @@ pub enum TrbParseError {
...
+        }
+
+        /// immediate data bit
+        pub fn with_idt(mut self) -> Self {
+            self.buffer[12] |= Self::IDT;
+            self
+        }
+
+        /// trb type field
+        pub fn with_trb_type(mut self, trb_type: u8) -> Self {
```
snue: nit: can be shortened to

```suggestion
        pub fn with_type(mut self, trb_type: u8) -> Self {
```

src/device/xhci/trb.rs
```diff
@@ -1088,10 +1088,201 @@ pub enum TrbParseError {
...
+
+        /// first 8 bytes of the trb
+        pub fn with_data_field(mut self, value: u64) -> Self {
+            let value_bytes: [u8; 8] = value.to_le_bytes();
+            self.buffer[..8].copy_from_slice(&value_bytes);
+            self
+        }
+
+        /// byte index 8 & 9
+        pub fn with_transfer_length(mut self, length: u16) -> Self {
```
snue: nit: can be shortened to

```suggestion
        pub fn with_length(mut self, length: u16) -> Self {
```

src/device/xhci/endpoint_handle.rs
```diff
@@ -745,3 +1321,1039 @@ impl<RIEH: RealInEndpointHandle> BaseEndpointHandle for InEndpointHandle<RIEH> {
...
+        control_endpoint
+            .submit_trb(data_stage)
+            .expect("this dummy hardware request should never fail");
+        control_endpoint
+            .submit_trb(status_stage)
+            .expect("this dummy hardware request should never fail");
+        control_endpoint
+            .next_completion()
+            .await
+            .expect("this dummy hardware request should never fail");
```
snue: I really don't see how you decide when to call `next_completion` 🤔

src/device/xhci/endpoint_handle.rs
```diff
@@ -642,88 +1255,51 @@ impl<RIEH: RealInEndpointHandle> EndpointHandle for InEndpointHandle<RIEH> {
...
-                    let (completion_code, processing_result) = match self
-                        .real_ep
-                        .next_completion()
-                        .await?
-                    {
+                NormalSubmissionState::AwaitingRealTransfer(transfer_trb) => {
+                    debug!("NormalSubmissionState::AwaitingRealTransfer");
+                    match self.real_ep.next_completion().await? {
                         InTrbProcessingResult::Disconnect => {
+                            warn!("NormalSubmissionState::AwaitingRealTransfer InTrbProcessingResult::Disconnect");
```
snue: For all the messages that should stay and are not just for debugging, let's make sure they are meaningful to a reader and don't just mention internal state names. The messages will be accompanied by the file/line number so it's not an issue figuring out where they are from.

For this message here, I think it is fully expected (the device probably got removed) and should just be `info`.

```suggestion
                            info!("Detected device disconnect");
```

Maybe we could add a hint which device this is for, but I consider that optional for now.

src/device/xhci/endpoint_handle.rs
```diff
@@ -591,7 +1088,112 @@ impl<RIEH: RealInEndpointHandle> InEndpointHandle<RIEH> {
...
+                completion_code = CompletionCode::ShortPacket;
+                hardware_data.len()
+            }
+            Ordering::Equal => {
+                debug!("received exactly as requested");
+                completion_code = CompletionCode::Success;
+                hardware_data.len()
+            }
+            Ordering::Greater => {
+                warn!("received more than requested");
```
snue: This is what #246 and #248 should address. Can we make this a highly visible `error!()` as I am pretty sure this leads to data loss/corruption currently.

src/device/xhci/endpoint_handle.rs
```diff
@@ -591,7 +1088,112 @@ impl<RIEH: RealInEndpointHandle> InEndpointHandle<RIEH> {
...
+                completion_code = CompletionCode::Success;
+                // device responded with more than requested
+                // idk where the overhead goes but we track the requested amount
+                normal_trb_data.transfer_length as usize
```
snue: Maybe we should even log the extra bytes here for now? (the `hardware_data[transfer_length..]` part)

src/device/xhci/endpoint_handle.rs
```diff
@@ -291,89 +724,100 @@ impl<RCEH: RealControlEndpointHandle> ControlEndpointHandle<RCEH> {
...
+        true,
+        endpoint_id,
+        slot_id,
+    );
+
+    event_sender.send(event)?;
+    normal_transfer_state.edtla = 0;
+
+    // It was not clear from the specification alone if the IOC bit is
+    // actually intended for the above event or as this separate one.
```
snue, line 916: This comment is currently not helpful. Let's rather just state what this does now. We handle IOC the same as for any other TRBs.

src/device/xhci/endpoint_handle.rs
```diff
@@ -291,89 +724,100 @@ impl<RCEH: RealControlEndpointHandle> ControlEndpointHandle<RCEH> {
...
+    normal_transfer_state: &mut EventDataTrbMetadata,
+    endpoint_id: u8,
+    slot_id: u8,
+    event_sender: &EventSender,
+) -> anyhow::Result<()> {
+    trace!("EventData TRB");
+
+    let event_data_trb_data = match &transfer_trb.variant {
+        TransferTrbVariant::EventData(event_data_trb_data) => event_data_trb_data,
+        _ => unreachable!("checked variant before calling this handle"),
```
snue: This redundant naming is exhausting.

```suggestion
    let event_data = match &transfer_trb.variant {
        TransferTrbVariant::EventData(data) => data,
```

Or even better, as you have all these `unreachable!()` branches in the called functions, can we modify the signature to take the embedded data directly and fix this in the caller?

src/device/xhci/endpoint_handle.rs
```diff
@@ -383,6 +827,7 @@ pub struct OutEndpointHandle<ROEH: RealOutEndpointHandle> {
     dma_bus: BusDeviceRef,
     event_sender: EventSender,
     submission_state: NormalSubmissionState,
+    event_data_trb_metadata: EventDataTrbMetadata,
```
snue: nit: how about `event_meta`?

src/device/xhci/endpoint_handle.rs
```diff
@@ -591,7 +1088,112 @@ impl<RIEH: RealInEndpointHandle> InEndpointHandle<RIEH> {
...
+
+        if !normal_trb_data.chain {
+            self.event_data_trb_metadata = EventDataTrbMetadata::new();
+        }
+
+        self.real_ep
+            .submit(normal_trb_data.transfer_length as usize)?;
+        pcap::in_submission(
+            self.pcap_meta,
+            transfer_trb.address,
```
snue: You only need the full `TransferTrb` because you need this address. For everything else the encapsulated data (`normal_trb_data`) is sufficient. Maybe we can move the address down there and avoid passing the full `TransferTrb` here and the additional parsing above. Or we bite the apple and just hand the address as extra argument.

snue (commented):
Did a partial review and left some comments. I really like the added unit tests.

Most importantly I'd like to get rid of most of the `unreachable!()` statements through code refactoring. And we also need to clarify what we do about the device overread / TD handling.

I also left lots of nits about naming and such. You really like long redundant variable names 😸

src/device/xhci/trb.rs
snue: And we leave this cleanup for a different PR of course :wink:

src/device/xhci/interrupter.rs
lbeierlieb: So far, InterrupterMessage could stay internal because I created methods on the EventSender, which wraps the Sender<InterrupterMessage> and creates the correct messages. And not just the InterrupterMessage is an implementation detail that I kept internal to the Interrupter, the fact that communication goes through a bounded channel is also hidden. I applied the same pattern of keeping the communication between Senders and the components internal throughout the codebase.

The tests now manually instantiate channels and access the Receiver<InterrupterMessage> to check the values sent through the sender.

To not unnecessarily expose implementation details, I would suggest to implement a test-only function in the Interrupter that returns a tuple (EventSender, InterrupterMock). The communication between the two is kept private to the interrupter file. Tests throughout the codebase can use InterrupterMock::assert_event_sent(&EventTrb) or similar.

pub(crate) seems like a quick fix, but we spread the details about the communication mechanism throughout the codebase (almost all xhci components need an EventSender and we are likely to create tests for more of those components in the future). Changes, e.g., switching away from using an UnboundedChannel will need to touch a lot of code.

src/device/xhci/trb.rs
Noi0103: I would like to vote to follow the standard library.

Yes, I will leave it as is for this PR, so it is similar to the preceding lines.

src/device/xhci/endpoint_handle.rs
Noi0103: Yes, this makes sense for the ControlEndpoint as well, will be changed.

src/device/xhci/endpoint_handle.rs
Noi0103: I called it whenever I knew I would need it. I'll change it to be the other way around and call it always, unless it should not be done.

src/device/xhci/endpoint_handle.rs
Noi0103: > The messages will be accompanied by the file/line number so it's not an issue figuring out where they are from.

The logs are only accompanied by the file. If multiple identical messages for similar cases are used (like this case would be), it will be an issue to figure out where exactly it came from. I will add endpoint and slot_id to your suggestion.

src/device/xhci/endpoint_handle.rs
Noi0103: I went with passing the address as an extra argument. I like that every field we have inside the parsed TransferTrb exists in the specifications tables that describe the fields of these 16 byte.
If we really want to move the address into the parsed Trb structures, we can do that instead. I wanted to do this first and get opinions.

src/device/xhci/endpoint_handle.rs
snue: Right, we omit the line numbers currently. So better avoid completely identical messages in the same file for now. We could of course configure the formatter to include line numbers as well. This really depends on what we want "production" logs to look like later.

src/device/xhci/trb.rs
```diff
@@ -1092,6 +1092,197 @@ pub enum TrbParseError {
...
+
+        pub fn build(self) -> RawTrb {
+            RawTrb {
+                address: self.address,
+                buffer: self.buffer,
+            }
+        }
+    }
+
+    //TODO    #[cfg(test)]
```
snue: This is already guarded by `#[cfg(test)]` due to the parent module `testutils`?

src/device/xhci/trb.rs
```diff
@@ -164,14 +164,14 @@ impl PortStatusChangeEventTrbData {
...
-    completion_code: CompletionCode,
-    event_data: bool,
-    endpoint_id: u8,
-    slot_id: u8,
+    pub trb_pointer: u64,
+    pub trb_transfer_length: u32,
+    pub completion_code: CompletionCode,
+    pub event_data: bool,
+    pub endpoint_id: u8,
+    pub slot_id: u8,
```
snue: The commit message still mentions public constructors, but they are not part of this commit? They are probably not necessary anymore due to @lbeierlieb's proposed handling?

I don't see where the need for `pub` on these struct members comes from. Probably a different test? I'd rather avoid that as well, or make it `pub(crate)` at least?

Noi0103: Yes, my mistake, after changing to the proposed handling by @lbeierlieb the constructor was obsolete.

Not all of those pub fields are used, but all are pub to not have it half pub.
The `pointer` and `trb_transfer_length` are the only ones used in the tests. Should I instead use a test only method to get these two fields?
```rust
assert_ne!(trb_data.trb_pointer, THIRD_ADDRESS);
assert_eq!(trb_data.trb_pointer, EVENT_DATA_FIELD);
assert_eq!(trb_data.trb_transfer_length, 1536);
```

src/device/xhci/trb.rs
Noi0103: Yes, thanks, I forgot to remove that comment.

src/device/xhci/endpoint_handle.rs
```diff
@@ -1367,3 +1367,1052 @@ impl<RIEH: RealInEndpointHandle> BaseEndpointHandle for InEndpointHandle<RIEH> {
...
+
+        // The incomplete sequence (the second trb; a lone setup stage) is valid
+        // and we expect an event.
+        let event_trb = interrupter.next_event();
+        assert!(event_trb.is_some());
+        if let EventTrb::Transfer(trb_data) = event_trb.unwrap() {
+            assert_eq!(trb_data.trb_pointer, SECOND_ADDRESS);
+        } else {
+            self::panic!("not a transfer event")
+        }
```
snue: Can we construct the expected event instead and just `assert_eq!(interrupter.next_event(), Some(expected))` to get around making the TRB details public?

snue: This could also be looped then.

src/device/xhci/trb.rs
snue: Yes, either have a test-only method, or see my remark below. You can probably construct the expected TRBs (maybe through a helper) and compare them directly. We also don't need to assert on the `trb_pointer` twice, but I get what you want to say with this. A fitting comment could go to the construction of the expected data instead.

src/device/xhci/endpoint_handle.rs
```diff
@@ -85,66 +141,508 @@ impl<RCEH: RealControlEndpointHandle> ControlEndpointHandle<RCEH> {
...
+                    info!(
+                        "received Setup Stage TRB abort ongoing control transfer in favour of this new one"
+                    );
+                    self.handle_setup_stage_trb_pre_hardware(address, setup_stage)?;
+                }
+                TransferTrbVariant::DataStage(data_stage) => {
+                    self.handle_data_stage_trb(address, data_stage)?;
+                }
+                TransferTrbVariant::Normal(normal) => {
+                    todo!("Normal Trb in a Control Chain: {:?}", normal);
```
snue: Is this allowed? Can we handle this gracefully instead?

src/device/xhci/endpoint_handle.rs
```diff
@@ -59,16 +65,66 @@ impl BaseEndpointHandle for DummyEndpointHandle {
...
+// Track how far we are with parsing the Control Transfer (chain of TRB).
+#[derive(Debug, PartialEq, Eq)]
+pub enum ControlTransferStage {
+    /// Nothing happened yet. Awaiting a Setup Stage Trb and dropping any other
+    /// Trb (they will not reach the hardware device).
+    ExpectSetupStageTrb,
+    /// Either collect data if a Data Stage Trb is received or skip the Data
+    /// Stage TD altogether if a Status Stage Trb is received.
+    MaybeDataStageTrb,
+    /// Collect or provide more Data slices through dma operations.
```
snue: nit: Or immediate data? And we should capitalize DMA.

src/device/xhci/endpoint_handle.rs
```diff
@@ -12,19 +13,24 @@ use crate::device::{
...
-        trb::{CompletionCode, EventTrb, RawTrb, TransferTrb, TransferTrbVariant},
+        trb::{
+            CompletionCode, DataStageTrbData, EventDataTrbData, EventTrb, NormalTrbData, RawTrb,
+            SetupStageTrbData, StatusStageTrbData, TransferTrb, TransferTrbVariant,
+        },
         usbrequest::UsbRequest,
     },
 };
 
+pub const MAX_VALUE_U24: u64 = 0xffffff;
```
snue: nit: this is easier to read with the allowed separator.

```suggestion
pub const MAX_VALUE_U24: u64 = 0xff_ffff;
```

src/device/xhci/endpoint_handle.rs
```diff
@@ -281,99 +765,106 @@ impl<RCEH: RealControlEndpointHandle> ControlEndpointHandle<RCEH> {
...
-                    }
-                    _ => return ControlFlow::Break(Err(())),
-                },
-            }
+/// When an Event Data is encountered two additional things are needed:
+/// - the EDTLA, counting already transmitted bytes of the current TD
+/// - the completion code of the previously handled TRB
+#[derive(Debug)]
+pub struct EventDataTrbMetadata {
+    edtla: u64,
```
snue: Does storing it as `u64` instead of `u32` avoid a lot of conversions? It's defined as a 24 bit value, obviously.

src/device/xhci/endpoint_handle.rs
```diff
@@ -59,16 +65,66 @@ impl BaseEndpointHandle for DummyEndpointHandle {
...
+    /// Either collect data if a Data Stage Trb is received or skip the Data
+    /// Stage TD altogether if a Status Stage Trb is received.
+    MaybeDataStageTrb,
+    /// Collect or provide more Data slices through dma operations.
+    MoreData,
+    /// Finished processing the Data Stage if there was one.
+    ExpectStatusStageTrb,
+    /// Status Stage TRB had a chain bit and there will be exactly one
+    /// Event Data Trb to finish the Control Transfer.
+    ExpectLastEventDataTrb,
```
snue: nit: I would call it `Final` instead of `Last`.

```suggestion
    ExpectFinalEventDataTrb,
```

src/device/xhci/endpoint_handle.rs
```diff
@@ -59,16 +66,113 @@ impl BaseEndpointHandle for DummyEndpointHandle {
...
     endpoint_id: u8,
     pcap_meta: EndpointPcapMeta,
     real_ep: RCEH,
-    trb_parser: ControlRequestParser,
     dma_bus: BusDeviceRef,
     event_sender: EventSender,
+    /// referring to usbvfiod to hardware communication
     submission_state: ControlSubmissionState,
+    /// referring to usbvfiod to guest communication
+    control_transfer_state: ControlTransferState,
```
snue: For consistency, this should just be `transfer_state`

src/device/xhci/endpoint_handle.rs
```diff
@@ -85,66 +189,572 @@ impl<RCEH: RealControlEndpointHandle> ControlEndpointHandle<RCEH> {
...
+                    data: None,
+                },
+            )),
+        }
+    }
+
+    fn handle_setup_stage_trb_pre_hardware(
+        &mut self,
+        address: u64,
+        setup_stage_trb: SetupStageTrb,
```
snue: nit: I think we can just rename this `trb`. The type and function name are explicit already, Same for the `post_hardware` part of course.

src/device/xhci/endpoint_handle.rs
```diff
@@ -85,66 +189,572 @@ impl<RCEH: RealControlEndpointHandle> ControlEndpointHandle<RCEH> {
...
+                    )?;
+                }
+
+                self.control_transfer_state.state = ControlTransferStage::MaybeDataStageTrb;
+                self.control_transfer_state.event_meta.zero_edtla();
+            }
+            ControlTransferData::Out(_) => {
+                unreachable!(
+                    "ControlOut SetupTrb have insufficient information to do the Hardware request so you never set a submission state that enables arriving here."
+                )
```
snue: But the `pre_hardware()` and `post_hardware()` naming makes it sound like these always go in a pair. I think I would prefer a regular `handle_setup_stage_trb` (dropping the `_pre_hardware`) and making this a `handle_setup_stage_hardware_response`. And then maybe the signature can take all the required info (the request) and we get rid of the `unreachable!` statement completely.

src/device/xhci/endpoint_handle.rs
```diff
@@ -85,66 +189,572 @@ impl<RCEH: RealControlEndpointHandle> ControlEndpointHandle<RCEH> {
...
+            }
+        }
+        Ok(())
+    }
+
+    fn data_slices<T: CanDoDma>(&mut self, address: u64, trb: &T) -> anyhow::Result<()> {
+        let data_pointer = trb.get_data_pointer();
+        let transfer_length = trb.get_trb_transfer_length();
+        let immediate_data = trb.get_immediate_data();
```
snue: Let's not use the `get_` naming conventions from Java getter/setters.


```suggestion
        let data_pointer = trb.data_pointer();
        let transfer_length = trb.transfer_length();
        let immediate_data = trb.immediate_data();
```

And for the `immediate_data()`, as that presumably is a `bool` and not the data, `has_immediate_data()` makes that clear.

src/device/xhci/endpoint_handle.rs
```diff
@@ -59,16 +66,113 @@ impl BaseEndpointHandle for DummyEndpointHandle {
...
+        }
+    }
+
+    const fn zero_edtla(&mut self) {
+        self.edtla = 0;
+    }
+    /// input is 17 Bit
+    const fn increase_edtla(&mut self, addend: u32) {
+        self.edtla = MAX_VALUE_U24 & (self.edtla.wrapping_add(addend));
+    }
```
snue: nit: these are in their own type and don't really need to mention the member name. I'd just call them `zero()` and `add()`, which makes the repeated use below look a little nicer.

src/device/xhci/endpoint_handle.rs
```diff
@@ -12,19 +14,24 @@ use crate::device::{
...
-        trb::{CompletionCode, EventTrb, RawTrb, TransferTrb, TransferTrbVariant},
+        trb::{
+            CanDoDma, CompletionCode, DataStageTrb, EventDataTrb, EventTrb, NormalTrb, RawTrb,
+            SetupStageTrb, StatusStageTrb, TransferTrbVariant,
+        },
         usbrequest::UsbRequest,
     },
 };
 
+pub const MAX_VALUE_U24: u32 = 0xffffff;
```
snue: nit: could use a `_`

```suggestion
pub const MAX_VALUE_U24: u32 = 0xff_ffff;
```

src/device/xhci/endpoint_handle.rs
```diff
@@ -85,66 +189,572 @@ impl<RCEH: RealControlEndpointHandle> ControlEndpointHandle<RCEH> {
...
+    ) -> anyhow::Result<()> {
+        match &mut self.control_transfer_state.data {
+            // collect hardware data
+            ControlTransferData::In(request) => {
+                trace!("control in data {:?}", hardware_data);
+
+                pcap::control_completion_in(self.pcap_meta, request.address, hardware_data);
+
+                // SAFETY: is always set in the preceding setup stage pre hardware part
+                request.data.as_mut().unwrap().append(hardware_data);
```
snue: What if we just demand this to be set instead of making it an option and then generating empty vectors on the fly below?

src/device/xhci/endpoint_handle.rs
```diff
@@ -85,66 +189,572 @@ impl<RCEH: RealControlEndpointHandle> ControlEndpointHandle<RCEH> {
...
+                // Data Stage TD (i.e. the sum of the Length fields of the Data Stage TRB and all Normal
+                // TRBs) is equal to wLength. Note that communicating with some non-compliant
+                // devices may require violating this rule.
+                if usb_request.data.clone().unwrap_or_else(|| vec![0; 0]).len()
+                    < transfer_length as usize
+                {
+                    self.control_transfer_state.state = ControlTransferStage::ExpectStatusStageTrb;
+                    self.control_transfer_state.event_meta.zero_edtla();
+                    self.submission_state = ControlSubmissionState::ParserError(address); // TODO maybe protocol error?
+                    return Ok(());
```
snue: Why does this method return `Ok()` on every single path? What happens when we just let this return an error?

src/device/xhci/endpoint_handle.rs
```diff
@@ -85,66 +189,572 @@ impl<RCEH: RealControlEndpointHandle> ControlEndpointHandle<RCEH> {
...
+
+                Ok(())
+            }
+        }
+    }
+
+    fn handle_data_stage_trb(
+        &mut self,
+        address: u64,
+        data_stage_trb: DataStageTrb,
```
snue: nit: same here, naming seems redundant. `trb` should be good enough.

src/device/xhci/endpoint_handle.rs
snue: Did you change the comment?

src/device/xhci/endpoint_handle.rs
```diff
@@ -471,83 +1099,69 @@ impl<ROEH: RealOutEndpointHandle> EndpointHandle for OutEndpointHandle<ROEH> {
...
-                            transfer_trb.address,
-                            0,
-                            completion_code,
-                            false,
-                            self.endpoint_id,
-                            self.slot_id,
-                        );
-                        self.event_sender.send(transfer_event)?;
-                    }
-
```
snue, line 579: This used to generate events for all completion codes. You explicitly do it for stall and omit all the others. Is this intended?

src/device/xhci/endpoint_handle.rs
```diff
@@ -642,90 +1369,57 @@ impl<RIEH: RealInEndpointHandle> EndpointHandle for InEndpointHandle<RIEH> {
...
-                        let transfer_event = EventTrb::new_transfer_event_trb(
-                            transfer_trb.address,
-                            0,
-                            completion_code,
-                            false,
-                            self.endpoint_id,
-                            self.slot_id,
-                        );
-
-                        self.event_sender.send(transfer_event)?;
```
snue: Same here, no more events for all the TRBs?

src/device/xhci/endpoint_handle.rs
```diff
@@ -628,7 +1346,16 @@ impl<RIEH: RealInEndpointHandle> EndpointHandle for InEndpointHandle<RIEH> {
         );
 
         Box::pin(async {
-            let result = match self.submission_state {
+            let result = match self.submission_state.clone() {
```
snue: Why does this need a `clone()` suddenly?

src/device/xhci/trb.rs
```diff
@@ -848,20 +840,27 @@ impl TransferTrbVariant {
     }
 }
 
+pub trait CanDoDma {
```
snue: nit: maybe `TrbDmaInfo`

src/device/xhci/trb.rs
```diff
@@ -848,20 +840,27 @@ impl TransferTrbVariant {
...
+    fn get_data_pointer(&self) -> u64;
+    fn get_trb_transfer_length(&self) -> u32;
+    fn get_immediate_data(&self) -> bool;
```
snue: Can these be blanket implementations? It's always accessing the same members, right?

snue (commented):
I have a bunch of comments stored up. We are getting there, but I think you missed treating some of the error paths correctly? Could these be caught by unit tests as well?

A lot of my comments are nits for the documentation or name suggestions. I see some of my previously suggested renames are for pre-existing data structures (dropping the redundant `Data` part). I still think these are better names, but obviously it makes the diff larger. Sorry for that.

There are also some comments on outdated sections that I can't figure out how to edit/remove through the GitHub UI.

src/device/xhci/endpoint_handle.rs
Noi0103: > makes it sound like these always go in a pair

I did not think of it that way, thanks, that really needs changing.
I don't think (or better: I don't know how) we can get rid of the unreachable here while making the code cleaner.

src/device/xhci/endpoint_handle.rs
snue: This is likely Ok, we use the errors to catch unrecoverable things like broken communication channels., which will lead to a task terminating somewhere. Regular operations are all based on `Ok` values. But we didn't really spell that out anywhere so far.

src/device/xhci/trb.rs
Noi0103: Afaik this is not possible even with both structs having these fields.

src/device/xhci/endpoint_handle.rs
Noi0103: Since the &mut self is used inside the match arms through method calls, the compiler complained and proposed this.

src/device/xhci/endpoint_handle.rs
Noi0103: You are right it does not need a Result, I blindly threw the anyhow error on this.

The actual result of this is kind of obfuscated by putting it into the state machine fields.

src/device/xhci/endpoint_handle.rs
Noi0103: No. I might have made a mistake when doing a rebase. Thanks for finding.

Noi0103:
Tests with a DummyControlEndpoint that would return specific errors are not yet included.

src/device/xhci/nusb.rs
snue:
```suggestion
                        data: &request.data,
```
Instead of the `let data` above?

src/device/xhci/endpoint_handle.rs
```diff
@@ -59,16 +66,143 @@ impl BaseEndpointHandle for DummyEndpointHandle {
...
+            previous_completion_code: CompletionCode::Success,
+        }
+    }
+
+    const fn zero(&mut self) {
+        self.edtla = 0;
+    }
+    /// input is 17 Bit
+    const fn add(&mut self, addend: u32) {
+        self.edtla = MAX_VALUE_U24 & (self.edtla.wrapping_add(addend));
```
snue: The `addend` name confuses me 😅  


```suggestion
        self.edtla = MAX_VALUE_U24 & (self.edtla.wrapping_add(transfered_bytes));
```

src/device/xhci/endpoint_handle.rs
```diff
@@ -59,16 +66,143 @@ impl BaseEndpointHandle for DummyEndpointHandle {
...
+#[derive(Debug, PartialEq, Eq)]
+pub struct ControlTransferState {
+    /// upcoming or current stage/TD of a control transfer to be handled
+    pub state: ControlTransferStage,
+    /// holding the UsbRequest and all associated data
+    pub data: ControlTransferData,
+    event_meta: EventDataTrbMetadata,
+}
+impl ControlTransferState {
+    // previous_completion_code should never be used as is, thus the error as a default value
```
snue: This is a remnant?

src/device/xhci/endpoint_handle.rs
```diff
@@ -85,66 +219,507 @@ impl<RCEH: RealControlEndpointHandle> ControlEndpointHandle<RCEH> {
...
+
+                    // SAFETY: is always set in the preceding setup stage
+                    usb_request.data.append(
+                        &mut data_pointer.to_le_bytes()[..transfer_length as usize].to_vec(),
+                    );
+                } else {
+                    let mut tmp = vec![0u8; transfer_length as usize];
+                    self.dma_bus.read_bulk(data_pointer, &mut tmp);
+
+                    // SAFETY: is always set in the preceding setup stage
```
snue: This is actually safe now.

src/device/xhci/endpoint_handle.rs
```diff
@@ -85,66 +219,507 @@ impl<RCEH: RealControlEndpointHandle> ControlEndpointHandle<RCEH> {
...
+                ControlTransferStage::MoreData => {
+                    self.transfer_state.state = ControlTransferStage::ExpectStatusStageTrb;
+                    self.transfer_state.event_meta.zero();
+                }
+                ControlTransferStage::ExpectFinalEventDataTrb => {
+                    self.transfer_state.state = ControlTransferStage::ExpectSetupStageTrb;
+                    self.transfer_state.event_meta.zero();
+                }
+                _ => {
+                    error!("driver did not provide a spec ocmpliant control transfer trb chain");
```
snue:
```suggestion
                    error!("driver did not provide a spec compliant control transfer trb chain");
```

snue (commented):
I have not done a full review, but it looks like we are getting there. The remaining unreachable statements are well explained and can be trusted. Left some comments, found typos etc.

src/device/xhci/endpoint_handle.rs
```diff
@@ -591,7 +1191,103 @@ impl<RIEH: RealInEndpointHandle> InEndpointHandle<RIEH> {
...
+        &mut self,
+        address: u64,
+        trb: NormalTrb,
+        hardware_data: Vec<u8>,
+    ) -> anyhow::Result<()> {
+        trace!("handle_normal_trb_post_hardware In");
+
+        let completion_code: CompletionCode;
+
+        // SACETY: in case the hardware_data.len() is bigger than u32 we take the u32 value regardless
```
snue: typo:
```suggestion
        // SAFETY: in case the hardware_data.len() is bigger than u32 we take the u32 value regardless
```

src/device/xhci/endpoint_handle.rs
```diff
@@ -85,66 +219,507 @@ impl<RCEH: RealControlEndpointHandle> ControlEndpointHandle<RCEH> {
...
+
+    fn handle_status_stage_trb_hardware_response(
+        &mut self,
+        address: u64,
+        trb: StatusStageTrb,
+    ) -> anyhow::Result<()> {
+        match &mut self.transfer_state.data {
+            ControlTransferData::In(_) => {
+                unreachable!("internal error: ControlIn requests do the Hardware request in the SetupStage; a submission state to arrive here should never be used");
+            }
```
snue: Ah, we put so much effort into ensuring you can only call this with the right kind of TRB, but I think it would make much more sense if we had a single `handle_hardware_response` function that does the right thing depending on `In`/`Out`, instead of splitting it and having `unreachable!` cases in the respective functions. I may be missing something here, though.

Noi0103: If we were to put them together we would at least have to
- change the two specific `trb` argument types to a more generic one
- and make the `hardware_data` an Option.

I'm not entirely convinced this would make it easier to understand what and why it is supposed to happen.

Noi0103:
From the xhci specification 4.11.2.2:
> The xHC shall NOT check for the following Control transfer error conditions.
Note: Some (non-compliant) USB devices use the SETUP Data wLength field as a
custom parameter for non-data control transfers. xHCI implementations should
not tie a non-zero wLength value to the existence of a Data Stage TD in a control
transfer to ensure compatibility with those devices.
• If a Data Stage TD follows a Setup Stage TD, where wLength = ‘0’.
• If a Status Stage TD does not follow a Setup Stage TD, where wLength = ‘0’.
• If a Data Stage TD does not follow a Setup Stage TD, where wLength > ‘0’38.
• If the total size of the Data Stage TD is not equal to wLength.
...

The current implementation very much relies on the wLength field to make the Control In request.

Following this we would probably need to walk through the control request twice like the TD aggregation does for the Normal In endpoint (collecting and parsing all as necessary, doing the hardware request and only then answering each TRB).

src/device/xhci/endpoint_handle.rs
```diff
@@ -105,66 +230,526 @@ impl<RCEH: RealControlEndpointHandle> ControlEndpointHandle<RCEH> {
...
+            ControlTransferData::Out(usb_request) => {
+                trace!("DMA for ControlOut");
+
+                if immediate_data {
+                    // Only event data should follow when immediate data is used here
+                    // but we do not check for that and allow multiple immediate data
+                    // TRB in the data stage TD.
+
+                    usb_request.data.append(
+                        &mut data_pointer.to_le_bytes()[..transfer_length as usize].to_vec(),
```
snue, line 381: The transfer length is guest controlled. We need to make this `min(8, transfer_length)`.

src/device/xhci/endpoint_handle.rs
```diff
@@ -105,66 +230,526 @@ impl<RCEH: RealControlEndpointHandle> ControlEndpointHandle<RCEH> {
...
+                self.transfer_state.state = ControlTransferStage::MaybeDataStageTrb;
+            }
+            ControlTransferData::Out(_) => {
+                unreachable!("internal error: ControlOut SetupTrb have insufficient information to do the Hardware request; a submission state to arrive here should never be used");
+            }
+        }
+        Ok(())
+    }
+
+    fn data_slices<T: TrbDmaInfo>(&mut self, address: u64, trb: &T) {
```
snue, line 334: The name sounds like the function returns slices, but it doesn't return anything and carries out the DMA itself.

```suggestion
    fn transfer_data_slices<T: TrbDmaInfo>(&mut self, address: u64, trb: &T) {
```

src/device/xhci/endpoint_handle.rs
```diff
@@ -705,7 +1303,7 @@ impl<RIEH: RealInEndpointHandle> EndpointHandle for TdBasedInEndpointHandle<RIEH
         let transfer_trb_variant = TransferTrbVariant::parse(trb.buffer);
         let supported_trb_variant =
             match SupportedInEndpointTrbVariant::try_from(transfer_trb_variant) {
-                Ok(supported_trb) => supported_trb,
+                std::result::Result::Ok(supported_trb) => supported_trb,
```
snue, line 1306: Why this specification?

src/device/xhci/endpoint_handle.rs
```diff
@@ -684,6 +1280,8 @@ impl<RIEH: RealInEndpointHandle> TdBasedInEndpointHandle<RIEH> {
             dma_bus,
             event_sender,
             submission_state: TdBasedNormalSubmissionState::default(),
+            // need this to be td transient
```
snue, line 1283: I don't understand this comment? Can you just remove it?

src/device/xhci/endpoint_handle.rs
```diff
@@ -609,17 +1201,19 @@ struct SupportedInEndpointTrb {
...
-    EventData(EventDataTrbData),
+    Normal(NormalTrb),
+    EventData(EventDataTrb),
 }
 
 impl TryFrom<TransferTrbVariant> for SupportedInEndpointTrbVariant {
     type Error = TransferTrbVariant;
 
-    fn try_from(value: TransferTrbVariant) -> Result<Self, Self::Error> {
+    fn try_from(value: TransferTrbVariant) -> core::result::Result<Self, Self::Error> {
```
snue, line 1211: Same here, are the specifications necessary? And why sometimes `core` sometimes `std`?

src/device/xhci/endpoint_handle.rs
```diff
@@ -834,21 +1434,25 @@ impl<'a> TdProcessingInfo<'a> {
...
+            }
+            SupportedInEndpointTrbVariant::EventData(event_data) => {
+                self.handle_event_data_trb(trb.addr, event_data, event_meta)
             }
-            SupportedInEndpointTrbVariant::EventData(_data) => todo!(),
         }
     }
 
-    fn process_normal_trb(
+    // TODO add event_meta here
```
snue, line 1449: Is this TODO still open?

snue (commented):
Leaving some more comments from a partial review. The extensive tests and added test utils/mocks are looking good. That highly improves the trust in the implementation 😉

src/device/xhci/endpoint_handle.rs
Noi0103: In my understanding the unspecified `Ok()` was taken from `anyhow` and needs to use the explicit `std::result::Result::Ok()` here (although i see that `Result::Ok()` is already enough so it can be shorter).
```
mismatched types
expected enum `Result<_, trb::TransferTrbVariant>`
   found enum `Result<_, anyhow::Error>`
```

src/device/xhci/endpoint_handle.rs
Noi0103: Using core was my mistake, probably with the VSCode "fix" button that can add specifications or imports.

src/device/xhci/endpoint_handle.rs
Noi0103: We can't generally make this a minimum of 8 if we want to use immediate data.

> Immediate Data (IDT). If this bit is set to ‘1’, it specifies that the Data Buffer Pointer field of this TRB contains data, not a pointer, and the Length field shall contain a value between ‘0’ and ‘8’ to indicate the number of valid bytes from offset 0 in the TRB that should be used as data.

I don't think I understand why a too small number is an issue when we already parse the byte field as unsigned.

snue: > We can't generally make this a minimum of 8 if we want to use immediate data.

Maximum of 8. The minimum of either 8 or the requested (guest controlled) transfer length. Because the pointer only has 8 bytes.

 > I don't think I understand why a too small number is an issue when we already parse the byte field as unsigned.

Transfer length is parsed as a 17 bit unsigned value. It is between 0 and 131071. I have not checked what happens for an empty range slice (`0..0`). But everything above 8 is broken and will cause a panic. The bounds checks should catch this even in release code. It would be even worse if they didn't.

Noi0103: Ah, yes, a maximum of 8 in case of immediate data makes sense. Thanks.

src/device/xhci/endpoint_handle.rs
```diff
@@ -105,66 +166,415 @@ impl<RCEH: RealControlEndpointHandle> ControlEndpointHandle<RCEH> {
...
+                // From xhci specification chapter 4.11.2.2:
+                //
+                // System software is responsible for ensuring that the total data length defined by a
+                // Data Stage TD (i.e. the sum of the Length fields of the Data Stage TRB and all Normal
+                // TRBs) is equal to wLength. Note that communicating with some non-compliant
+                // devices may require violating this rule.
+                if usb_request.data.len() < transfer_length as usize {
+                    self.transfer_state.state = ControlTransferStage::ExpectStatusStageTrb;
+                    self.submission_state = ControlSubmissionState::ParserError(address); // TODO maybe protocol error?
+                    return;
```
snue, line 364: Please log a warning for this condition.

src/device/xhci/endpoint_handle.rs
```diff
@@ -79,16 +79,77 @@ impl BaseEndpointHandle for DummyEndpointHandle {
...
+}
+
+fn interrupt_on_completion(
+    address: u64,
+    completion_code: CompletionCode,
+    event_data: bool,
+    endpoint_id: u8,
+    slot_id: u8,
+    event_sender: &EventSender,
+) -> anyhow::Result<()> {
```
snue: Why not as a method on `ControlEndpointHandle`? That would save at least three arguments.

src/device/xhci/endpoint_handle.rs
```diff
@@ -863,13 +869,62 @@ impl<RCEH: RealControlEndpointHandle> ControlEndpointHandle<RCEH> {
                 )
             }
             ControlRequestProcessingResult::SuccessfulControlOut => {
-                panic!("SuccessfulControlOut should be handled elsewhere")
+                unreachable!(
+                    "internal error: Don't try processing an error with a successful ControlRequestProcessingResult."
+                )
```
snue, line 886: nit: I think you moved these into a different commit.

src/device/xhci/endpoint_handle.rs
```diff
@@ -974,79 +1078,84 @@ impl<ROEH: RealOutEndpointHandle> EndpointHandle for OutEndpointHandle<ROEH> {
...
+                                0,
+                                CompletionCode::StallError,
+                                false,
+                                self.endpoint_id,
+                                self.slot_id,
+                            );
+                            self.event_sender.send(event)?;
+
+                            // TODO should this be some?
+                            TrbProcessingResult::Stall(None)
```
snue: The `Some` is intended to rewind the dequeue pointer to the correct TRB in case of TD-based endpoint implementations. The Out endpoint is strictly TRB-based for now, so `None` is correct.

src/device/xhci/endpoint_handle.rs
```diff
@@ -1089,8 +1198,8 @@ impl TryFrom<TransferTrbVariant> for SupportedInEndpointTrbVariant {
 
     fn try_from(value: TransferTrbVariant) -> Result<Self, Self::Error> {
         match value {
-            TransferTrbVariant::Normal(data) => Ok(Self::Normal(data)),
-            TransferTrbVariant::EventData(data) => Ok(Self::EventData(data)),
+            TransferTrbVariant::Normal(normal) => Result::Ok(Self::Normal(normal)),
+            TransferTrbVariant::EventData(event_data) => Result::Ok(Self::EventData(event_data)),
```
snue: The specification is needed because of the added `use anyhow::Ok` above. Not sure where that is used. It may be better to use a specification for `anyhow` in-place there instead.

src/device/xhci/endpoint_handle.rs
```diff
@@ -1239,6 +1352,7 @@ impl<RIEH: RealInEndpointHandle> EndpointHandle for TdBasedInEndpointHandle<RIEH
     }
 }
 
+#[allow(clippy::too_many_arguments)]
```
snue, line 1359: This is unfortunate, but I don't have a suggestion for an easy refactor.

src/device/xhci/endpoint_handle.rs
```diff
@@ -1280,7 +1395,6 @@ fn process_real_transfer_response(
 
     Ok(TrbProcessingResult::Ok)
 }
-
```
snue, line 812: I think this line should stay? Does the formatter not complain?

src/device/xhci/endpoint_handle.rs
```diff
@@ -1305,21 +1419,25 @@ impl<'a> TdProcessingInfo<'a> {
...
+            }
+            SupportedInEndpointTrbVariant::EventData(event_data) => {
+                self.handle_event_data_trb(trb.addr, event_data, event_meta)
             }
-            SupportedInEndpointTrbVariant::EventData(_data) => todo!(),
         }
     }
 
-    fn process_normal_trb(
+    // TODO add event_meta here
```
snue: Comment resolved, but still a stale TODO?

src/device/xhci/trb.rs
snue: Can we make this `expect(unused)` or remove it while here?

src/device/xhci/endpoint_handle.rs
```diff
@@ -105,66 +236,531 @@ impl<RCEH: RealControlEndpointHandle> ControlEndpointHandle<RCEH> {
...
+            self.endpoint_id,
+            self.slot_id,
+        );
+
+        self.event_sender.send(event)?;
+        self.transfer_state.event_meta.zero();
+
+        // According to the spec Event Data shall always have the IOC bit set.
+        // We handle the IOC bit with the same generic function as we do on TransferTrb's.
+        if trb.interrupt_on_completion {
```
snue: nit: You have the old comment on the Normal endpoint still. I think you can combine both to explain we actually generate two events here.

src/device/xhci/endpoint_handle.rs
```diff
@@ -301,104 +887,65 @@ impl<RCEH: RealControlEndpointHandle> ControlEndpointHandle<RCEH> {
...
+    trace!("EventData TRB on Normal Ep");
+
+    let event = EventTrb::new_transfer_event_trb(
+        event_data_trb.event_data,
+        event_meta.edtla,
+        event_meta.previous_completion_code,
+        true,
+        endpoint_id,
+        slot_id,
+    );
```
snue: To be spec compliant, for the Short Packet or Error case, this must check the PAE bit before generating an event.

snue (commented):
Left some more comments. I think the amount of duplication is fine. The use of shared functions and then passing a lot of the same `self.something` over and over may be avoided/reduced by providing another slim wrapper on `self` for these calls.

src/device/xhci/endpoint_handle.rs
Noi0103: Yeah, this might have been a rebase mistake. Clippy did not complain.

src/device/xhci/endpoint_handle.rs
Noi0103: The currently easiest/fewest lines of code solution should be to set this bit in HCCPARAMS1 and handle every Event Data TRB compared to tracking if this is the first after a short transfer or similar.

src/device/xhci/trb.rs
```diff
@@ -857,21 +848,28 @@ impl TransferTrbVariant {
     }
 }
 
+pub trait TrbDmaInfo {
+    fn data_pointer(&self) -> u64;
+    fn transfer_length(&self) -> u32;
+    fn has_immediate_data(&self) -> bool;
+}
```
snue, line 856: This could use a doc comment.

src/device/xhci/trb.rs
```diff
@@ -962,17 +973,18 @@ impl TrbData for SetupStageTrbData {
...
     pub interrupt_on_completion: bool,
     pub immediate_data: bool,
     pub direction: bool,
 }
 
-impl TrbData for DataStageTrbData {
+impl TrbData for DataStageTrb {
     /// Parse data of a Data Stage TRB.
     ///
     /// Only `TransferTrb::try_from` should call this function.
```
snue: This needs an update with `TransferTrb` gone. I think it's just `TransferTrbVariant::` now. Please check for other comments that may need adjustment.

src/device/pci/constants.rs
```diff
@@ -291,12 +291,13 @@ pub mod xhci {
...
     /// Constants for the capability register.
     pub mod capability {
+
         /// We only emulate version 1.0.0 of the XHCI spec for simplicity.
         pub const HCIVERSION: u64 = 0x100;
         pub const HCSPARAMS1: u64 =
             (super::MAX_PORTS << 24) | (super::MAX_INTRS << 8) | super::MAX_SLOTS;
         pub const HCSPARAMS2: u64 = super::MAX_ERST_SIZE_EXP << 4;
-        pub const HCCPARAMS1: u64 = super::offset::SUPPORTED_PROTOCOLS << 14;
+        pub const HCCPARAMS1: u64 = hccparams1::PAE | super::offset::SUPPORTED_PROTOCOLS << 14;
```
snue, line 300: We should add a doc comment here explaining our implementation parses all Event Data TRBs in a TD.

src/device/xhci/endpoint_handle.rs
snue: The comment gives away that we probably didn't put enough thought into this 😉 

I think we should not generate the additional event here. The Spec says IOC is pretty straightforward "If IOC is set, generate an Event". EventData and regular IOC use are mutually exclusive when used correctly. And I don't see a compelling reason to generate more than one event and interrupt here.

> Note:The IOC flag is treated generically by the xHC. If it is set in a TRB, then the xHC
shall generate an Event for that TRB. If the IOC flag is not set in an Event Data
TRB, the xHC will advance past it, clearing the EDTLA in the process.

(section 4.11.5.2)

So even though the Spec mandates IOC for EventData, the controller will still only generate an event when IOC is set.

Can you switch to only generating the ED Event if IOC is set and check if Windows struggles when we only generate that one Event?

src/device/xhci/endpoint_handle.rs
```diff
@@ -105,66 +235,537 @@ impl<RCEH: RealControlEndpointHandle> ControlEndpointHandle<RCEH> {
...
+                Ok(())
+            }
         }
     }
+
+    fn handle_event_data_trb(&mut self, address: u64, trb: EventDataTrb) -> anyhow::Result<()> {
+        trace!("EventData TRB");
+
+        // Parse every event data after non success completion including all
+        // consecutive ones after after that since we set the HCCPARAMS1 PAE bit.
```
snue: We can also remove this comment here, as we have no special handling. Rather explain for the constant that `PAE=1` simplifies the logic by letting us handle all event data TRBs the same.

There is also a doubled "after" in the current version:
```suggestion
        // Parse every event data after non success completion including all
        // consecutive ones after that since we set the HCCPARAMS1 PAE bit.
```

src/device/xhci/endpoint_handle.rs
```diff
@@ -429,16 +987,76 @@ impl<ROEH: RealOutEndpointHandle> OutEndpointHandle<ROEH> {
...
+        if trb.immediate_data {
+            todo!("immediate data in a Normal Trb on a Normal Out Endpoint")
+        }
+
+        if !trb.chain {
+            self.event_meta = EventDataTrbMetadata::default();
+        }
+
+        let mut data = vec![0; trb.transfer_length as usize];
+        self.dma_bus.read_bulk(trb.data_pointer, &mut data);
```
snue: Any reason to leave this a `todo`? Doesn't this undo #285 and is required for serial support?

Noi0103: Thanks for catching, this is a mistake that must have slipped the cracks with a rebase.

PR #243, opened 2026-06-03
Improve testing data integrity on blockdevice tests

Noi0103:
The current data that is tested for corruption on transfer is too small to reliably test for changes. Instead using 32MB, being double the maximum edtla value (see xhci spec) seems sufficient. This should help us to catch data corruption on successful looking transfers.

### DIFF
--- a/nix/tests.nix
+++ b/nix/tests.nix
@@ -228,7 +228,7 @@ let
   # Some static values for ...
   # ... creating blockdevice backing files.
   imagePathPart = "/tmp/image";
-  imageSize = "8M";
+  imageSize = "48M";
   # ... identifying QEMU's virtual Devices.
   blockdeviceVendorId = "46f4";
   blockdeviceProductId = "0001";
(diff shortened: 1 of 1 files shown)


nix/tests.nix
```diff
@@ -631,14 +631,28 @@ let
...
+
+    # Create a file and compute a checksum
+    cloud_hypervisor.succeed("dd if=/dev/urandom of=/tmp/file count=32000 bs=1024", timeout=60)
+    out = cloud_hypervisor.succeed("sha256sum /tmp/file", timeout=60)
+    hash_tmp = out.split()
+    print(f"hash_tmp: {hash_tmp}")
+
+    # Copy the file on the blockdevice
+    cloud_hypervisor.succeed("cp /tmp/file /mnt/file", timeout=60)
+    out = cloud_hypervisor.succeed("sha256sum /mnt/file", timeout=60)
```
snue: There is a real possibility that the file read is fulfilled from the linux page cache and not actually read back from the device. Either drop the vm caches before computing the sha sum from `/mnt`, or only check the sha sum after the umount/mount cycle, or both. See `drop_caches` in https://www.kernel.org/doc/Documentation/sysctl/vm.txt

nix/tests.nix
```diff
@@ -631,14 +631,28 @@ let
...
     cloud_hypervisor.wait_until_succeeds("lsblk /dev/sda1", timeout=60)
 
-    # Test filesystem
+    # Make a filesystem
     cloud_hypervisor.succeed("mkfs.ext4 /dev/sda1", timeout=60)
     cloud_hypervisor.wait_until_succeeds("mount /dev/sda1 /mnt", timeout=60)
-    cloud_hypervisor.succeed("echo 123TEST123 > /mnt/file.txt", timeout=60)
+
+    # Create a file and compute a checksum
+    cloud_hypervisor.succeed("dd if=/dev/urandom of=/tmp/file count=32000 bs=1024", timeout=60)
```
snue: nit: you can also encode the size as:

```suggestion
    cloud_hypervisor.succeed("dd if=/dev/urandom of=/tmp/file count=32 bs=1M", timeout=60)
```
or
```suggestion
    cloud_hypervisor.succeed("dd if=/dev/urandom of=/tmp/file count=32768 bs=1024", timeout=60)
```

Noting that the current size is only `31.2MiB`:

```
 % 32000 * 1024
int32   32768000
uint32  32768000
hex     0x1f40000
octal   0175000000
unit    31.2M

% 32 * 1024 * 1024
int32   33554432
uint32  33554432
hex     0x2000000
octal   0200000000
unit    32M
```

snue (commented):
Thanks. We'll need to tweak it a little to guarantee we read from the device.

nix/tests.nix
Noi0103: Thanks. Upon revisiting I'm questioning how I did not catch the remount being too late.
I will add a sync and cache drop as per the docs.

snue:
The test pattern for the multiple block devices test needs fixing:
```
> !!! RequestedAssertionFailed: pattern `sda\s+\d+:\d+\s+0\s+8M\s+0\s+disk` not found in NAME  MAJ:MIN RM SIZE RO TYPE MOUNTPOINTS
```

Noi0103:
Already included in the fixup :D

snue (approved):
And it still passes the CI. That's a good sign 😁

PR #244, opened 2026-06-04
Centralize port ID offsetting in MMIO helpers

ziyifu225:
This updates the xHCI port MMIO helpers to return one-based port IDs instead of zero-based port indices.
It follows up on the discussion in [previous PR](https://github.com/cyberus-technology/usbvfiod/pull/236#discussion_r3189701323).

`PortArray` uses `OneIndexed` storage and its `PORTSC` / `PORTPMSC` accessors already expect xHCI port IDs. The MMIO path was previously converting the helper result with `port_index + 1` at each call site. This change moves that conversion into the address helper itself and renames the public helpers from `*_index` to `*_id` to match the actual semantics.

### DIFF
--- a/src/device/pci/xhci.rs
+++ b/src/device/pci/xhci.rs
@@ -16,7 +16,7 @@ use crate::device::{
     },
     xhci::{
         endpoint_launcher::EndpointLauncher,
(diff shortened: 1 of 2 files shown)


snue (approved):
Thanks!

PR #246, opened 2026-06-05
Batch processing of TDs on the IN endpoints

lbeierlieb:
The current IN-endpoint-handle strategy is individual processing of each TRB. As far as we observed, this strategy worked fine on Linux, but on Windows, we regularly observe TDs in that the first TRB is not a multiple of the max_packet_size. We then request the next higher multiple of max_packet_size, receive more data than the driver requested and drop the rest--which we later wait for.

This draft PR experiments with an alternate strategy, where we first collect all TRBs that belong to the same TD, sum up the transfer lengths and request all the data in one big nusb request. Afterwards, we distribute the device response to the DMA buffers of the individual TRBs. This PR so far only implements the happy path, any non-successful transfers will cause a panic (the disconnect test in the CI must have been only passed by chance; the disconnects happened when there was no active transfer).

Rough performance tests with USB-3 stick (`sudo dd if=/dev/sda of=/dev/null bs=4M status=progress):
- on host, without passthrough (for reference): 115 MB/sec
- TRB-based processing (current main branch): 82 MB/sec
- TD-based processing (this PR): 74 MB/sec

I have no explanation why grouping many smaller requests into few, larger requests causes a slow down, I would have expected the opposite.


snue:
> I have no explanation why grouping many smaller requests into few, larger requests causes a slow down, I would have expected the opposite.

You were right on the expectation. My own experiments show a slight performance improvement to `main` with the proposed batch processing. In any case, we need to fix the current broken approach, and I favor continuing with this draft.

lbeierlieb:
Performance measurements with the full implementation:

- on Host (no passthrough): 250 MB/sec
- current main (): 102 MB/sec
- TD-aggregation (this PR): 113 MB/sec

Average read throughput of `sudo dd if=/dev/sda of=/dev/null bs=4M status=progress` after 20 sec; release build of usbvfiod, no debug prints; 8 vcpus for the CHV VM. Patriot Rage R550 stick.

snue (approved):
I like it. I don't even have nitpicks.

PR #248, opened 2026-06-08
Async normal in endpoint reader as background task

Noi0103:
See also #246 

This Draft aims to prototype using the higher level API the nusb crate provides. This is only done for bulk and interrupt In Endpoints.

In it's current state error handling is _very_ limited. This results in panics on device disconnect/stall/... .


Noi0103:
closing in favor of #246

PR #249, opened 2026-06-08
Add fsck to blockdevice tests

Noi0103:
We have observed the blockdevice tests to be flaky due to failing `mount` calls. At first USB 1.1 later USB 2 and one time on my local machine even USB 3. The error message from `mount` was not really helpful enough to identify the source of this flakiness.

As a first attempt to limit the list of causes, we decided to add a filesystem check before calling mount.

The flags `-r` and `-V` and printing all output seemed like a nice to have.

### DIFF
--- a/nix/tests.nix
+++ b/nix/tests.nix
@@ -632,7 +632,9 @@ let
     cloud_hypervisor.wait_until_succeeds("lsblk /dev/sda1", timeout=60)
 
     # Make a filesystem
-    cloud_hypervisor.succeed("mkfs.ext4 /dev/sda1", timeout=60)
+    out = cloud_hypervisor.succeed("mkfs.ext4 -v /dev/sda1", timeout=60)
+    print(out)
+    cloud_hypervisor.succeed("fsck -t ext4 -V -r /dev/sda1 -- -y", timeout=60)
     cloud_hypervisor.wait_until_succeeds("mount /dev/sda1 /mnt", timeout=60)
 
     # Create a file and compute a checksum


nix/tests.nix
```diff
@@ -633,6 +633,8 @@ let
 
     # Make a filesystem
     cloud_hypervisor.succeed("mkfs.ext4 /dev/sda1", timeout=60)
+    out = cloud_hypervisor.succeed("fsck -r -V -y /dev/sda1", timeout=60)
```
snue: The filesystem is fresh and should not need any repair. We just want to verify it wasn't corrupted on write. The `-r` and `-V` is probably fine (depending on how verbose that is), but let's go with `-n` and not auto-repair, just complain.

Noi0103: Fine by me, the exit code shouldn't be 0 in both cases when a corrupt filesystem is encountered.

nix/tests.nix
```diff
@@ -633,6 +633,8 @@ let
 
     # Make a filesystem
     cloud_hypervisor.succeed("mkfs.ext4 /dev/sda1", timeout=60)
```
snue: Should we also print the output of `mkfs` for completeness?

snue (commented):
That should give us a good indication if the new filesystem is actually usable and detected correctly. Left one small suggestion, but otherwise this is fine and I'm looking forward to the mount failing again 😁

nix/tests.nix
```diff
@@ -633,6 +633,8 @@ let
...
+    out = cloud_hypervisor.succeed("fsck -t ext4 -N -V -r /dev/sda1", timeout=60)
```
snue: Ah, not quite ... this is a dry-run
```
     -N
         Don’t execute, just show what would be done.
```

Noi0103: So this is not meant as "Look for mistakes but do not do anything with them" but rather is "show how you would look for mistakes without doing so"...ah, CI logs show that too:
```
fsck from util-linux 2.42
[/run/current-system/sw/bin/fsck.ext4 (1) -- /dev/sda1] fsck.ext4 /dev/sda1 
```

snue: You want to pass additional options for the filesystem specific tool instead. I think this should work:

```
fsck -V -r -t ext4 /dev/sda1 -- -n
```

nix/tests.nix
Noi0103: I doubt it will be of much use since we have the successful exit code, but it won't flood the logs so I will add them just in case.

nix/tests.nix
```diff
@@ -632,7 +632,10 @@ let
     cloud_hypervisor.wait_until_succeeds("lsblk /dev/sda1", timeout=60)
 
     # Make a filesystem
-    cloud_hypervisor.succeed("mkfs.ext4 /dev/sda1", timeout=60)
+    out = cloud_hypervisor.succeed("mkfs.ext4 -v /dev/sda1", timeout=60)
+    print(out)
+    out = cloud_hypervisor.succeed("fsck -t ext4 -y -V -r /dev/sda1", timeout=60)
```
snue: `fsck` is quite forgiving here. It doesn't know about `-y` and passes it on silently. The man page says it should be invoked as `fsck ... /dev/sda1 -- -y` instead. It can stay like this as it works correctly and there are no warnings in the log.

It doesn't matter if we use `-y` or `-n`, as the return code is always non-zero when errors were encountered. That's what we want.

snue (approved):

nix/tests.nix
snue: Ah wait, we probably don't need the output then, as we only get that for the clean fs case. Output of failing commands should be printed by the helper.

snue (approved):

PR #250, opened 2026-06-11
docs: fix invocation in quick-start

snue:
The basic invocation broke when we added a second binary (remote). Be explicit about the invoked tool.

Otherwise, we run into:
```
cargo run -- --socket-path /tmp/usbvfiod.sock --hotplug-socket-path /tmp/hotplug.sock
error: `cargo run` could not determine which binary to run. Use the `--bin` option to specify a binary, or the `default-run` manifest key.
available binaries: remote, usbvfiod
```

### DIFF
--- a/docs/developers/quick-start.md
+++ b/docs/developers/quick-start.md
@@ -8,7 +8,7 @@ An easy way to get a testing setup is to connect `usbvfiod` with Cloud
 Hypervisor. For this, start `usbvfiod` in one terminal:
 
 ```console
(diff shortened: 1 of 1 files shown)


snue:
> or the `default-run` manifest key.

Any opinions if we rather want that to mark the main executable? Using `cargo run` also implies debug builds by default.

lbeierlieb (approved):
I'm fine with both, no clear preference.

PR #252, opened 2026-06-15
Convert PORTSC test into a tokio::test

Noi0103:
Simple fix for an existing `// TODO` by using `#[tokio::test]` to get a more private runtime instead of manually creating a global one.


snue (approved):
Nice!

PR #253, opened 2026-06-16
Recognize when encountering a set immediate data bit

Noi0103:
The immediate data bit is used when the 8 byte pointer field is not a pointer and instead should be used as data field directly. It is not guaranteed to use the whole 8 bytes available.

|Trb|we...|this pr does...|
|---|---|---|
|Normal Trb|should handle it|a todo!()|
|Setup Stage Trb|can ignore it because it is always set to 1|nothing|
|Data Stage Trb|should handle it|a todo!()|
|Isoch Trb|not yet have isochronous transfers|nothing|


snue (approved):
Looks good. That at least prevents us from doing the wrong thing with the data for now.

PR #256, opened 2026-06-24
Print logs if nested wait_until_succeeds() hits a timeout (or other exception)

Noi0103:
Whenever the CI hits a timeout we currently do not get the guest logs (nested cloud hypervisor journal).
With this, after the retry raises either a timeout or any other exception the latest stored command output and the guest journal logs will be added to the build logs of the test.

A timeout example I used for testing this:
```
singleBlockDeviceTestScript=''
    cloud_hypervisor.wait_until_succeeds("echo hello && sleep 20", timeout=10)
    ...
    <original test script for the basic blockdevice test>
    ...
'';
```

### DIFF
--- a/nix/tests.nix
+++ b/nix/tests.nix
@@ -170,6 +170,7 @@ let
             with vm_host.nested(f"must succeed in cloud-hypervisor: {command}"):
                 (status, out) = vm_host.execute("ssh -q -o UserKnownHostsFile=/dev/null -o StrictHostKeyChecking=no root@192.168.100.2 '" + command + "'", timeout=timeout)
                 if status != 0:
+                    print('\nnested.succeed() FAILED')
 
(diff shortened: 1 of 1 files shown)


nix/tests.nix
snue, line 191: In order to see the output from the unsuccessful tries as well, should this better follow the pattern of `succeed` above (with the `output += out` appending)? Or should we only return the output of successful execution to not confuse test scripts? I'm unsure what's better. Some retries are usually expected when this method is used, and the failure messages aren't too exciting in those cases.

snue (approved):
Looks correct. Let's see what this catches in CI.

nix/tests.nix
Noi0103: Discussed this offline without a finite decision.

Noi0103:
Small addition with a searchable `FAILED` above the additional Logs for a failed command.

PR #261, opened 2026-07-01
Clean up the hotplug socket on exit after client disconnect

Noi0103:
This change will affect the happy case where only the hotplug socket remains while the vfio-user socket is removed. The vfio-user socket is cleaned on Drop when the vfio-user server will shutdown due to the client disconnecting.

The not-so-clean case when no client disconnected prior to service exit, both sockets will remain just like before.

### DIFF
--- a/src/main.rs
+++ b/src/main.rs
@@ -64,7 +64,7 @@ fn main() -> Result<()> {
     };
 
     // listen on socket for hot-attach fds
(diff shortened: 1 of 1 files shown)


snue (approved):
Thanks. This makes it easier to understand the socket behavior. All paths are owned by the server and are also cleaned up. We will likely add cleaning up for the involuntary exit case in a follow-up.

Also, we still have to enable sockets passed by open fd to implement externally managed sockets.

PR #263, opened 2026-07-06
Add cli option to disable ansi color code usage

Noi0103:
When redirecting output into a file with `&>` the color codes are noise making logs less nice readable.

To avoid going down an auto detect rabbit hole this is a simple option to deactivate color codes.


snue (approved):
Nice. Thanks.

PR #264, opened 2026-07-07
github-pages: switch markdown processor

snue:
> [!NOTE]
> sections like this do not render on the deployed pages currently

> [!TIP]
> see https://cyberus-technology.github.io/usbvfiod/docs/users/basic.html

Add a Jekyll configuration to select the markdown processor. GFM is used throughout GitHub, so this should fix it.

This is documented here: https://docs.github.com/en/pages/setting-up-a-github-pages-site-with-jekyll/setting-a-markdown-processor-for-your-github-pages-site-using-jekyll

### DIFF
--- a/_config.yml
+++ b/_config.yml
@@ -1,0 +1,5 @@
+# Github Pages / Jekyll configuration
+
+# We use some Github Flavored Markdown features in the documentation.
+markdown: GFM
+


Noi0103 (approved):

snue:
Well, it's still broken the same way.

Noi0103:
Looks like we are not the only ones who followed that docs page and it does not work as we expected.
https://github.com/mozilla/policy-templates/blob/master/_config.yml
https://mozilla.github.io/policy-templates/

PR #265, opened 2026-07-08
Switch llvm coverage report output to html format

Noi0103:
We are not using the lcov output of this derivation (https://crane.dev/API.html#cranelibcargollvmcov).

With the html format we have a human readable format for local use.


snue (approved):
Yes, let's try that for now. So far, we don't look at coverage data at all. Hopefully we find a nice way to integrate that in reviews eventually.

PR #266, opened 2026-07-08
Update flake lock

Noi0103:
There have been no renovate lock file update PR's lately, so this is a manual one to fill in.

This includes fixes for the deprecation of `os.system()` in favor of `subprocess.run()` used in the integration test scripts.

### DIFF
--- a/flake.lock
+++ b/flake.lock
@@ -3,11 +3,11 @@
     "advisory-db": {
       "flake": false,
       "locked": {
-        "lastModified": 1781339692,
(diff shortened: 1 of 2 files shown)


snue (commented):
Oh, I was looking into this as well today. We don't need the explicit monitor socket access. We can instead just use the NixOS test framework wrappers `machine.send_monitor_command` and `machine.send_key`.

Noi0103:
closing in favor of #267

PR #267, opened 2026-07-08
Lock file maintenance with some test script simplification

snue:
Update the Cargo and Flake dependencies.

* Remove deprecated `os.system` usage
* Simplify QMP commands with the built-in monitor commands
* Adapt loop iterations on attach/detach to end on iteration 16 (which checks the device shows up in the guest)

The cargo update (which isn't done by renovate for us :thinking:) cleaned up the lock file quite a bit.

This supersedes #266.

### DIFF
--- a/nix/tests.nix
+++ b/nix/tests.nix
@@ -310,10 +310,10 @@ let
       filepath = "${imagePathPart}-${testname}-${device.udevRule.symlink}.img";
     in
     ''
-      import os
-      os.system("rm ${filepath}")
+      import subprocess
+      subprocess.run(["rm", "${filepath}"])
       print("Creating file image at ${filepath}")
-      os.system("dd bs=1  count=1 seek=${imageSize} if=/dev/zero of=${filepath}")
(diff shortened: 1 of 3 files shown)


nix/tests.nix
```diff
@@ -873,11 +863,8 @@ blockdeviceTests
         search("No attached devices", out)
 
         # plug in a blockdevice
-        os.system(textwrap.dedent("""
-        ${pkgs.socat}/bin/socat - UNIX-CONNECT:/tmp/qmp.sock >> /dev/null <<EOF
-        {"execute": "qmp_capabilities"}
-        {"execute": "device_add", "arguments": {"driver": "usb-storage", "id": "hotplug", "bus": "${usbVersions."3".busName}.0", "drive": "hotplug", "port": "1"} }
-        EOF"""))
+        print(machine.send_monitor_command("drive_add 0 id=hotplug,if=none,file=/tmp/hotplug.img,format=raw"))
```
snue, line 866: Note: using this interface the drive is removed when the usb device is removed. We need to call this in every loop iteration now.

Noi0103 (approved):

PR #268, opened 2026-07-09
Controller hardware reset

ziyifu225:
When we tried to unload and reload the `xhci_pci` module in Linux, the warning `WARN usbvfiod::device::xhci::command_ring: received useless write to CRCR while running` can be triggered. 
The reason is that we didn't really implement host controller reset before. When the Linux guest unloaded and reloaded `xhci_pci`, the driver reset the host controller and then reprogrammed the command/event/slot state. Since usbvfiod still kept the old state, the command ring could still be considered running and CRCR writes during re-initialization produced warnings.

This PR fixes this error by
- Stopping the command ring
- Resetting event ring/interrupter state
- Resetting slot manager state

### DIFF
--- a/nix/checks/controller-reset.nix
+++ b/nix/checks/controller-reset.nix
@@ -1,0 +1,92 @@
+{
+  lib,
+  cloud-hypervisor,
+  usbvfiod,
+  testutils,
+}:
+let
+  systemd-config = args: {
+    systemd.services = {
+      usbvfiod = {
+        wantedBy = [ "multi-user.target" ];
+        serviceConfig = {
+          User = "usbaccess";
+          Group = "usbaccess";
+          Restart = "on-failure";
+          RestartSec = "2s";
+          ExecStart = ''
+            ${lib.getExe usbvfiod} ${
+              if args.debug then "-v" else ""
+            } --socket-path ${testutils.usbvfiodSocket} --hotplug-socket-path ${testutils.usbvfiodSocketHotplug} ${lib.concatStringsSep " " (builtins.map testutils.mkDeviceFlag args.virtualDevices)}
+          '';
+        };
+        environment = {
+          RUST_BACKTRACE = "full";
+        };
+      };
+
+      cloud-hypervisor =
+        let
+          netboot = testutils.mkNetboot args.debug;
+        in
+        {
+          wantedBy = [ "multi-user.target" ];
+          requires = [ "usbvfiod.service" ];
+          after = [ "usbvfiod.service" ];
+          serviceConfig = {
+            Restart = "on-failure";
+            RestartSec = "2s";
+            ExecStart = ''
+              ${lib.getExe cloud-hypervisor} --memory size=2G,shared=on --console file=${testutils.guestLogFile} --serial off \
+                --kernel ${netboot.kernel} \
+                --cmdline ${lib.escapeShellArg netboot.cmdline} \
+                --initramfs ${netboot.initrd} \
+                --user-device socket=${testutils.usbvfiodSocket} \
+                --net "tap=tap0,mac=,ip=192.168.100.1,mask=255.255.255.0"
+            '';
+          };
+        };
+    };
+  };
+
+  testScript = ''
+    # Wait until the USB drive is recognized.
+    out = cloud_hypervisor.wait_until_succeeds("lsusb -d ${testutils.blockdeviceVendorId}:${testutils.blockdeviceProductId}", timeout=120)
+    search("ID ${testutils.blockdeviceVendorId}:${testutils.blockdeviceProductId} QEMU QEMU USB HARDDRIVE", out)
+    cloud_hypervisor.wait_until_succeeds("lsblk /dev/sda", timeout=120)
+
+    # Run the controller reset loop a few times.
+    for i in range(1, 4):
+      print(f"CONTROLLER RESET LOOP {i}")
+
+      # Reload the guest xhci_pci driver to trigger a host controller reset.
+      cloud_hypervisor.succeed("modprobe -r xhci_pci", timeout=120)
+      cloud_hypervisor.succeed("modprobe xhci_pci", timeout=120)
+
+      # Confirm raw block I/O still works after the controller reset.
+      out = cloud_hypervisor.wait_until_succeeds("lsusb -d ${testutils.blockdeviceVendorId}:${testutils.blockdeviceProductId}", timeout=120)
+      search("ID ${testutils.blockdeviceVendorId}:${testutils.blockdeviceProductId} QEMU QEMU USB HARDDRIVE", out)
+      cloud_hypervisor.wait_until_succeeds("lsblk /dev/sda", timeout=120)
+      cloud_hypervisor.succeed(f"printf after-reset-{i} > /tmp/after-reset-{i}.txt", timeout=60)
+      cloud_hypervisor.succeed(f"dd if=/tmp/after-reset-{i}.txt of=/dev/sda bs=512 seek=2048 count=1 conv=sync,fsync status=none", timeout=60)
+      cloud_hypervisor.succeed("sync", timeout=60)
+      cloud_hypervisor.succeed("echo 3 > /proc/sys/vm/drop_caches", timeout=60)
+      cloud_hypervisor.succeed(f"dd if=/dev/sda of=/tmp/read-after-reset-{i}.txt bs=512 skip=2048 count=1 status=none", timeout=60)
+      cloud_hypervisor.succeed(f"grep -ao after-reset-{i} /tmp/read-after-reset-{i}.txt", timeout=60)
+  '';
+in
+builtins.listToAttrs (
+  builtins.map (usbVersion: {
+    name = "controller-reset-usb-${builtins.replaceStrings [ "." ] [ "_" ] usbVersion}";
+    value = testutils.mkUsbTest {
+      name = "controller-reset-usb-${usbVersion}";
+      virtualDevices = [
+        {
+          type = "blockdevice";
+          inherit usbVersion;
+        }
+      ];
+      inherit testScript;
+    } systemd-config;
+  }) (builtins.attrNames testutils.usbVersions)
+)
--- a/src/device/pci/xhci.rs
+++ b/src/device/pci/xhci.rs
@@ -5,13 +5,14 @@
 use std::sync::{Arc, Mutex};
 
 use tokio::runtime;
+use tracing::{error, info};
 
 use crate::device::{
     bus::{BusDeviceRef, SingleThreadedBusDevice},
     interrupt_line::InterruptLine,
     pci::{
         config_space::{ConfigSpace, ConfigSpaceBuilder},
-        constants::xhci::{offset, MAX_INTRS, RUN_BASE},
+        constants::xhci::{offset, operational::usbcmd, MAX_INTRS, RUN_BASE},
         traits::PciDevice,
     },
     xhci::{
@@ -114,7 +115,24 @@ impl<CRD: CompleteRealDevice> PciDevice for XhciController<CRD> {
 
         match req.addr {
             // xHC Operational Registers
-            offset::USBCMD => self.usbcmd.write(value),
+            offset::USBCMD => {
+                self.usbcmd.write(value);
+                if value & usbcmd::HCRST == usbcmd::HCRST {
+                    info!("Host Controller Reset requested");
+                    self.command_ring
+                        .reset()
+                        .map_err(|err| error!("failed to reset command ring: {err}"))
+                        .ok();
+                    self.interrupter
+                        .reset()
+                        .map_err(|err| error!("failed to reset event ring: {err}"))
+                        .ok();
+                    self.slot_manager
+                        .reset()
+                        .map_err(|err| error!("failed to reset slots: {err}"))
+                        .ok();
+                }
+            }
             offset::DNCTL => assert_eq!(value, 2, "debug notifications not supported"),
             offset::CRCR => self
                 .command_ring
--- a/src/device/pci/xhci.rs
+++ b/src/device/pci/xhci.rs
@@ -5,17 +5,17 @@
 use std::sync::{Arc, Mutex};
 
 use tokio::runtime;
-use tracing::{error, info};
 
 use crate::device::{
     bus::{BusDeviceRef, SingleThreadedBusDevice},
     interrupt_line::InterruptLine,
     pci::{
         config_space::{ConfigSpace, ConfigSpaceBuilder},
-        constants::xhci::{offset, operational::usbcmd, MAX_INTRS, RUN_BASE},
+        constants::xhci::{offset, MAX_INTRS, RUN_BASE},
         traits::PciDevice,
     },
     xhci::{
+        controller_reset::ResetCoordinator,
         endpoint_launcher::EndpointLauncher,
         port::{get_portli_id, get_portpmsc_id, get_portsc_id, HotplugControl, PortArray},
         real_device::CompleteRealDevice,
@@ -60,6 +60,15 @@ impl<CRD: CompleteRealDevice> XhciController<CRD> {
             interrupter.create_event_sender(),
             slot_manager.create_slot_worker_handle(),
             usbcmd.value_reference(),
+        );
+        ResetCoordinator::start(
+            usbcmd.clone(),
+            [
+                Box::new(command_ring.reset_sender()),
+                Box::new(interrupter.reset_sender()),
+                Box::new(slot_manager.reset_sender()),
+            ],
+            &async_runtime,
         );
 
         Self {
@@ -115,24 +124,7 @@ impl<CRD: CompleteRealDevice> PciDevice for XhciController<CRD> {
 
         match req.addr {
             // xHC Operational Registers
-            offset::USBCMD => {
-                self.usbcmd.write(value);
-                if value & usbcmd::HCRST == usbcmd::HCRST {
-                    info!("Host Controller Reset requested");
-                    self.command_ring
-                        .reset()
-                        .map_err(|err| error!("failed to reset command ring: {err}"))
-                        .ok();
-                    self.interrupter
-                        .reset()
-                        .map_err(|err| error!("failed to reset event ring: {err}"))
-                        .ok();
-                    self.slot_manager
-                        .reset()
-                        .map_err(|err| error!("failed to reset slots: {err}"))
-                        .ok();
-                }
-            }
+            offset::USBCMD => self.usbcmd.write(value),
             offset::DNCTL => assert_eq!(value, 2, "debug notifications not supported"),
             offset::CRCR => self
                 .command_ring
--- a/src/device/xhci/command_ring.rs
+++ b/src/device/xhci/command_ring.rs
@@ -54,6 +54,7 @@ enum WorkerMessage {
     SetDequeuePointerAndCS(u64, bool),
     Doorbell,
     Stop,
+    Reset,
 }
 
 impl CommandRing {
@@ -144,6 +145,10 @@ impl CommandRing {
         }
     }
 
+    pub fn reset(&self) -> anyhow::Result<()> {
+        self.send_to_worker(WorkerMessage::Reset)
+    }
+
     fn send_to_worker(&self, msg: WorkerMessage) -> anyhow::Result<()> {
         self.sender_to_worker.send(msg)?;
 
@@ -166,6 +171,9 @@ impl CommandWorker {
         loop {
             match &self.state {
                 WorkerState::Stopped => match self.next_msg().await? {
+                    WorkerMessage::Reset => {
+                        self.commandring_running.store(false, Ordering::Relaxed);
+                    }
                     WorkerMessage::SetDequeuePointerAndCS(dp, cs) => {
                         debug!("Updating command ring parameters: dp={dp:#x}, cs={cs}");
                         self.ring.set_dequeue_pointer(dp, cs);
@@ -183,6 +191,10 @@ impl CommandWorker {
                     msg => warn!("Unexpected message: msg={msg:?}, state={:?}", self.state),
                 },
                 WorkerState::Idle => match self.next_msg().await? {
+                    WorkerMessage::Reset => {
+                        self.commandring_running.store(false, Ordering::Relaxed);
+                        self.state = WorkerState::Stopped;
+                    }
                     WorkerMessage::Doorbell => {
                         self.state = WorkerState::LookingForNewCommand;
                     }
@@ -198,6 +210,11 @@ impl CommandWorker {
                         };
 
                         match msg {
+                            WorkerMessage::Reset => {
+                                self.commandring_running.store(false, Ordering::Relaxed);
+                                self.state = WorkerState::Stopped;
+                                break;
+                            }
                             WorkerMessage::Doorbell => {
                                 // we are already active and running, silently consume
                             }
--- a/src/device/xhci/command_ring.rs
+++ b/src/device/xhci/command_ring.rs
@@ -8,7 +8,10 @@ use std::sync::{
 use anyhow::anyhow;
 use tokio::{
     runtime,
-    sync::mpsc::{self, error::TryRecvError},
+    sync::{
+        mpsc::{self, error::TryRecvError},
+        oneshot,
+    },
 };
 use tracing::{debug, info, trace, warn};
 
@@ -16,6 +19,7 @@ use crate::device::{
     bus::BusDeviceRef,
     pci::constants::xhci::operational::{crcr, usbcmd},
     xhci::{
+        controller_reset::ResetSender,
         interrupter::EventSender,
         linked_ring::LinkedRing,
         slot_manager::SlotWorkerHandle,
@@ -27,6 +31,20 @@ use crate::device::{
 pub struct CommandRing {
     running: Arc<AtomicBool>,
     sender_to_worker: mpsc::UnboundedSender<WorkerMessage>,
+}
+
+#[derive(Debug)]
+pub struct CommandRingResetSender {
+    sender_to_worker: mpsc::UnboundedSender<WorkerMessage>,
+}
+
+impl ResetSender for CommandRingResetSender {
+    fn send_reset(&self, completion_notifier: oneshot::Sender<()>) -> anyhow::Result<()> {
+        self.sender_to_worker
+            .send(WorkerMessage::Reset(completion_notifier))?;
+
+        Ok(())
+    }
 }
 
 #[derive(Debug)]
@@ -54,7 +72,7 @@ enum WorkerMessage {
     SetDequeuePointerAndCS(u64, bool),
     Doorbell,
     Stop,
-    Reset,
+    Reset(oneshot::Sender<()>),
 }
 
 impl CommandRing {
@@ -145,8 +163,10 @@ impl CommandRing {
         }
     }
 
-    pub fn reset(&self) -> anyhow::Result<()> {
-        self.send_to_worker(WorkerMessage::Reset)
+    pub fn reset_sender(&self) -> CommandRingResetSender {
+        CommandRingResetSender {
+            sender_to_worker: self.sender_to_worker.clone(),
+        }
     }
 
     fn send_to_worker(&self, msg: WorkerMessage) -> anyhow::Result<()> {
@@ -171,8 +191,9 @@ impl CommandWorker {
         loop {
             match &self.state {
                 WorkerState::Stopped => match self.next_msg().await? {
-                    WorkerMessage::Reset => {
+                    WorkerMessage::Reset(completion) => {
                         self.commandring_running.store(false, Ordering::Relaxed);
+                        completion.send(()).ok();
                     }
                     WorkerMessage::SetDequeuePointerAndCS(dp, cs) => {
                         debug!("Updating command ring parameters: dp={dp:#x}, cs={cs}");
@@ -191,9 +212,10 @@ impl CommandWorker {
                     msg => warn!("Unexpected message: msg={msg:?}, state={:?}", self.state),
                 },
                 WorkerState::Idle => match self.next_msg().await? {
-                    WorkerMessage::Reset => {
+                    WorkerMessage::Reset(completion) => {
                         self.commandring_running.store(false, Ordering::Relaxed);
                         self.state = WorkerState::Stopped;
+                        completion.send(()).ok();
                     }
                     WorkerMessage::Doorbell => {
                         self.state = WorkerState::LookingForNewCommand;
@@ -210,9 +232,10 @@ impl CommandWorker {
                         };
 
                         match msg {
-                            WorkerMessage::Reset => {
+                            WorkerMessage::Reset(completion) => {
                                 self.commandring_running.store(false, Ordering::Relaxed);
                                 self.state = WorkerState::Stopped;
+                                completion.send(()).ok();
                                 break;
                             }
                             WorkerMessage::Doorbell => {
--- a/src/device/xhci/controller_reset.rs
+++ b/src/device/xhci/controller_reset.rs
@@ -64,3 +64,54 @@ impl ResetCoordinator {
         Ok(())
     }
 }
+
+#[cfg(test)]
+mod tests {
+    use tokio::sync::mpsc;
+
+    use super::*;
+
+    #[derive(Debug, Clone)]
+    struct TestResetSender {
+        completion_sender: mpsc::UnboundedSender<oneshot::Sender<()>>,
+    }
+
+    impl ResetSender for TestResetSender {
+        fn send_reset(&self, completion_notifier: oneshot::Sender<()>) -> anyhow::Result<()> {
+            self.completion_sender.send(completion_notifier)?;
+
+            Ok(())
+        }
+    }
+
+    #[tokio::test]
+    async fn reset_clears_hcrst_after_all_components_complete() {
+        let usbcmd = UsbcmdRegister::new();
+        usbcmd.write(usbcmd::HCRST);
+
+        let (completion_sender, mut completion_receiver) = mpsc::unbounded_channel();
+        let coordinator = ResetCoordinator {
+            usbcmd: usbcmd.clone(),
+            reset_senders: [
+                Box::new(TestResetSender {
+                    completion_sender: completion_sender.clone(),
+                }),
+                Box::new(TestResetSender {
+                    completion_sender: completion_sender.clone(),
+                }),
+                Box::new(TestResetSender { completion_sender }),
+            ],
+        };
+
+        let reset_task = tokio::spawn(async move { coordinator.reset().await });
+
+        for _ in 0..3 {
+            let completion = completion_receiver.recv().await.unwrap();
+            assert_eq!(usbcmd.read() & usbcmd::HCRST, usbcmd::HCRST);
+            completion.send(()).unwrap();
+        }
+
+        reset_task.await.unwrap().unwrap();
+        assert_eq!(usbcmd.read() & usbcmd::HCRST, 0);
+    }
+}
--- a/src/device/xhci/controller_reset.rs
+++ b/src/device/xhci/controller_reset.rs
@@ -1,0 +1,66 @@
+use tokio::{runtime, sync::oneshot};
+use tracing::error;
+
+use crate::device::{pci::constants::xhci::operational::usbcmd, xhci::registers::UsbcmdRegister};
+
+/// Sends reset requests to a component and reports when the reset finished.
+///
+/// The completion notifier must be triggered by the worker after it has applied
+/// its reset side effects.
+pub trait ResetSender: Send + Sync {
+    fn send_reset(&self, completion_notifier: oneshot::Sender<()>) -> anyhow::Result<()>;
+
+    fn reset(&self) -> anyhow::Result<oneshot::Receiver<()>> {
+        let (send, recv) = oneshot::channel();
+        self.send_reset(send)?;
+
+        Ok(recv)
+    }
+}
+
+/// Coordinates host controller reset across all resettable xHCI components.
+///
+/// The coordinator waits for `USBCMD.HCRST`, resets all registered components,
+/// and clears `HCRST` after all reset completion signals were received.
+pub struct ResetCoordinator {
+    usbcmd: UsbcmdRegister,
+    reset_senders: [Box<dyn ResetSender>; 3],
+}
+
+impl ResetCoordinator {
+    pub fn start(
+        usbcmd: UsbcmdRegister,
+        reset_senders: [Box<dyn ResetSender>; 3],
+        async_runtime: &runtime::Handle,
+    ) {
+        let coordinator = Self {
+            usbcmd,
+            reset_senders,
+        };
+
+        async_runtime.spawn(coordinator.run_loop());
+    }
+
+    async fn run_loop(self) {
+        loop {
+            self.usbcmd.hcrst_notification().await;
+
+            if self.usbcmd.read() & usbcmd::HCRST == 0 {
+                continue;
+            }
+
+            if let Err(err) = self.reset().await {
+                error!("failed to reset host controller: {err}");
+            }
+        }
+    }
+
+    async fn reset(&self) -> anyhow::Result<()> {
+        for reset_sender in &self.reset_senders {
+            reset_sender.reset()?.await?;
+        }
+        self.usbcmd.clear_hcrst();
+
+        Ok(())
+    }
+}
--- a/src/device/xhci/interrupter.rs
+++ b/src/device/xhci/interrupter.rs
@@ -58,6 +58,7 @@ struct EventWorker {
 enum InterrupterMessage {
     SendEvent(EventTrb),
     UpdateInterruptLine(Arc<dyn InterruptLine>),
+    Reset,
 }
 
 #[derive(Debug, Clone)]
@@ -115,6 +116,19 @@ impl Interrupter {
             sender: self.msg_sender.clone(),
         }
     }
+
+    pub fn reset(&self) -> anyhow::Result<()> {
+        self.registers.interrupt_management.write(0);
+        self.registers
+            .interrupt_moderation_interval
+            .write(IMOD_DEFAULT);
+        self.registers.erst_base_address.reset();
+        self.registers.erst_size.write(0);
+        self.registers.eventring_dequeue_pointer.write(0);
+        self.msg_sender.send(InterrupterMessage::Reset)?;
+
+        Ok(())
+    }
 }
 
 impl EventWorker {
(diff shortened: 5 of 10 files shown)


src/device/xhci/registers.rs
```diff
@@ -210,7 +210,7 @@ impl UsbcmdRegister {
         const BITMASK_PRESERVED: u64 = usbcmd::RS | usbcmd::INTE;
         // ... ignoring any other bits and printing a warning when attempted.
         if value & usbcmd::HCRST == usbcmd::HCRST {
-            info!("Host Controller Reset attempted; no action");
+            info!("Host Controller Reset requested");
         }
```
snue: Let's remove the check/log here. It was mostly so we see this happens. Now we check and handle the side effects on write, and we can just ignore it here. Let's move the `info!()` over there as well when we handle the reset. We still exclude `HCRST` from the warning for not preserved bits below.

src/device/xhci/slot_manager.rs
```diff
@@ -346,6 +356,17 @@ impl SlotWorker {
...
         Ok(CompletionCode::Success)
     }
+
+    async fn reset(&mut self) -> anyhow::Result<()> {
+        for slot_id in 1..=MAX_SLOTS as usize {
+            if let Some(slot) = self.slots[slot_id].as_mut() {
+                slot.pre_drop().await?;
+            }
+            self.slots[slot_id] = None;
+        }
```
snue, line 384: I'd prefer just iterating over `self.slots` here. I see a little issue that when `pre_drop` fails, we don't actually drop the slot (never set it to `None`). And also that a failure in one slot prevents dropping the others (partial resets). We should ensure the slots are cleared even if `pre_drop()` failed. We may just reinitialize the whole array before returning the result instead of clearing the individual members in the loop (something like `self.slots = [None; MAX_SLOTS];`).

Using `?` is generally convenient, but sometimes it leads to subtle issues.

src/device/xhci/slot_manager.rs
```diff
@@ -614,9 +635,9 @@ impl Slot {
 
     // call before dropping (disabling this slot)
     async fn pre_drop(&mut self) -> anyhow::Result<()> {
-        // disable EP0 if active
-        if !matches!(self.state, SlotState::Enabled) {
-            self.deconfigure_endpoint(1).await?;
+        // disable all active endpoints
+        for endpoint_id in 1..=31 {
+            self.deconfigure_endpoint(endpoint_id).await?;
```
snue: I don't remember the details why this function was required, but the new implementation looks more sensible 😁  ... wasn't the `Enabled` check inverted before?

There is a slight issue that when one endpoint fails the deconfigure, we won't attempt to deconfigure later ones. Alternatively you can collect all the results instead of using `?` and return the first error (or something more elaborate with anyhow) or `Ok()`  after the loop.

I also haven't checked what happens when you deconfigure an unconfigured endpoint? An iterator chain over all configured endpoints (with `.map()` or `.filter()` etc.) could be nice here, but we probably don't have such an iterator currently.

src/device/xhci/registers.rs
```diff
@@ -258,6 +266,10 @@ impl GenericRwRegister {
     pub fn write(&self, new_value: u64) {
         self.value.store(new_value, Ordering::Relaxed);
     }
+
+    pub fn reset(&self, value: u64) {
+        self.value.store(value, Ordering::Relaxed);
+    }
```
snue: This is the same as `write()`? Is the reset value not fixed?

src/device/pci/xhci.rs
```diff
@@ -114,7 +114,20 @@ impl<CRD: CompleteRealDevice> PciDevice for XhciController<CRD> {
 
         match req.addr {
             // xHC Operational Registers
-            offset::USBCMD => self.usbcmd.write(value),
+            offset::USBCMD => {
+                self.usbcmd.write(value);
+                if value & usbcmd::HCRST == usbcmd::HCRST {
+                    self.command_ring
+                        .reset()
+                        .expect("command worker should be alive");
```
snue: Instead of `expect`, can we just issue ` warn!` or `error!` if that ever happens? (same for the others)

And also let's add the `info!` here about the detected reset in any case.

src/device/xhci/interrupter.rs
```diff
@@ -134,44 +148,58 @@ impl EventWorker {
...
+        // clears that configuration and sends us back to waiting for ERSTBA.
+        loop {
+            self.wait_for_event_ring_configuration().await?;
+            self.event_ring.configure(
+                self.registers.erst_base_address.erstba(),
+                self.registers.erst_size.read() as u32,
+            );
+
+            while self.process_configured_message().await? {}
+        }
```
snue, line 171: This loop is nice 👍

src/device/xhci/interrupter.rs
```diff
@@ -115,6 +116,19 @@ impl Interrupter {
             sender: self.msg_sender.clone(),
         }
     }
+
+    pub fn reset(&self) -> anyhow::Result<()> {
+        self.registers.interrupt_management.reset(0);
+        self.registers
+            .interrupt_moderation_interval
+            .reset(IMOD_DEFAULT);
```
snue: Oh, I see. That's why the generic `reset` takes the value. Let's just call `write()`  and remove the redundant `reset()`. I know it looks kind of nice when you can just call `reset()` on everything 😉

snue (commented):
Really cool! I suspect you have tested that the controller reset works now when the guest reloads the xhci driver? This would make for a good integration test, or maybe addition to an existing test.

I left some comments. Looks pretty good already 👍

snue:
> Since usbvfiod still kept the old state, the command ring could still be considered running and CRCR writes during re-initialization produced warnings.

It is also broken in that state :wink: Thanks for fixing this :tada:

src/device/pci/xhci.rs
```diff
@@ -114,7 +115,21 @@ impl<CRD: CompleteRealDevice> PciDevice for XhciController<CRD> {
...
         match req.addr {
             // xHC Operational Registers
-            offset::USBCMD => self.usbcmd.write(value),
+            offset::USBCMD => {
+                self.usbcmd.write(value);
+                if value & usbcmd::HCRST == usbcmd::HCRST {
+                    info!("Host Controller Reset requested");
+                    if let Err(err) = self.command_ring.reset() {
+                        error!("failed to reset command ring: {err}");
+                    }
```
snue: nit: It's a matter of taste, but IMHO it would read better with a `.map_err(|e| error!("failed ..."))`

src/device/xhci/registers.rs
```diff
@@ -201,9 +209,6 @@ impl UsbcmdRegister {
         // Currently writable bits ...
         const BITMASK_PRESERVED: u64 = usbcmd::RS | usbcmd::INTE;
         // ... ignoring any other bits and printing a warning when attempted.
```
snue: nit: we should slightly adapt this comment now
```suggestion
        // ... ignoring any other bits and printing a warning when anything but a reset is attempted.
        // The MMIO write path handles the `HCRST` bit already.
```

src/device/xhci/slot_manager.rs
```diff
@@ -346,6 +356,21 @@ impl SlotWorker {
...
     }
+
+    async fn reset(&mut self) -> anyhow::Result<()> {
+        let mut result = Ok(());
+
+        for slot in self.slots.iter_mut() {
+            if let Some(slot) = slot.as_mut() {
+                if let Err(err) = slot.pre_drop().await {
+                    result = result.and(Err(err));
+                }
```
snue: Wouldn't this work as well?

```suggestion
                result = result.and(slot.pre_drop().await)
```

src/device/xhci/slot_manager.rs
```diff
@@ -614,12 +639,16 @@ impl Slot {
...
-        // disable EP0 if active
-        if !matches!(self.state, SlotState::Enabled) {
-            self.deconfigure_endpoint(1).await?;
+        let mut result = Ok(());
+
+        // disable all active endpoints
+        for endpoint_id in 1..=31 {
+            if let Err(err) = self.deconfigure_endpoint(endpoint_id).await {
+                result = result.and(Err(err));
+            }
```
snue: Same here, we can just `and` in the `Ok(())` values as well?

```suggestion
            result = result.and(self.deconfigure_endpoint(endpoint_id).await)
```

src/device/xhci/slot_manager.rs
```diff
@@ -614,12 +639,16 @@ impl Slot {
 
     // call before dropping (disabling this slot)
     async fn pre_drop(&mut self) -> anyhow::Result<()> {
-        // disable EP0 if active
-        if !matches!(self.state, SlotState::Enabled) {
-            self.deconfigure_endpoint(1).await?;
+        let mut result = Ok(());
+
+        // disable all active endpoints
+        for endpoint_id in 1..=31 {
```
snue: I fear calling this on every endpoint will cause a lot of log messages for ignored endpoint deconfigures. I think we can filter for the valid ones using the `self.endpoint_senders`? Or rather, instead of calling `deconfigure_endpoint`, which internally does a lookup on `self.endpoint_senders`, implement the "deconfigure" manually here. You need to terminate all senders and reset the senders array.

snue (commented):
Pretty nice now. Functionally this seems sufficient, but I did not test it yet. I left some more suggestions, and I'd like to have a test (unit or integration or both) before merge.

src/device/xhci/command_ring.rs
```diff
@@ -144,6 +145,11 @@ impl CommandRing {
         }
     }
 
+    pub fn reset(&self) -> anyhow::Result<()> {
+        self.running.store(false, Ordering::Relaxed);
```
lbeierlieb: The worker sets `running` to `false` when processing the message, so this assignment seems superfluous. This change also improves separation of concerns--`pub fn reset()` is only responsible for relaying the reset request to the worker, and the worker implements the logic.

src/device/xhci/command_ring.rs
lbeierlieb, line 247: Not directly related to the PR, but we should probably also set the `commandring_running` to `false` when stopping.
"It [the CRR bit] is cleared to ‘0’ when the Command Ring is “stopped” after writing a ‘1’
to the Command Stop (CS) or Command Abort (CA) flags, or if the R/S bit is cleared to ‘0’" (XHCI spec)

src/device/xhci/interrupter.rs
```diff
@@ -134,44 +148,58 @@ impl EventWorker {
...
-                        self.registers.eventring_dequeue_pointer.read(),
-                    );
-                    self.interrupt_line.interrupt();
-                    debug!("Sent event: {event_trb:?}");
-                }
-                InterrupterMessage::UpdateInterruptLine(interrupt_line) => {
-                    self.interrupt_line = interrupt_line;
-                    debug!("Updated interrupt line");
-                }
+    async fn process_configured_message(&mut self) -> anyhow::Result<bool> {
```
lbeierlieb: I would appreciate a short doc comment about the `bool` return value.

lbeierlieb: `process_configured_message` sounds like the method processes a configured message. However, the method is the method for processing messages while the event ring is configured.

src/device/pci/xhci.rs
```diff
@@ -114,7 +115,24 @@ impl<CRD: CompleteRealDevice> PciDevice for XhciController<CRD> {
...
+                        .ok();
+                    self.interrupter
+                        .reset()
+                        .map_err(|err| error!("failed to reset event ring: {err}"))
+                        .ok();
+                    self.slot_manager
+                        .reset()
+                        .map_err(|err| error!("failed to reset slots: {err}"))
+                        .ok();
+                }
```
lbeierlieb: The `XhciController` is supposed to only connect the components, not implement any logic itself. The `usbcmd.write` should recognize the reset bit and then either send instruct command ring, interrupter, and slot manager to reset itself, or we introduce a small dedicated reset component, which the usbcmd knows and which is responsible for informing all relevant other components.

src/device/pci/xhci.rs
```diff
@@ -114,7 +115,24 @@ impl<CRD: CompleteRealDevice> PciDevice for XhciController<CRD> {
 
         match req.addr {
             // xHC Operational Registers
-            offset::USBCMD => self.usbcmd.write(value),
+            offset::USBCMD => {
+                self.usbcmd.write(value);
+                if value & usbcmd::HCRST == usbcmd::HCRST {
```
lbeierlieb: We recognize that the bit is set here, but I did not see setting the HCRST bit to fulfill the spec's HCRST description: "This bit is cleared to ‘0’ by the Host Controller when the reset process is complete".
Did I miss the HCRST zero write?
Even then, the timing must be off. We should wait until all components that we instruct to reset report that they have finished resetting (might take a while, we are passing messages around), and then update HCRST to zero.

lbeierlieb (changes requested):
Thanks for the PR! I tested the feature with `rmmod xhci_pci; modprobe xhci_pci` and the controller came back to life successfully. The implement logic looks alright, but I do not fully agree with which logic is implemented where. I have to checked all commits yet.

src/device/xhci/interrupter.rs
snue: Right. Maybe just `process_message()`? And document the fact that this is called in configured state in the comment together with explaining the return value?

lbeierlieb: on second thought: Would it not be more straightforward to loop in `process_messages`/`run_configured`, with a `break` when we detect a reset message? Getting rid of the bool return value and the while loop in the `run_loop`.

src/device/pci/xhci.rs
snue: Yes, this sounds like a clean design, but also a bit overengineered IMHO. So far I am fine with the controller implementing the logic that concerns orchestrating multiple components and leaving any register access side effects and everything that is internal to a component to these. Our register abstractions cannot really deal with causing side effects currently, and I would like to keep it that way for now.

src/device/pci/xhci.rs
snue: We only check the written value and never actually store the bit, as the guest has no guarantee it will be able to see the bit flip (as far as I interpret the specification). When this was a `noop`, we were just "instantly done". We should probably document this behavior here, but I would like to see a stress test hitting register writes when we are not ready for them before actually implementing the observable set/clear on this bit.

src/device/pci/xhci.rs
lbeierlieb: We have
```
#[derive(Debug, Default, Clone)]
pub struct ErstbaRegister {
    value: Arc<AtomicU64>,
    notify: Arc<Notify>,
}

impl ErstbaRegister {
    pub fn read(&self) -> u64 {
        self.value.load(Ordering::Relaxed)
    }

    pub fn erstba(&self) -> u64 {
        self.value.load(Ordering::Relaxed) & !0x1f
    }

    pub fn write(&self, new_value: u64) {
        self.value.store(new_value, Ordering::Relaxed);
        self.notify.notify_waiters();
    }
...
```
so that the interrupter can notice writes.

A reset component could listen the same way for usbcmd writes. The majority of notifications would not result in an action, but the total number of usbcmd writes is low, so overall no performance issue.

I would really like to keep any logic out of the XhciController. Controller reset is probably the case that would be the most acceptable to handle directly, but I fear it's a slippery slope and will cause more logic creeping into the top-level controller.

src/device/pci/xhci.rs
snue: If we want to make this bit observable, the special `Reset` component owned by `usbcmd` you proposed above suddenly becomes a more appealing option :wink:

lbeierlieb: suggestion:

- `usbcmd::write` notices HCRST set, notifies reset component; `write` also preserves HCRST
- reset component triggers reset in other components
- reset component waits for all components to report that they finished the reset operation
- reset component sets HCRST bit to zero

snue: >     * `usbcmd::write` notices HCRST set, notifies reset component; `write` also preserves HCRST

Yes.

> 
>     * reset component triggers reset in other components

Through shared references to the relevant components I presume? Or just message passing?

> 
>     * reset component waits for all components to report that they finished the reset operation

Yes. Just `await` a bunch of Futures with the reset component being a separate task. Do we need extra cleanup for it?

> 
>     * reset component sets HCRST bit to zero

It also has access to the `Arc<AtomicU64>` value of `usbcmd` for that and take care only to handle `HCRST`? Or do we go for separate `AtomicBool` and assemble them in `usbcmd`? We could revert the recent refactor of the `running` bit from `bool`, as we have different "audiences" for these bits.

lbeierlieb: > Through shared references to the relevant components I presume? Or just message passing?

The same way as currently the XhciController does it. All components are split in split in half--the "front end" is what the controller stores, storing atomics and message senders, the "back end" is a worker that owns state and typically processes messages coming from the front end. The reset component could take the front ends directly and call `reset` on them, which will send reset messages to the back-end worker. Alternatively, the front ends offer a `create_reset_sender` function, a thin wrapper around the sender to the respective back end.



> Yes. Just `await` a bunch of Futures with the reset component being a separate task. Do we need extra cleanup for it?

cleanup for what? :see_no_evil: 

> It also has access to the `Arc<AtomicU64>` value of `usbcmd` for that and take care only to handle `HCRST`? Or do we go for separate `AtomicBool` and assemble them in `usbcmd`? We could revert the recent refactor of the `running` bit from `bool`, as we have different "audiences" for these bits.

Either direct access to the `Arc<AtomicU32>` with `fetch_and` or implement `UsbcmdRegister::unsetHcrst` and share the `UsbcmdRegister` between controller and reset component with an `Arc`.

snue: > cleanup for what? 🙈

All good :grin: the design checks out. Just a `fetch_and` on the `Atomic` sounds right as well.

lbeierlieb: I drafted a bit.

```
trait ResetSender {
    fn send_reset(&self, completion_notifier: oneshot::Sender<()>) -> anyhow::Result<()>;
    fn reset(&self) -> oneshot::Receiver<()> {
        let (send, recv) = oneshot::channel();
        self.send_reset(send);
        recv
    }
}
```

Each resetable component has to implement a sender and a function that returns such a sender. Command-ring example:

```
struct CommandRingResetSender {
     sender_to_worker: mpsc::UnboundedSender<WorkerMessage>,
}

impl ResetSender for CommandRingResetSender {
    fn send_reset(&self, completion_notifier: oneshot::Sender<()>) -> anyhow::Result<()> {
        self.sender_to_worker.send(WorkerMessage::Reset(completion_notifier)).map_err(|e| e.into())
    }
}

impl CommandRing {
    pub fn reset_sender(&self) -> CommandRingResetSender {
        CommandRingResetSender {
            sender_to_worker: self.sender_to_worker.clone(),
        }
    }
}
```

The reset component:

```
struct ResetCoordinator<const N: usize> {
    hcrst_notify: Notify,
    reset_senders: [Box<dyn ResetSender>; N],
    usbcmd: Arc<AtomicU32>,
}

impl<const N: usize> ResetCoordinator<N> {
    async fn run_loop(&self) -> anyhow::Result<()> {
        loop {
            self.hcrst_notify.notified().await;

            for reset_sender in &self.reset_senders {
                reset_sender.reset().await?;
            }

            self.usbcmd
                .fetch_and(!(usbcmd::HCRST as u32), Ordering::Relaxed);
        }
    }
}
```
```

Comment (snue) [reply src/device/pci/xhci.rs]: Ok, that's a nice separation of concerns. And the only reason for the `reset().await` to fail is when a channel is closed, in which case we don't need to bother sending the other resets or ever clearing the bit?

Comment (lbeierlieb) [reply src/device/pci/xhci.rs]: yep, exactly; `reset()` return value is directly the `oneshot::Receiver`, whose Future returns [RecvError](https://docs.rs/tokio/latest/tokio/sync/oneshot/error/struct.RecvError.html) when the sender is dropped.

File: src/device/xhci/command_ring.rs
Comment (lbeierlieb) [reply src/device/xhci/command_ring.rs:247]: Sorry, please remove again 🙈  we set the state to `Stopping`, in which we set CRR to false and then transition to `Stopped`.

File: src/device/xhci/interrupter.rs [hunk 17]
```diff
@@ -1,23 +1,46 @@
...
 use crate::device::interrupt_line::{DummyInterruptLine, InterruptLine};
 use crate::device::pci::constants::xhci::runtime::IMOD_DEFAULT;
+use crate::device::xhci::controller_reset::ResetSender;
 use crate::device::xhci::event_ring::EventRing;
 use crate::device::xhci::registers::{ErstbaRegister, GenericRwRegister};
 use crate::device::xhci::trb::EventTrb;
 use std::sync::Arc;
 
-#[derive(Debug)]
+#[derive(Debug, Clone)]
```
lbeierlieb: Why this change? You never use the clone functionality, the code works perfectly without Clone.

src/device/xhci/interrupter.rs
```diff
@@ -1,23 +1,46 @@
...
 pub struct Interrupter {
     pub registers: InterrupterRegisters,
     /// Transmits events to send to the worker
     msg_sender: mpsc::UnboundedSender<InterrupterMessage>,
 }
 
+#[derive(Debug, Clone)]
```
lbeierlieb: The idea is to create an `InterrupterResetSender` once and give it to the `ResetCoordinator`. There is no need for cloning intended, so I would rather not remove the `Clone` from here.

src/device/xhci/interrupter.rs
```diff
@@ -1,23 +1,46 @@
...
+
+impl ResetSender for InterrupterResetSender {
+    fn send_reset(&self, completion_notifier: oneshot::Sender<()>) -> anyhow::Result<()> {
+        self.registers.interrupt_management.write(0);
+        self.registers
+            .interrupt_moderation_interval
+            .write(IMOD_DEFAULT);
+        self.registers.erst_base_address.reset();
+        self.registers.erst_size.write(0);
+        self.registers.eventring_dequeue_pointer.write(0);
```
lbeierlieb: The interrupter worker should execute all this code once it receives the message. `send_reset` should really only do sending the reset message to the worker.

src/device/xhci/interrupter.rs
```diff
@@ -173,7 +190,10 @@ impl EventWorker {
                 msg = self.msg_recv.recv() => match msg.ok_or_else(|| anyhow!("event channel closed"))? {
                     InterrupterMessage::SendEvent(_) => {}
                     InterrupterMessage::UpdateInterruptLine(interrupt_line) => self.interrupt_line = interrupt_line,
-                    InterrupterMessage::Reset => self.event_ring.reset(),
+                    InterrupterMessage::Reset(completion) => {
+                        self.event_ring.reset();
+                        completion.send(()).ok();
```
lbeierlieb: Please use `completion.send_anyhow(())?;` (needs `use crate::oneshot_anyhow::SendWithAnyhowError;`)

src/device/xhci/interrupter.rs
```diff
@@ -198,8 +218,9 @@ impl EventWorker {
                     self.interrupt_line = interrupt_line;
                     debug!("Updated interrupt line");
                 }
-                InterrupterMessage::Reset => {
+                InterrupterMessage::Reset(completion) => {
                     self.event_ring.reset();
+                    completion.send(()).ok();
```
lbeierlieb: Please use `completion.send_anyhow(())?;` (needs `use crate::oneshot_anyhow::SendWithAnyhowError;`)

src/device/xhci/registers.rs
```diff
@@ -188,15 +188,17 @@ impl DcbaapRegister {
 }
 
 /// USB Command Register (chapter 5.4.1)
-#[derive(Debug)]
+#[derive(Debug, Clone)]
 pub struct UsbcmdRegister {
     value: Arc<AtomicU32>,
+    notify: Arc<Notify>,
```
lbeierlieb: This notify is used only for reset notify, so reset_notify or similar would be a better name, then one look at the struct is enough to understand the purpose; otherwise, you have to look through the code and the usages of notify to understand the purpose.

src/device/xhci/slot_manager.rs
```diff
@@ -23,13 +24,31 @@ use crate::{
     oneshot_anyhow::SendWithAnyhowError,
 };
 
-#[derive(Debug)]
+#[derive(Debug, Clone)]
```
lbeierlieb: another unnecessary Clone

src/device/xhci/slot_manager.rs
```diff
@@ -23,13 +24,31 @@ use crate::{
...
+pub struct SlotManagerResetSender {
+    config_reg: ConfigureRegister,
+    dcbaap: DcbaapRegister,
+    msg_send: mpsc::UnboundedSender<SlotMessage>,
+}
+
+impl ResetSender for SlotManagerResetSender {
+    fn send_reset(&self, completion_notifier: oneshot::Sender<()>) -> anyhow::Result<()> {
+        self.config_reg.reset();
+        self.dcbaap.reset();
```
lbeierlieb: should be in the worker

src/device/xhci/slot_manager.rs
```diff
@@ -23,13 +24,31 @@ use crate::{
...
 pub struct SlotManager {
     pub config_reg: ConfigureRegister,
     pub dcbaap: DcbaapRegister,
     msg_send: mpsc::UnboundedSender<SlotMessage>,
 }
 
+#[derive(Debug, Clone)]
+pub struct SlotManagerResetSender {
+    config_reg: ConfigureRegister,
+    dcbaap: DcbaapRegister,
```
lbeierlieb: not necessary when all the reset logic is in the worker

src/device/xhci/slot_manager.rs
```diff
@@ -236,7 +255,11 @@ impl SlotWorker {
                     let result = slot.handle_reset_device().await?;
                     sender.send_anyhow(result)?;
                 }
-                SlotMessage::ResetAllSlots => self.reset().await?,
+                SlotMessage::ResetAllSlots(completion) => {
+                    let result = self.reset().await;
+                    completion.send(()).ok();
+                    result?;
```
lbeierlieb: if the reset failed, then we don't necessarily want to signal successful completion of the reset.
Suggestion:
```
self.reset().await?;
completion.send_anyhow(())?;
```

lbeierlieb (changes requested):
Thank you for integrating the suggested changes! We are getting there, but there is some cleanup necessary to properly divide responsibilities.

src/device/xhci/slot_manager.rs
```diff
@@ -346,6 +379,17 @@ impl SlotWorker {
...
         Ok(CompletionCode::Success)
     }
+
+    async fn reset(&mut self) -> anyhow::Result<()> {
+        let mut result = Ok(());
+
+        for slot in self.slots.iter_mut().filter_map(Option::as_mut) {
+            result = result.and(slot.pre_drop().await);
+        }
+        self.slots = [const { None }; MAX_SLOTS as usize].into();
```
snue: We can get rid of this by also just using `Option::take` like below, right?

src/device/xhci/registers.rs
```diff
@@ -200,26 +210,42 @@ impl UsbcmdRegister {
...
         }
 
-        let value = value & BITMASK_PRESERVED;
-        // SAFETY: USBCMD is defined as 32 bit and the masks used above enforce it.
+        // Preserve an ongoing reset until the reset coordinator clears HCRST.
         self.value
-            .store(value.try_into().unwrap(), Ordering::Relaxed);
+            .fetch_update(Ordering::Relaxed, Ordering::Relaxed, |current| {
+                let hcrst = current as u64 & usbcmd::HCRST;
+                let value = (value & BITMASK_PRESERVED) | (value & usbcmd::HCRST) | hcrst;
```
snue:
```suggestion
                let value = (value & BITMASK_WRITABLE) | hcrst;
```

src/device/xhci/registers.rs
```diff
@@ -200,26 +210,42 @@ impl UsbcmdRegister {
...
+                value & !BITMASK_WRITABLE
             );
         }
 
-        let value = value & BITMASK_PRESERVED;
-        // SAFETY: USBCMD is defined as 32 bit and the masks used above enforce it.
+        // Preserve an ongoing reset until the reset coordinator clears HCRST.
         self.value
-            .store(value.try_into().unwrap(), Ordering::Relaxed);
+            .fetch_update(Ordering::Relaxed, Ordering::Relaxed, |current| {
```
snue, line 222: Please double-check with @lbeierlieb whether `Relaxed` is really sufficient here.

lbeierlieb: There is no other memory involved that we prepare and **release** with this write, so `Relaxed` is correct here.

nix/checks/controller-reset.nix
```diff
@@ -0,0 +1,130 @@
...
+      cloud_hypervisor.wait_until_succeeds("mount LABEL=reset-test /mnt", timeout=120)
+
+      # Create a file and compute a checksum.
+      cloud_hypervisor.succeed(f"dd if=/dev/urandom of=/tmp/file-{i} count=8 bs=1M", timeout=60)
+      out = cloud_hypervisor.succeed(f"sha256sum /tmp/file-{i}", timeout=60)
+      hash_tmp = out.split()
+      print(f"hash_tmp: {hash_tmp}")
+
+      # Copy the file on the blockdevice.
+      cloud_hypervisor.succeed(f"cp /tmp/file-{i} /mnt/file", timeout=60)
```
Noi0103: I would like for this to be shorter/do less. The idea in the blockdevice test is to make sure we do not accidentally corrupt the data when we copy it over. I do not think we need to test that here too.

We could use the same content and make sure read caches are cleared, make minimal changes (like add one number on each loop) or even not do this data integrity check at all.

nix/checks/controller-reset.nix
```diff
@@ -0,0 +1,130 @@
...
+      cloud_hypervisor.succeed("sync", timeout=60)
+      cloud_hypervisor.succeed("umount /mnt", timeout=60)
+
+      # Reloading xhci_pci makes the Linux guest reset the host controller and
+      # reprogram command ring, event ring and slot state afterwards.
```
Noi0103: nit: I would not consider the specific affected rings important on the nix integration test level.

nix/checks/controller-reset.nix
```diff
@@ -0,0 +1,130 @@
...
+      # Check if the file checksum changed.
+      out = cloud_hypervisor.wait_until_succeeds("lsusb -d ${testutils.blockdeviceVendorId}:${testutils.blockdeviceProductId}", timeout=120)
+      search("ID ${testutils.blockdeviceVendorId}:${testutils.blockdeviceProductId} QEMU QEMU USB HARDDRIVE", out)
+      cloud_hypervisor.wait_until_succeeds("lsblk /dev/sd*", timeout=120)
+      cloud_hypervisor.wait_until_succeeds("mount LABEL=reset-test /mnt", timeout=120)
+      out = cloud_hypervisor.succeed("sha256sum /mnt/file", timeout=60)
+      hash_mnt = out.split()
+      print(f"hash_mnt: {hash_mnt}")
+      if (hash_tmp[0] != hash_mnt[0]):
+        raise RequestedAssertionFailed("The checksum changed after xhci_pci reload and host controller reset.")
```
Noi0103: Same as above.
Is it possible that we read different data after a host controller reset or would it break functionality completely so the guest stops using the emulated xhci?

nix/checks/controller-reset.nix
```diff
@@ -0,0 +1,130 @@
...
+
+      # Confirm write I/O still works after the controller reset.
+      cloud_hypervisor.succeed(f"echo after-reset-{i} > /mnt/after-reset-{i}.txt", timeout=60)
+      out = cloud_hypervisor.succeed(f"cat /mnt/after-reset-{i}.txt", timeout=60)
```
Noi0103: It might be possible the kernel did not write on the hardware yet. The `sync` command above would guarantee that.

While me might not need to check for full data integrity, triggering IO functionality after the reset is a good idea.

nix/checks/controller-reset.nix
```diff
@@ -0,0 +1,130 @@
...
+    search("ID ${testutils.blockdeviceVendorId}:${testutils.blockdeviceProductId} QEMU QEMU USB HARDDRIVE", out)
+    cloud_hypervisor.wait_until_succeeds("lsblk /dev/sda", timeout=120)
+
+    # Test partitioning.
+    cloud_hypervisor.succeed("echo ',,L' | sfdisk --label=gpt /dev/sda", timeout=60)
+
+    # The Script is sometimes too fast for the nested guest to detect the new partition.
+    cloud_hypervisor.wait_until_succeeds("lsblk /dev/sda1", timeout=60)
+
+    # Make a filesystem.
```
Noi0103: Just an idea but maybe we could even get away with no filesystem at all and just using `dd`.

nix/checks/controller-reset.nix
```diff
@@ -0,0 +1,130 @@
...
+  builtins.map (usbVersion: {
+    name = "controller-reset-usb-${builtins.replaceStrings [ "." ] [ "_" ] usbVersion}";
+    value = testutils.mkUsbTest {
+      name = "controller-reset-usb-${usbVersion}";
+      virtualDevices = [
+        {
+          type = "blockdevice";
+          inherit usbVersion;
+          udevRule.symlink = "reset-device";
+          attachedOnStartup = "guest";
```
Noi0103: Why not use the defaults?

I see that the defaults are not described in the functions doc comment. That is probably something we should add later on.

Noi0103 (commented):
Pretty good overall. I would like the test script to have a few less overlaps with the existing blockdevice tests.

nix/checks/controller-reset.nix
ziyifu225: Simplified the comment

nix/checks/controller-reset.nix
ziyifu225: For the single device, of course not. It may happens in multiple devices case. But for now we don't need multi-device testing. So you are right.

nix/checks/controller-reset.nix
ziyifu225: I split the switch to the `dd`-based version into a separate commit because I am not fully sure it is strictly better. My understanding is that it makes the test simpler and more focused on raw I/O after the controller reset, while dropping filesystem-level coverage such as partition detection and mounting. But this filesystem-level coverage seems to already be covered by the existing blockdevice test.

nix/checks/controller-reset.nix
ziyifu225: After discuss with @Noi0103 , he will fix this.

src/device/xhci/registers.rs
```diff
@@ -200,26 +210,42 @@ impl UsbcmdRegister {
     pub fn write(&self, value: u64) {
         // Currently writable bits ...
         const BITMASK_PRESERVED: u64 = usbcmd::RS | usbcmd::INTE;
-        // ... ignoring any other bits and printing a warning when attempted.
-        if value & usbcmd::HCRST == usbcmd::HCRST {
-            info!("Host Controller Reset attempted; no action");
-        }
-        if value & !(BITMASK_PRESERVED | usbcmd::HCRST) != 0 {
+        const BITMASK_WRITABLE: u64 = BITMASK_PRESERVED | usbcmd::HCRST;
```
snue: After the refactor, now that the register triggers the side effects, this is just the `BITMASK_PRESERVED`.


```rust
        const BITMASK_PRESERVED: u64 = usbcmd::RS | usbcmd::INTE | usbcmd::HCRST;
```

The comment below also needs an update. It can be shortened / cut off after "printing a warning".

src/device/xhci/command_ring.rs
```diff
@@ -8,27 +8,45 @@ use std::sync::{
...
+        controller_reset::ResetSender,
         interrupter::EventSender,
         linked_ring::LinkedRing,
         slot_manager::SlotWorkerHandle,
         trb::{CommandTrb, CommandTrbVariant, CompletionCode, EventTrb},
     },
 };
 
-#[derive(Debug)]
+#[derive(Debug, Clone)]
```
snue: Why is this `Clone` needed? Let's try to avoid it?

src/device/xhci/command_ring.rs
```diff
@@ -8,27 +8,45 @@ use std::sync::{
...
 pub struct CommandRing {
     running: Arc<AtomicBool>,
     sender_to_worker: mpsc::UnboundedSender<WorkerMessage>,
 }
 
+#[derive(Debug, Clone)]
```
snue: What's different about the `CommandRingResetSender` that it requires `Clone`? The others seem to get away without it, and you create explicit senders with clones of the mpsc senders.

src/device/xhci/controller_reset.rs
```diff
@@ -0,0 +1,120 @@
...
+    }
+}
+
+#[cfg(test)]
+mod tests {
+    use tokio::sync::mpsc;
+
+    use super::*;
+
+    #[derive(Debug, Clone)]
```
snue, line 74: It doesn't look like this `Clone` is necessary, either.

nix/checks/controller-reset.nix
```diff
@@ -0,0 +1,93 @@
...
+      out = cloud_hypervisor.wait_until_succeeds("lsusb -d ${testutils.blockdeviceVendorId}:${testutils.blockdeviceProductId}", timeout=120)
+      search("ID ${testutils.blockdeviceVendorId}:${testutils.blockdeviceProductId} QEMU QEMU USB HARDDRIVE", out)
+      cloud_hypervisor.wait_until_succeeds("lsblk /dev/sda", timeout=120)
+      cloud_hypervisor.succeed(f"printf after-reset-{i} > /tmp/after-reset-{i}.txt", timeout=60)
+      cloud_hypervisor.succeed(f"dd if=/tmp/after-reset-{i}.txt of=/dev/sda bs=512 seek=2048 count=1 conv=sync,fsync status=none", timeout=60)
+      cloud_hypervisor.succeed("sync", timeout=60)
+      cloud_hypervisor.succeed("echo 3 > /proc/sys/vm/drop_caches", timeout=60)
+      cloud_hypervisor.succeed(f"dd if=/dev/sda of=/tmp/read-after-reset-{i}.txt bs=512 skip=2048 count=1 status=none", timeout=60)
+      out = cloud_hypervisor.succeed(f"grep -ao after-reset-{i} /tmp/read-after-reset-{i}.txt", timeout=60)
+      search(f"after-reset-{i}", out)
```
snue: The search isn't really helpful here. The `grep` already ensures the same and uses `.succeed`. Failures would actually contain the searched string as well:

```
grep: /tmp/read-after-reset-1.txt: No such file or directory
```

You can just remove this `search`. The rest of the test logic looks good.

nix/checks/controller-reset.nix
```diff
@@ -0,0 +1,93 @@
...
+                --initramfs ${netboot.initrd} \
+                --user-device socket=${testutils.usbvfiodSocket} \
+                --net "tap=tap0,mac=,ip=192.168.100.1,mask=255.255.255.0"
+            '';
+          };
+        };
+    };
+  };
+
+  singleBlockDeviceTestScript = ''
```
snue: nit: it could also be called `controllerResetTestScript`.

nix/checks/controller-reset.nix
```diff
@@ -0,0 +1,93 @@
+{
+  lib,
+  cloud-hypervisor,
+  usbvfiod,
+  testutils,
+}:
+let
+  systemd-config = args: {
```
snue, line 8: I talked to @Noi0103 about moving the systemd service setup to the test utils function as well and just controlling the aspects we are interested in through additional args. No need to change anything here, just a heads-up that we may be removing more boilerplate soon.

snue (commented):
Just a couple remarks left. I really like the new design. Love the added tests.

nix/checks/controller-reset.nix
```diff
@@ -0,0 +1,93 @@
...
+    name = "controller-reset-usb-${builtins.replaceStrings [ "." ] [ "_" ] usbVersion}";
+    value = testutils.mkUsbTest {
+      name = "controller-reset-usb-${usbVersion}";
+      virtualDevices = [
+        {
+          type = "blockdevice";
+          inherit usbVersion;
+        }
+      ];
+      testScript = singleBlockDeviceTestScript;
```
snue: Or just call it `testScript` above and `inherit` here. Now that the tests are split up into separate files, that looks like a clean approach.

snue (approved):
Thanks! This looks good now. Just a bit of cleanup on the commit history (squashing the fixups) and I believe this can be merged. You can leave the "refactor" in the commit history. The commit messages are very nicely explaining the reasoning.

src/device/xhci/registers.rs
```diff
@@ -206,26 +208,41 @@ impl UsbcmdRegister {
...
+            .fetch_update(Ordering::Relaxed, Ordering::Relaxed, |current| {
+                let hcrst = current as u64 & usbcmd::HCRST;
+                let value = (value & BITMASK_PRESERVED) | hcrst;
+                // SAFETY: USBCMD is defined as 32 bit and the masks used above enforce it.
+                Some(value.try_into().unwrap())
+            })
+            .unwrap();
+        if value & usbcmd::HCRST == usbcmd::HCRST {
+            self.reset_notify.notify_waiters();
+        }
```
snue, line 232: The MMIO path does not issue an `info` message anymore for this. We should add that here instead.

src/device/xhci/controller_reset.rs
```diff
@@ -0,0 +1,120 @@
...
+        loop {
+            self.usbcmd.hcrst_notification().await;
+
+            if self.usbcmd.read() & usbcmd::HCRST == 0 {
+                continue;
+            }
+
+            self.reset()
+                .await
+                .map_err(|err| error!("failed to reset host controller: {err}"))
```
lbeierlieb: nit: looks like you want to use `.context()` here

src/device/xhci/registers.rs
snue: Oh, I see the message went to the `controller_reset` and wasn't completely gone. Thanks for moving it here. I think this is the better place. :+1:

src/device/xhci/controller_reset.rs
```diff
@@ -0,0 +1,118 @@
...
+                .ok();
```
lbeierlieb: You want to do something with the potential error if it exists, and discard the result afterwards. `map_err` and `ok` do the right thing, but the code can be more clear when you use functions intended for what you need.

`map_err`'s intention is to transform the error into something else; you use the function solely to produce a side effect.
`ok` transform into an `Option`. You do not want an `Option`, you convert because you need a value without `must_use`.

I see two idiomatic ways to express the intention:

```suggestion
            let _ = self.reset()
                .await
                .inspect_err(|err| error!("failed to reset host controller: {err}"));
```

```suggestion
            if let Err(err) = self.reset().await {
                error!("failed to reset host controller: {err}")
            }
```

lbeierlieb (approved):
Looks good, just one nit. Thanks!

PR #273, opened 2026-07-16
Detach already attached devices and remove AlreadyAttached response

snue:
When an already attached device is reattached through the hotplug socket, the remote binary already performed a reset and likely broke the attached device in the process. Instead of refusing the attach, perform a detach of the conflicting device in the controller.

Remove the now unused `AlreadyAttached` response from the protocol definition. There is no sane place where it could be useful. Callers should inspect the list of attached devices first, to avoid breaking devices that are in-use by a guest.

BREAKING: modifies the public protocol response codes.

### DIFF
--- a/src/device/xhci/port.rs
+++ b/src/device/xhci/port.rs
@@ -6,7 +6,7 @@ use tokio::{
     sync::{mpsc, oneshot},
 };
 use tokio_util::sync::CancellationToken;
-use tracing::{debug, info, warn};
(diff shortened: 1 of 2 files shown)


lbeierlieb (approved):
Nice and easy. Looks good to me.

snue:
This could still do with a test, but I won't get to it right now :disappointed_relieved: 

Will merge as is and add a test eventually.

PR #281, opened 2026-07-20
Refactor and split nix integration tests into separate files

Noi0103:
While working towards the file descriptor cli argument, writing a new test in the existing test landscape felt harder than it should. This refactor splits the different kinds of tests and gives individual control over the declared systemd configuration parts.

### DIFF
--- a/nix/checks/default.nix
+++ b/nix/checks/default.nix
@@ -1,0 +1,48 @@
+{
+  lib,
+  pkgs,
+  usbvfiod,
+}:
+let
+  testutils = import ./testutils.nix { inherit lib pkgs; };
+  mkNixosIntegrationTest =
+    file:
+    pkgs.callPackage file {
+      inherit (pkgs) cloud-hypervisor;
+      inherit usbvfiod testutils;
+    };
+in
+{
+  multiple-blockdevices = mkNixosIntegrationTest ./multiple-blockdevices.nix;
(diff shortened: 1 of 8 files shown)


nix/checks/testutils.nix
```diff
@@ -1,12 +1,9 @@
...
-  This file contains integration tests for usbvfiod.
-*/
 {
   lib,
   pkgs,
-  usbvfiod,
+  ...
 }:
-let
+rec {
```
lbeierlieb: Please don't use `rec` here. Use `let` to define everything you define here, and then after `in` create an attribute set with the attributes that should be available/is necessary for the other files to define tests. For example, `globalTimeout` or `nestedPythonClass` do not have to be "exported".

nix/checks/default.nix
```diff
@@ -0,0 +1,48 @@
+{
```
lbeierlieb, line 1: More usage of `callPackage` can make this file even cleaner, but we can leave that for a follow up.

lbeierlieb (changes requested):
I really like the split, but the test utils should differentiate between "public API" and "internal helpers".

nix/checks/default.nix
Noi0103: Yes, lets leave that for later. CallPackage adds some Attributes, so this is not a drop in replacement.

lbeierlieb (approved):

PR #285, opened 2026-07-21
Implement a simple version of immediate data usage

Noi0103:
This does no checks that are in theory required by the spec.

For example, if immediate data is used it shall be the only TRB of a TD.

### DIFF
--- a/src/device/xhci/endpoint_handle.rs
+++ b/src/device/xhci/endpoint_handle.rs
@@ -361,12 +361,19 @@ impl ControlRequestParser {
                 },
                 ControlRequestParserState::SetupStageConsumed => match transfer_trb {
                     TransferTrbVariant::DataStage(data_trb_data) => {
-                        if data_trb_data.immediate_data {
-                            todo!("immediate data on data stage trb")
-                        }
-                        let mut data = vec![0; self.request_builder.length as usize];
-                        self.dma_bus
-                            .read_bulk(data_trb_data.data_pointer, &mut data);
+                        let data = if data_trb_data.immediate_data {
+                            if self.request_builder.length > 8 {
+                                todo!("using IDT with length > 8");
+                            }
+                            data_trb_data.data_pointer.to_le_bytes()
+                                [..self.request_builder.length as usize]
+                                .to_vec()
+                        } else {
+                            let mut data = vec![0; self.request_builder.length as usize];
+                            self.dma_bus
+                                .read_bulk(data_trb_data.data_pointer, &mut data);
+                            data
+                        };
 
                         self.request_builder.data = Some(data);
                         self.request_builder.data_pointer = Some(data_trb_data.data_pointer);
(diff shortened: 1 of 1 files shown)


src/device/xhci/endpoint_handle.rs
```diff
@@ -361,12 +361,16 @@ impl ControlRequestParser {
...
                     TransferTrbVariant::DataStage(data_trb_data) => {
-                        if data_trb_data.immediate_data {
-                            todo!("immediate data on data stage trb")
-                        }
-                        let mut data = vec![0; self.request_builder.length as usize];
-                        self.dma_bus
-                            .read_bulk(data_trb_data.data_pointer, &mut data);
+                        let data = if data_trb_data.immediate_data {
+                            data_trb_data.data_pointer.to_le_bytes()
+                                [..self.request_builder.length as usize]
```
lbeierlieb, line 369: You should only use the `[]` indexing operator if you know or verified that the index is valid. Here, nothing prevents a driver to set IDT and transfer length > 8, in which case this code will panic. The clean solution is probably to send an event with status code TrbError, but I'm fine if we for now run into another `todo!()` when this scenario occurs.

You can check the length beforehand or use `..to_le_bytes().get(..self.request_builder.length as usize]` and handle the `Option`.

src/device/xhci/endpoint_handle.rs
```diff
@@ -447,11 +451,15 @@ impl<ROEH: RealOutEndpointHandle> EndpointHandle for OutEndpointHandle<ROEH> {
         let transfer_trb = TransferTrbVariant::parse(trb.buffer);
         match &transfer_trb {
             TransferTrbVariant::Normal(normal_data) => {
-                if normal_data.immediate_data {
-                    todo!("immediate data on normal trb")
-                }
-                let mut data = vec![0; normal_data.transfer_length as usize];
-                self.dma_bus.read_bulk(normal_data.data_pointer, &mut data);
+                let data = if normal_data.immediate_data {
+                    normal_data.data_pointer.to_le_bytes()[..normal_data.transfer_length as usize]
```
lbeierlieb, line 461: Same here.

lbeierlieb (commented):
The happy path looks alright, we should handle an error case explicitly. Also, let's see if we can test this functionality with a serial device before merging.

Noi0103:
I tested these changes with a serial USB adapter and ran into a panic that is fixed by #286.

The test setup looked like this:
```mermaid
graph TD
host_usb_A-->serial_A
serial_A-->serial_male_to_male_connector
serial_male_to_male_connector-->serial_B
serial_B-->host_usb_B
host_usb_B-->usbvfiod
usbvfiod-->cloud_hypervisor_guest
```

I was able to `echo "hello" > ttyUSB0` in the "cloud_hypervisor_guest" and use minicom on the host to read the hello string.

lbeierlieb (approved):

PR #286, opened 2026-07-21
Fix missing bytes subtraction

Noi0103:
Testing with the debug build and a serial usb adapter this hit a panic.


lbeierlieb:
The mistake was hard to miss, but I managed to :sweat_smile: 

In debug builds, the calculation causes a panic.
```
thread 'tokio-rt-worker' (557872) panicked at src/device/xhci/endpoint_handle.rs:856:41:                                                                                      
attempt to subtract with overflow
```

A ran a release build, there is no check compiled in. I verified that a USB stick worked that previously ran into short packet `todo`. However, the kernel also had a warning present.
```
[    4.371011] xhci_hcd 0000:00:02.0: bad transfer trb length 16777028 in event trb
```

Thanks for fixing!

lbeierlieb (approved):

PR #292, opened 2026-07-27
Implementing file descriptor usage in addition to socket path

Noi0103:
This enables to start usbvfiod with the given file descriptors instead of using a socket path.

With this, usbvfiod implements the file deascriptor requirement as mentioned in the vfio-user specification. Previously this was a easy to reach `unreachable!()` statement.
(ref. https://www.qemu.org/docs/master/interop/vfio-user.html#backend-program-conventions)

It enables externally managing the sockets. This is done in the new nix integration test, using a systemd socket unit to create both the vfio-server and hotplug sockets. Additionally there are some docs included to hopefully avoid having to read the systemd manual when attempting to use the systemd socket units in the same fashion as the nix integration test does.

This will stay a Draft until we get the upstream vfio-user release that includes the feature to create a vfio-server with a filedescriptor. Until then there is a temporary commit to override the crate input to the source repositoy instead.

### DIFF
--- a/docs/users/systemd.md
+++ b/docs/users/systemd.md
@@ -1,0 +1,108 @@
+# Systemd setup with socket units
+
+**Why?**
+
+An upstream service will be started automatically when someone connects to the socket. For example, Cloud Hypervisor will start and be given a socket path managed by a systemd socket unit. When the Cloud Hypervisor process connects to the socket, systemd will start the usbvfiod service unit.
+
+When the Cloud Hypervisor recreates the VM due to a reboot, the socket connection will be broken.
+Upon breaking the vfio-user connection usbvfiod will exit and clean up it's remaining sockets.
+Cloud Hypervisor's VM will fail to restart due to the missing socket. Because Cloud Hypervisor is restarted on failure systemd will only then look at its dependencies and trigger a start for usbvfiod. After usbvfiod is restarted and has recreated the socket, the Cloud Hypervisor service can leave the restart on failure loop.
+If usbvfiod exited with a non zero exit code it will be blocked by preexisting files (not previously cleaned up sockets) that will not be overwritten. Because Cloud Hypervisor in the `--socket-path` variant depends on usbvfiod, it might be blocked infinitely due to that missing cleanup.
+
+Using systemd to manage the socket and providing it over the `--fd` argument decouples `usbvfiod.service` from its socket. It is not a direct dependency for the `cloud-hypervisor.service` anymore, the socket unit is. With usbvfiod's behavior to exit if the single vfio-user client disconnects it is now the only service to restart on a guest reboot.
+
+When depending on the socket unit, the file system path for the socketis guaranteed to exists when the socket unit succeeds. A dependence on the service itself does not give any guarantee the socket has been created before dependent services start. This may lead to spurious service startup failures.
+
+## Following the integration test example
+
+This section aims to provide a summary about using systemd socket units with usbvfiod and Cloud Hypervisor.
+
+The following systemd config snippets are created with a nix integration test and should help you get up and running. They include some noise in the `Service.Environment` fields but are kept as is to avoid untested examples. Values for `Socket.Group`, `Socket.Mode`, `Socket.User`, `Service.Group`, `Service.User` are not strict for the sake of simplicity in the integration test.
+
+The example will manage both of usbvfiods available socket options with systemd: the vfio-user server and the hotplug server.
+
+### Systemd Unit: usbvfiod.socket
+
+The Socket Unit's most important field is `Socket.ListenStream`. The order (if multiple fields exist as do in this example) will be used in the environment of the service when providing the file descriptors. The fields `Socket.SocketGroup`, `Socket.SocketMode` and `Socket.SocketUser` are for permission management of the resulting Sockets.
+
+Use the same name for this socket unit and the associated service to let systemd associate them automatically (equal to explicitly declaring `Socket.Service=usbvfiod.service`). Alternatively set `Socket.Service` explicitly.
+
+```ini
+[Unit]
+Description=sockets to trigger usbvfiod service start and provide communication channels
+
+[Socket]
+Accept=no
+ListenStream=/tmp/usbvfiod.sock
+ListenStream=/tmp/hotplug.sock
+SocketGroup=usbaccess
+SocketMode=660
+SocketUser=usbaccess
+
+[Install]
+WantedBy=sockets.target
+```
+
+### Systemd Unit: usbvfiod.service
+
+Following after the three default file descriptors `stdin`, `stdout` and `stderr` systemd provides the above defined order of `Socket.ListenStream` file descriptors. They are available in the `usbvfiod.service` configuration field `Service.ExecStart`, to be used as `--fd 3` (the vfio-user socket) and `--hotplug-fd 4` (the hotplug socket).
+
+```ini
+[Unit]
+
+[Service]
+Environment="LOCALE_ARCHIVE=/nix/store/54jg0kk0cn5h7j19r0l8x36fmbfywwjj-glibc-locales-2.42-67/lib/locale/locale-archive"
+Environment="PATH=/nix/store/di26b1kkbammy0sj70nq5qzvfrh78wxl-coreutils-9.11/bin:/nix/store/fcqbwp5hx5wh3lzf0457wmf0divknz7y-findutils-4.11.0/bin:/nix/store/aak8d9mrdv9sgn0lcg7xss7wxdg9sqh3-gnugrep-3.12/bin:/nix/store/7mfsmbhsd3arnypipwm251rcd0b45riy-gnused-4.10/bin:/nix/store/r6sz8p6sd6c73fp9z8nzl04dri7lyx8n-systemd-261.1/bin:/nix/store/di26b1kkbammy0sj70nq5qzvfrh78wxl-coreutils-9.11/sbin:/nix/store/fcqbwp5hx5wh3lzf0457wmf0divknz7y-findutils-4.11.0/sbin:/nix/store/aak8d9mrdv9sgn0lcg7xss7wxdg9sqh3-gnugrep-3.12/sbin:/nix/store/7mfsmbhsd3arnypipwm251rcd0b45riy-gnused-4.10/sbin:/nix/store/r6sz8p6sd6c73fp9z8nzl04dri7lyx8n-systemd-261.1/sbin"
+Environment="RUST_BACKTRACE=full"
+Environment="TZDIR=/nix/store/2nndxyf3phkb5aggxm3c16sapa0f49kz-tzdata-2026c/share/zoneinfo"
+ExecStart=/nix/store/lcl5ngmqmqk7idgrw83ghag59sw7kz9y-usbvfiod-0.2.0/bin/usbvfiod -v --fd 3 --hotplug-fd 4 --device "/dev/bus/usb/testdevice"
+
+Group=usbaccess
+Restart=on-failure
+RestartSec=2s
+User=usbaccess
+```
+
+### Systemd Unit: cloud-hypervisor.service
+
+The `cloud-hypervisor.service` declares a dependency on `usbvfiod.socket` (via `Unit.Requires`) to ensure the path in the filesystem is created before starting Cloud Hypervisor. A dependency directly on `usbvfiod.service` is not necessary.
+
+When starting Cloud Hypervisor, the vfio-user socket path is provided in the same way as it is without a systemd socket unit (see the `Service.ExecStart` field below). In this example the `--user-device` receives the first of the two defined `Socket.ListenStream` paths, because the `usbvfiod.service` configuration above used file descriptor 3 for the vfio-user server.
+
+When the `Install.WantedBy` starts Cloud Hypervisor, it connects to the socket path provided with the `--user-device` argument. Establishing a connection to a socket unit defined path will prompt systemd to start the service unit behind the socket and pass on all communication.
+
+```ini
+[Unit]
+Requires=usbvfiod.socket
+
+[Service]
+Environment="LOCALE_ARCHIVE=/nix/store/54jg0kk0cn5h7j19r0l8x36fmbfywwjj-glibc-locales-2.42-67/lib/locale/locale-archive"
+Environment="PATH=/nix/store/di26b1kkbammy0sj70nq5qzvfrh78wxl-coreutils-9.11/bin:/nix/store/fcqbwp5hx5wh3lzf0457wmf0divknz7y-findutils-4.11.0/bin:/nix/store/aak8d9mrdv9sgn0lcg7xss7wxdg9sqh3-gnugrep-3.12/bin:/nix/store/7mfsmbhsd3arnypipwm251rcd0b45riy-gnused-4.10/bin:/nix/store/r6sz8p6sd6c73fp9z8nzl04dri7lyx8n-systemd-261.1/bin:/nix/store/di26b1kkbammy0sj70nq5qzvfrh78wxl-coreutils-9.11/sbin:/nix/store/fcqbwp5hx5wh3lzf0457wmf0divknz7y-findutils-4.11.0/sbin:/nix/store/aak8d9mrdv9sgn0lcg7xss7wxdg9sqh3-gnugrep-3.12/sbin:/nix/store/7mfsmbhsd3arnypipwm251rcd0b45riy-gnused-4.10/sbin:/nix/store/r6sz8p6sd6c73fp9z8nzl04dri7lyx8n-systemd-261.1/sbin"
+Environment="TZDIR=/nix/store/2nndxyf3phkb5aggxm3c16sapa0f49kz-tzdata-2026c/share/zoneinfo"
+ExecStart=/nix/store/kxnddbaypz5sic0dgqx60lp4k9z662yg-cloud-hypervisor-53.0/bin/cloud-hypervisor --memory size=2G,shared=on --console file=/tmp/console.log --serial off \
+  --kernel /nix/store/nmyhzfvkn5zagl7lc16w9m6d54v4696d-linux-6.18.41/bzImage \
+  --cmdline 'init=/nix/store/wgjjv8axl1rlfk277h0p4al3x1jn1w95-nixos-system-nixos-kexec-26.11.20260803.104240a/init usbcore.autosuspend=-1 console=hvc0 console=tty0 xhci_pci.dyndbg==pmfl xhci_hcd.dyndbg==pmfl nohibernate root=fstab loglevel=8 lsm=landlock,yama,bpf' \
+  --initramfs /nix/store/swc8c4c3kv1w21mlzcrw7kfpj64pq515-initrd/initrd \
+  --user-device socket=/tmp/usbvfiod.sock \
+  --net "tap=tap0,mac=,ip=192.168.100.1,mask=255.255.255.0"
+
+Restart=on-failure
+RestartSec=2s
+
+[Install]
+WantedBy=multi-user.target
+```
+
+### Using hotplug events to start usbvfiod
+
+In these example configurations the hotplug socket is also managed through systemd. Udev rules trigger commands for managing device attachment state using the hotplug socket. Doing this triggers systemd to start the `usbvfiod.service` independent of Cloud Hypervisor service status.
+
+In this scenario either one can start the `usbvfiod.service`: a hotplug event or a starting Cloud Hypervisor.
+
+
+Further reading:
+- https://www.freedesktop.org/software/systemd/man/latest/systemd.socket.html
+- https://www.freedesktop.org/software/systemd/man/latest/systemd.exec.html
+- https://www.freedesktop.org/software/systemd/man/latest/systemd.unit.html
+- https://www.freedesktop.org/software/systemd/man/latest/systemd.service.html
+- https://www.freedesktop.org/software/systemd/man/latest/sd_listen_fds.html
--- a/nix/checks/filedescriptor.nix
+++ b/nix/checks/filedescriptor.nix
@@ -1,0 +1,34 @@
+{
+  testutils,
+}:
+testutils.mkUsbTest {
+  debug = true;
+  name = "systemd-managed-filedescriptor";
+  useFileDescriptor = true;
+  virtualDevices = [
+    {
+      type = "blockdevice";
+      usbVersion = "3";
+    }
+  ];
+  testScript = ''
+    out = cloud_hypervisor.succeed("lsusb", timeout=60)
+    print(out)
+    search("ID ${testutils.blockdeviceVendorId}:${testutils.blockdeviceProductId} QEMU QEMU USB HARDDRIVE", out)
+
+    # A nested guest reboot will break the vfio-user connection and usbvfiod will exit 0.
+    # When Cloud Hypervisor re-connects to the systemd socket usbvfiod will be restarted.
+    out = cloud_hypervisor.succeed("systemctl reboot", timeout=60)
+    print(out)
+
+    # confirm nested guest shutdown via usbvfiod exit message triggered by the closed vfio-user connection
+    machine.wait_until_succeeds("journalctl -b -u usbvfiod.service | grep 'Deactivated successfully.' ", timeout=60)
+
+    out = cloud_hypervisor.wait_until_succeeds("lsusb -d ${testutils.blockdeviceVendorId}:${testutils.blockdeviceProductId}", timeout=240)
+
+    # When using the socket path setup, we encounter failed to start logs for the cloud hypervisor service.
+    # We do not expect to see this log when using the file descriptor setup.
+    out = machine.fail("journalctl -b -u cloud-hypervisor.service | grep -q 'cloud-hypervisor.service: Failed with result'", timeout=60)
+    print(out)
+  '';
+}
--- a/nix/checks/testutils.nix
+++ b/nix/checks/testutils.nix
@@ -110,8 +110,8 @@ let
   # Putting the socket in a world-readable location is obviously not a
   # good choice for a production setup, but for this test it works
   # well.
-  usbvfiodSocket = "/tmp/usbvfio";
-  usbvfiodSocketHotplug = "/tmp/hotplug";
+  usbvfiodSocket = "/tmp/usbvfiod.sock";
+  usbvfiodSocketHotplug = "/tmp/hotplug.sock";
 
   guestLogFile = "/tmp/console.log";
   qemuLogFile = "/tmp/qemu-vc.log";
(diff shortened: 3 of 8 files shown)


nix/checks/fd-test.nix
```diff
@@ -0,0 +1,28 @@
...
+  testScript = ''
+    out = cloud_hypervisor.succeed("lsusb", timeout=60)
+    print(out)
+    search("ID ${testutils.blockdeviceVendorId}:${testutils.blockdeviceProductId} QEMU QEMU USB HARDDRIVE", out)
+
+    out = cloud_hypervisor.succeed("systemctl reboot", timeout=60)
+    print(out)
+
+    # confirm nested guest shutdown via usbvfiod exit message triggered by the closed vfio-user connection
+    machine.wait_until_succeeds("journalctl -u usbvfiod.service | grep 'Deactivated successfully.' ", timeout=60)
```
snue: nit: this should only ask for the logs of the current boot (`-b`). Only an issue with the driver and `--keep-vm-state` outside the sandbox.

nix/checks/filedescriptor.nix
```diff
@@ -0,0 +1,28 @@
...
+      type = "blockdevice";
+      usbVersion = "3";
+    }
+  ];
+  testScript = ''
+    out = cloud_hypervisor.succeed("lsusb", timeout=60)
+    print(out)
+    search("ID ${testutils.blockdeviceVendorId}:${testutils.blockdeviceProductId} QEMU QEMU USB HARDDRIVE", out)
+
+    out = cloud_hypervisor.succeed("systemctl reboot", timeout=60)
```
snue, line 21: This could have a comment explaining what the test does. (Successful guest reboot and restart usbvfiod through the systemd socket units.)

nix/checks/fd-test.nix
```diff
@@ -0,0 +1,28 @@
+{
+
```
snue: nit: remove the blank line.

nix/checks/testutils.nix
```diff
@@ -314,9 +314,68 @@ let
...
+          "usbvfiod" = {
+            wantedBy = [ ];
+            requires = [ ];
+            serviceConfig = {
+              User = "usbaccess";
+              Group = "usbaccess";
+              Restart = "on-failure";
+              RestartSec = "2s";
+              ExecStart = ''
+                ${lib.getExe usbvfiod} ${if debug then "-v" else ""} --fd 3 --hotplug-fd 4 ${lib.concatStringsSep " " (builtins.map mkDeviceFlag virtualDevices)}
```
snue: There is quite some redundancy here. How about a `mkSocketFlag` that produces the `--fd` or `--path` strings? And then concatenate only the `sockets` unit on demand.

src/cli.rs
snue: In hindsight, this should have been `#[expect(dead_code)]`. Now it should be removed.

src/cli.rs
```diff
@@ -72,10 +75,15 @@ pub enum ServerSocket<'a> {
 }
 
 impl Cli {
+    #[allow(clippy::option_if_let_else)]
```
snue: What does it suggest? Why don't you want it?

src/main.rs
```diff
@@ -73,6 +84,16 @@ fn main() -> Result<()> {
...
+    if let Some(hotplug_fd) = args.hotplug_fd {
+        // SAFETY: we have to assume the given fd is valid, there is not much else we can do
+        let owned_fd_hotplug: OwnedFd = unsafe { OwnedFd::from_raw_fd(hotplug_fd) };
+        let hotplug_control = backend.hotplug_control();
+        let socket = UnixListener::from(owned_fd_hotplug);
+        thread::Builder::new()
+            .name("hot-attach-socket listener".to_string())
+            .spawn(move || run_hotplug_server(socket, hotplug_control, runtime.clone()))
+            .unwrap();
+    }
```
snue: How about getting a `hotplug_socket: Option<UnixListener>` first, and then spawning the thread if it is `Some` to get rid of the duplication? You could add an `args.hotplug_socket()` analogous to `args.server_socket()`.

src/cli.rs
```diff
@@ -47,9 +47,12 @@ pub struct Cli {
 
     /// The path where to create a listening Unix domain socket and listen
     /// for hotplug commands.
-    #[arg(long, value_name = "PATH")]
+    #[arg(long, value_name = "PATH", conflicts_with = "hotplug_fd")]
     pub hotplug_socket_path: Option<PathBuf>,
 
+    #[arg(long, conflicts_with = "hotplug_socket_path")]
```
snue: I think you strictly only need to declare the conflict on one of them.

src/cli.rs
Noi0103: I found clippys example much harder to read here:
`error: use Option::map_or_else instead of an if let/else`
```rust
self.fd.map_or_else(|| if let Some(socket_path) = &self.socket_path {
      ServerSocket::Path(socket_path)
} else {
     // The clap configuration above prevents that we run into this case.
    unreachable!()
}, |fd| ServerSocket::Fd(fd))
```

nix/checks/testutils.nix
```diff
@@ -268,55 +268,77 @@ let
...
-        "useFileDescriptor == true is not yet implemented for the usbvfiod service setup"
-        { };
+      };
+
+      systemd.sockets = lib.mkIf useFileDescriptor {
+        usbvfiod = {
+          description = "vfio-user socket";
+          wantedBy = [ "sockets.target" ];
+          socketConfig = {
+            # if this ever makes problems try ListenSequentialPacket instead
```
snue: What kinds of problems would you expect here? And have you tried whether `SOCK_SEQPACKET` works transparently in practice?

src/cli.rs
```diff
@@ -73,9 +78,35 @@ pub enum ServerSocket<'a> {
...
+    pub fn hotplug_socket(&self) -> Option<UnixListener> {
+        // The clap configuration above ensures that maximum of one of the two hotplug options is used.
+        self.hotplug_fd.map_or_else(
+            || {
+                self.hotplug_socket_path.as_ref().map_or_else(
+                    || None,
+                    |path| {
+                        Some(
+                            UnixListener::bind(path.as_path())
+                                .expect("failed to use provided hotplug socket path"),
```
snue, line 100: The `bind` can fail for reasons like the path already exists. Can we provide a decent error message here? (or what would a failure due to this `expect` currently look like?)

src/main.rs
```diff
@@ -57,20 +60,23 @@ fn main() -> Result<()> {
...
+            Server::from_owned_fd(owned_fd, true, backend.irqs(), backend.regions())
+        }
     };
 
-    // listen on socket for hot-attach fds
-    if let Some(hotplug_socket_path) = args.hotplug_socket_path.clone() {
-        let hotplug_control = backend.hotplug_control();
-        let socket = UnixListener::bind(hotplug_socket_path.as_path()).unwrap();
+    let hotplug_control = backend.hotplug_control();
+    if let Some(listener) = args.hotplug_socket() {
```
snue: `hotplug_control` is only needed if there is a hotplug socket.

```suggestion
    if let Some(socket) = args.hotplug_socket() {
        let hotplug_control = backend.hotplug_control();
```

(I don't really care if you call it `socket` or rename to `listener`. The former just gives a smaller diff 😛 )

Cargo.toml
```diff
@@ -45,8 +45,10 @@ tracing-subscriber = { version = "0.3.19", default-features = false, features =
   "ansi",
   "tracing-log",
 ] }
-vfio-bindings = { version = "0.6.0", default-features = false }
-vfio_user = { version = "0.1.1" }
+#vfio-bindings = { version = "0.6.0", default-features = false }
+#vfio_user = { version = "0.1.1" }
```
snue: nit: we don't need to keep the commented version around. That info is always in the git history if needed.

src/cli.rs
Noi0103: I think the current should suffice:
```
thread 'main' (146298) panicked at src/cli.rs:100:34:
failed to use provided hotplug socket path: Os { code: 98, kind: AddrInUse, message: "Address already in use" }
```

Cargo.toml
Noi0103: I left it there because of the `default-feature = false`. I do not think I would remember to add this once the override can be removed.

nix/checks/testutils.nix
Noi0103: I was not sure how the difference is important and left this just in case, because so far there is no issue with the `ListenStream`, we probably don't need the comment.

Cargo.toml
snue: But you still have the same setting on the override? And it should probably be visible in `Cargo.lock` if you did something wrong there (pulling in many new dependencies).

Noi0103: Ah, right I do, my bad. I never intended to merge this commit but in case we want to I'll remove the comment now.

nix/checks/testutils.nix
```diff
@@ -268,55 +268,76 @@ let
...
-            environment = {
-              RUST_BACKTRACE = "full";
-            };
+    let
+      netboot = mkNetboot debug;
+    in
+    {
+      systemd.services = {
+        usbvfiod = {
+          wantedBy = if useFileDescriptor == false then [ "multi-user.target" ] else [ ];
```
snue: nit: this shouldn't need if/then/else when the else is empty. Have not checked whether `mkIf` works for lists, though.

```suggestion
          wantedBy = lib.mkIf !useFileDescriptor [ "multi-user.target" ];
```

nix/checks/testutils.nix
```diff
@@ -268,55 +268,76 @@ let
...
-              ExecStart = ''
-                ${lib.getExe usbvfiod} ${if debug then "-v" else ""} --socket-path ${usbvfiodSocket} --hotplug-socket-path ${usbvfiodSocketHotplug} ${lib.concatStringsSep " " (builtins.map mkDeviceFlag virtualDevices)}
-              '';
-            };
-            environment = {
-              RUST_BACKTRACE = "full";
-            };
+    let
+      netboot = mkNetboot debug;
+    in
```
snue: This is only needed on the cloud-hypervisor service definition. I'd prefer if it stays there.

nix/checks/testutils.nix
```diff
@@ -268,55 +268,76 @@ let
     ACTION=="add|change|bind", ATTRS{serial}=="0000:00:${pciAddr}.0", SUBSYSTEM=="usb", ATTRS{product}=="${controller}", ATTR{devpath}=="${port}", MODE="0660", GROUP="usbaccess", SYMLINK+="bus/usb/${symlink}"
   '';
 
+  # Create the partial usbvfiod argument string for either fs paths or fd ids
+  mkSocketPath =
```
snue: nit: there is `mkDeviceFlag` already. This should be called `mkSocketFlag`.

nix/checks/testutils.nix
```diff
@@ -268,55 +268,76 @@ let
...
-      }
-    else
-      lib.assertMsg false
-        "useFileDescriptor == true is not yet implemented for the usbvfiod service setup"
-        { };
+      };
+
+      systemd.sockets = lib.mkIf useFileDescriptor {
+        usbvfiod = {
+          description = "vfio-user socket";
```
snue: This is the socket unit for the USB pass-through Controller (usbvfiod). The vfio-user socket is just one of the sockets it provides.

docs/users/systemd.md
```diff
@@ -0,0 +1,99 @@
...
+SocketMode=660
+SocketUser=usbaccess
+
+[Install]
+WantedBy=sockets.target
+```
+
+#### Systemd Unit: usbvfiod.service
+
+The services field `Unit.Requires` does not need to point to the above defined `usbvfiod.socket` as long as they share the same name. Systemd will associate them by default and explicitly setting `Unit.Requires=usbvfiod.socket` will have no effect. If the names diverge, set this field accordingly.
```
snue: Not exactly how this works? Access on the sockets provided by the above unit needs to launch this service (with the additional opened fds). This is automatic when the name matches. But adding a requirement on the socket unit for a differently named service will not make systemd launch that service on socket access instead. Such a requirement just ensures any socket paths provided by the unit will exist before the service is launched. And when the socket unit launches this service, no such requirement exists (but it is also trivially fulfilled).

You need [`Service=` on the Socket unit](https://www.freedesktop.org/software/systemd/man/latest/systemd.socket.html?#Service=) for that socket access to service launch connection.

docs/users/systemd.md
```diff
@@ -0,0 +1,99 @@
...
+Following after the three default file descriptors `stdin`, `stdout` and `stderr` systemd provides the above defined order of file descriptors and we pass them in the `Service.ExecStart` field as `--fd 3` for the vfio-user socket and `--hotplug-fd 4` for the hotplug socket.
+Using the fields `Socket.User`and `Socket.Group` the service has sufficient permissions to use the socket.
```
snue: I don't think any permissions are needed here, The service gets the opened socket at launch from systemd. Only cloud-hypervisor needs access to the path provided through the socket unit, e.g. through matching user/group.

In the CI, the service also directly opens a USB device, which is why it runs as a fitting user. The socket unit and cloud-hypervisor service could share a different UID/GID when the socket unit based setup is in use. We share the same user for all services for simplicity.

docs/users/systemd.md
```diff
@@ -0,0 +1,99 @@
...
+
+Group=usbaccess
+Restart=on-failure
+RestartSec=2s
+User=usbaccess
+```
+
+#### Systemd Unit: cloud-hypervisor.service
+
+The cloud-hypervisor unit does not share the name with the above socket unit and the field `Unit.Requires` needs to declare the dependence on `usbvfiod.socket`.
```
snue: See above about the name-sharing. This service needs the filesystem path to the socket to exist. Either by requiring `usbvfiod.service` if that creates the socket, or the socket unit instead. And it needs to have the permissions to open this socket.

docs/users/systemd.md
```diff
@@ -0,0 +1,99 @@
...
+When starting Cloud Hypervisor we simply provide the socket path declared in the socket units `Socket.ListenStream` field as `--user-device` value.
```
snue: Needs to be a bit more specific, as we have two `ListenStream`s in the unit. We are talking about the vfio-user socket path.

docs/users/systemd.md
```diff
@@ -0,0 +1,99 @@
...
+When this services `Install.WantedBy` starts Cloud Hypervisor the Guest OS is expected to attempt communicating with the `--user-device`. This socket communication will prompt systemd to start the `usbvfiod.service` unit and pass on all communication that happened so far.
```
snue: Cloud-Hypervisor opens the socket and establishes the connection way before the guest runs. You can see it fail very early when the socket isn't available.

docs/users/systemd.md
```diff
@@ -0,0 +1,99 @@
+# Systemd setup
+
+## Using systemd socket units
+
+**Why?**
+
+A upstream service will be started automatically when the socket is receiving input. For example, Cloud Hypervisor will start and be given a file descriptor managed by systemd. Only when Cloud Hypervisor is sending over this file descriptor systemd will start the Usbvfiod service unit.
```
snue: Isn't the service launched when the socket is accessed/opened? Does it require explicit data transfer?

Also, Cloud-Hypervisor will not receive a file-descriptor. It accesses the socket by path.

docs/users/systemd.md
```diff
@@ -0,0 +1,99 @@
...
+
+When the Guest OS in Cloud Hypervisor reboots the vfio-user connection will be broken and the Usbvfiod service will have exited. The socket created with the `--socket-path` argument will not be available anymore when the Guest OS is starting again. The attempt to enumerate the vfio-user pci device will then fail. This issue disappears when the file descriptor is managed externally and given to the services instead of the service creating it itself.
```
snue: The Guest OS is not really involved here. The Guest reboot destroys and recreates the VM in Cloud Hypervisor. The same can probably be achieved through the `ch-remote` API. Cloud-Hypervisor simply tries to open the vfio-user device again to add it to the emulated PCI bus. It fails before the Guest runs and tries to enumerate PCI devices.

snue (commented):
Looking better with the removed duplication. Thanks. I took a first look at the added documentation as well.

docs/users/systemd.md
snue: Requirements on just the service are also slightly racy, as systemd considers the service launched before usbvfiod actually creates the socket. Socket units solve this as well.

docs/users/systemd.md
snue: The systemd socket unit will keep the socket path available and relaunch the usbvfiod service as necessary.

src/cli.rs
```diff
@@ -47,9 +50,12 @@ pub struct Cli {
 
     /// The path where to create a listening Unix domain socket and listen
     /// for hotplug commands.
-    #[arg(long, value_name = "PATH")]
+    #[arg(long, value_name = "PATH", conflicts_with = "hotplug_fd")]
     pub hotplug_socket_path: Option<PathBuf>,
 
+    #[arg(long)]
+    pub hotplug_fd: Option<RawFd>,
```
snue, line 57: nit: you could have added the conflict annotation only on the new option. Also, the `value_name` for the CLI help is usually called `FDNUM` (file descriptor number). I haven't checked how the `RawFd` shows up in the help (and we currently don't explicitly set it for `--fd` either).

nix/checks/testutils.nix
```diff
@@ -110,7 +110,7 @@ let
   # Putting the socket in a world-readable location is obviously not a
   # good choice for a production setup, but for this test it works
   # well.
-  usbvfiodSocket = "/tmp/usbvfio";
+  usbvfiodSocket = "/tmp/usbvfiod";
   usbvfiodSocketHotplug = "/tmp/hotplug";
```
snue: nit: I think you had added `.sock` suffixes at some point.

nix/checks/filedescriptor.nix
```diff
@@ -0,0 +1,29 @@
...
+      usbVersion = "3";
+    }
+  ];
+  testScript = ''
+    out = cloud_hypervisor.succeed("lsusb", timeout=60)
+    print(out)
+    search("ID ${testutils.blockdeviceVendorId}:${testutils.blockdeviceProductId} QEMU QEMU USB HARDDRIVE", out)
+
+    # A nested guest reboot will trigger a usbvfiod service restart by breaking vfio-user connection so usbvfiod will exit 0.
+    # The underlying socket connection to cloud hypervisor will not be broken by using systemd sockets.
```
snue: Oh, the vfio socket connection is closed and reopened by Cloud Hypervisor during VM reboot. The clean exit of usbvfiod will not trigger any service restart (when not using explicit `Restart=always`). We just don't remove and recreate the socket this way, and let systemd restart the usbvfiod service on reconnect.

It also works for the guest reboot with the simple service dependency. In this case, usbvfiod exits cleanly and removes the stale sockets. Cloud hypervisor then fails to find the socket to reattach when recreating the VM and crashes. As it is set to restart on-failure, systemd restarts it and the stopped/finished usbvfiod service it requires. Cloud hypervisor may fail again if the socket is not recreated fast enough, but succeeds eventually. This leaves ugly service start failure messages in the logs, but it does work.

Does this test fail reliably without the socket unit?

nix/checks/testutils.nix
```diff
@@ -268,55 +268,75 @@ let
...
+          SocketGroup = "usbaccess";
+          Accept = "no";
         };
-      }
-    else
-      lib.assertMsg false
-        "useFileDescriptor == true is not yet implemented for the usbvfiod service setup"
-        { };
+        unitConfig = {
+        };
```
snue: nit: Do you need it if it is empty? Does the NixOS module expect this to be set?

docs/users/systemd.md
```diff
@@ -0,0 +1,107 @@
...
+
+When the Cloud Hypervisor recreates the VM due to a reboot the socket connection will be broken.
+Upon breaking the vfio-user connection usbvfiod will exit and the `--socket-path` be removed.
+Until usbvfiod is restarted and has recreated the socket, Cloud Hypervisor will fail to restart due to the missing socket. This will work for a graceful shutdown where usbvfiod can clean up sockets before completely exiting and being restarted.
+In any other case the restart of usbvfiod will be blocked by preexisting files (not previously cleaned sockets) that will not be overwritten. Because Cloud Hypervisor in the `--socket-path` variant depends on usbvfiod it might be blocked infinitely due to that missing cleanup.
+
+Using systemd to manage the socket and providing it over the `--fd` argument decouples the upstream `usbvfiod.service` from its socket. It is not a direct dependency for the `cloud-hypervisor.service` anymore, the socket unit is. With usbvfiod's behavior to exit if the single vfio-user client disconnects both services will still restart on a guest reboot.
```
snue: I would avoid 'upstream' here. And not "both services" restart. Only usbvfiod restarts. Cloud hypervisor keeps running. It just re-establishes the socket connection when it recreates the VM.

docs/users/systemd.md
```diff
@@ -0,0 +1,107 @@
+# Systemd setup
+
+## Using systemd socket units
```
snue: nit: section headings only make sense when there are multiple. Do you plan on adding more sections to this document? A single `# H1` is fine, but if you use `## H2` and lower, there should be more than one.

docs/users/systemd.md
```diff
@@ -0,0 +1,107 @@
...
+  --net "tap=tap0,mac=,ip=192.168.100.1,mask=255.255.255.0"
+
+Restart=on-failure
+RestartSec=2s
+
+[Install]
+WantedBy=multi-user.target
+```
+
+In these example configurations the hotplug service is also managed through systemd. Udev rules could trigger oneshot services to attempt hotplug commands that for example attach a device. Doing this would trigger systemd to start the `usbvfiod.service` before the Cloud Hypervisor service is started.
```
snue: The "could" and "would" is too weak. Just remove those and the sentences already read better. It's a precise description of the example configuration after all. You can also add the device unit template example and udev rule here that invokes `remote`.

```suggestion
In these example configurations the hotplug socket is also managed through systemd. Udev rules trigger commands for managing device attachment state through the hotplug socket. Doing this triggers systemd to start the `usbvfiod.service` independent of the Cloud Hypervisor service status.
```

I see the original intention was to describe a direct udev command invocation without the intermediate device service. You can add such an example as well and then refine that to use a device service unit.

docs/users/systemd.md
```diff
@@ -0,0 +1,107 @@
...
+In any other case the restart of usbvfiod will be blocked by preexisting files (not previously cleaned sockets) that will not be overwritten. Because Cloud Hypervisor in the `--socket-path` variant depends on usbvfiod it might be blocked infinitely due to that missing cleanup.
+
+Using systemd to manage the socket and providing it over the `--fd` argument decouples the upstream `usbvfiod.service` from its socket. It is not a direct dependency for the `cloud-hypervisor.service` anymore, the socket unit is. With usbvfiod's behavior to exit if the single vfio-user client disconnects both services will still restart on a guest reboot.
+
+When depending on the socket unit there is a definitive moment when the file system path for the socket exists (when the socket unit succeeded). This removes the uncertainty that a simple dependence on the service itself does not give any indication when the service has created the socket. In an extreme case the downstream service could have started and exited too early and so often, it might remain in a failed status.
```
snue: We can shorten this.

```suggestion
When depending on the socket unit, the file system path for the socket is guaranteed to exist when the unit succeeds. A dependence on the service itself does not give any guarantee the socket has been created before dependent services start. This may lead to spurious service startup failures.
```

docs/users/systemd.md
```diff
@@ -0,0 +1,107 @@
...
+User=usbaccess
+```
+
+#### Systemd Unit: cloud-hypervisor.service
+
+The `cloud-hypervisor.service` needs to declare a dependence on `usbvfiod.socket` (via `Unit.Requires`) to ensure the path in the filesystem is created before starting Cloud Hypervisor. A dependence directly on `usbvfiod.service` is not necessary.
+
+When starting Cloud Hypervisor, the vfio-user socket path is provided in the same way as it is without a systemd socket unit (see the `Service.ExecStart` field below). In this example the `--user-device` receives the first of the two defined `Socket.ListenStream` paths, because the `usbvfiod.service` configuration above used file descriptor 3 for the vfio-user server.
+
+When the `Install.WantedBy` starts Cloud Hypervisor, it is expected to connect to the socket path provided with the `--user-device` argument. Anyone establishing a connection to a socket unit defined path will prompt systemd to start the service unit behind the socket and pass on all communication.
```
snue: We can shorten this as well.


```suggestion
When the `Install.WantedBy` starts Cloud Hypervisor, it connects to the socket path provided with the `--user-device` argument. Establishing a connection to a socket unit defined path will prompt systemd to start the service unit behind the socket and pass on all communication.
```

docs/users/systemd.md
```diff
@@ -0,0 +1,107 @@
...
+
+Group=usbaccess
+Restart=on-failure
+RestartSec=2s
+User=usbaccess
+```
+
+#### Systemd Unit: cloud-hypervisor.service
+
+The `cloud-hypervisor.service` needs to declare a dependence on `usbvfiod.socket` (via `Unit.Requires`) to ensure the path in the filesystem is created before starting Cloud Hypervisor. A dependence directly on `usbvfiod.service` is not necessary.
```
snue:
```suggestion
The `cloud-hypervisor.service` declares a dependency on `usbvfiod.socket` (via `Unit.Requires`) to ensure the path in the filesystem exists before starting Cloud Hypervisor. A dependency directly on `usbvfiod.service` is not necessary.
```

snue (commented):
Yay, the feature is released. Thanks for adding documentation with it. I left a bunch of suggestions.

src/cli.rs
Noi0103: The variable name is used, it shows up in the help as `--fd <FD>`.

docs/users/systemd.md
Noi0103: I don't plan to. It was a "holding a door open", but this can be changed back into separate headlines when it becomes necessary.

nix/checks/filedescriptor.nix
```diff
@@ -0,0 +1,34 @@
...
+  ];
+  testScript = ''
+    out = cloud_hypervisor.succeed("lsusb", timeout=60)
+    print(out)
+    search("ID ${testutils.blockdeviceVendorId}:${testutils.blockdeviceProductId} QEMU QEMU USB HARDDRIVE", out)
+
+    # When using the socket path setup we encounter failed to start logs for the cloud hypervisor service.
+    # We do not expect to see this log when using the file descriptor setup.
+    out = machine.fail("journalctl -b -u cloud-hypervisor.service | grep -q 'cloud-hypervisor.service: Failed with result'", timeout=60)
+    print(out)
```
snue: But only after the guest reboot? It may also happen before occasionally with the service, but on reboot we are guaranteed to see at least one failure with the service dependency instead of the sockets.

docs/users/systemd.md
```diff
@@ -0,0 +1,108 @@
...
+[Socket]
+Accept=no
+ListenStream=/tmp/usbvfiod
+ListenStream=/tmp/hotplug
+SocketGroup=usbaccess
+SocketMode=660
+SocketUser=usbaccess
+
+[Install]
+WantedBy=sockets.target
```
snue, line 43: I think you need to update the examples. At least the socket paths.

docs/users/systemd.md
```diff
@@ -0,0 +1,108 @@
...
+
+When depending on the socket unit, the file system path for the socketis guaranteed to exists when the socket unit succeeds. A dependence on the service itself does not give any guarantee the socket has been created before dependent services start. This may lead to spurious service startup failures.
+
+## Following the integration test example
+
+This section aims to provide enough information about the usage of socket units with usbvfiod and Cloud Hypervisor to avoid searching for the relevant bits of information in the systemd manual.
+
+The following systemd config snippets are created with a nix integration test and should help you get up and running. They include some noise in the `Service.Environment` fields but are kept as is to avoid untested examples. Values for `Socket.Group`, `Socket.Mode`, `Socket.User`, `Service.Group`, `Service.User` are not strict for the sake of simplicity in the integration test.
```
snue, line 20: I get the desire to take examples from working configurations. But they are not working for copy&paste anyway. Let's rather omit the noisy environment, and maybe even replace the binary path with a generic `/path/to/usbvfiod`. That should make it easier to follow for people handcrafting service units.

A reader will also not quite understand what "Values for `Socket.Group`, `Socket.Mode`, `Socket.User`, `Service.Group`, `Service.User` are not strict for the sake of simplicity in the integration test." means. We should probably be specific about the outcome of these service settings and what we omitted.

docs/users/systemd.md
```diff
@@ -0,0 +1,108 @@
...
+If usbvfiod exited with a non zero exit code it will be blocked by preexisting files (not previously cleaned up sockets) that will not be overwritten. Because Cloud Hypervisor in the `--socket-path` variant depends on usbvfiod, it might be blocked infinitely due to that missing cleanup.
+
+Using systemd to manage the socket and providing it over the `--fd` argument decouples `usbvfiod.service` from its socket. It is not a direct dependency for the `cloud-hypervisor.service` anymore, the socket unit is. With usbvfiod's behavior to exit if the single vfio-user client disconnects it is now the only service to restart on a guest reboot.
+
+When depending on the socket unit, the file system path for the socketis guaranteed to exists when the socket unit succeeds. A dependence on the service itself does not give any guarantee the socket has been created before dependent services start. This may lead to spurious service startup failures.
+
+## Following the integration test example
+
+This section aims to provide enough information about the usage of socket units with usbvfiod and Cloud Hypervisor to avoid searching for the relevant bits of information in the systemd manual.
```
snue: nit: So you dislike the systemd manual pages? I'd rather avoid that/reword it to sound less harsh.

docs/users/systemd.md
```diff
@@ -0,0 +1,108 @@
+# Systemd setup with socket units
+
+**Why?**
+
+An upstream service will be started automatically when someone connects to the socket. For example, Cloud Hypervisor will start and be given a socket path managed by a systemd socket unit. When the Cloud Hypervisor process connects to the socket, systemd will start the usbvfiod service unit.
+
```
snue, line 6: I think it makes sense to describe the systemd services and the socket unit separately.

snue (approved):
The code changes look good. The documentation can still be improved, but we can make a mental note to do that later and merge what we got so far.

The "no failures in the log" check in the new test needs to be moved. I am already leaving an approval, though.

nix/checks/filedescriptor.nix
Noi0103: Comparing the results of `useFileDescriptor = false;` and `useFileDescriptor = true;` this check worked fine locally to allow only the Fd variant.
I would rather add this check a second time at the end of the script instead of moving the existing one.

docs/users/systemd.md
Noi0103: > So you dislike the systemd manual pages?

Not quite, the idea was to invite the reader to avoid 5 manual pages with experimenting in favor of one 2 pages with an explained example. Maybe that is more or less what should be written there instead to avoid confusion.

nix/checks/filedescriptor.nix
snue: I think this check is still racy for `useFileDescriptor = false` (it can succeed when cloud-hypervisor starts sufficiently late). But we don't run it that way in CI, so there will not be flakiness. I think the current placement with the comment is still misleading. I'm fine with adding the check again after the reboot, but just moving it looks sufficient to me.

Noi0103: The comment being a bit misleading I did not yet consider. I will move the check down like said.

Noi0103:
The left open touch ups on docs are:
- describing the socket unit and service unit separately
- going more in depth about the permission setup with users and groups.

PR #293, opened 2026-07-30
Various fixes to enable usb to serial adapter

Noi0103:
After adding a simple immediate data implementation in #285, using a USB to serial adapter is possible.

There are some smaller fixes included in this PR and a new nix integration test using a qemu emulated usb-serial device.

For the qemu chardevice I tried to use a ringbuf first but that is just one-directional. Using a socket without any extra option or with server=on,wait=off did not work. To avoid blindly testing I decided to use named pipes because they rely less on qemu documentation.

---

Things this PR does (because it grew and has not been split up while doing so):
- fix to ensure cancel safety for `next_completion()`
- fix for events sent in case of a short transfer
- prevent panic due to the assertion that ensures always calling `submit_trb()` with following `next_completion()` before another `submit_trb()` is possible
- fix missing reset for `EndpointHandle.submission_state` when using `cancel()` on a `TdBasedInEndpointHandle`
- implement `TransferTrbVariant::NoOp` and skip the nusb hardware requests if the submitted length is zero
- add missing endpoint context writes when entering stopped state
- change `WorkerState::StoppedWithContinuableTrb` to behave the same as `WorkerState::Stopped` and add a note that `WorkerState::StoppedWithContinuableTrb` will need more work to behave as intended
- add a nix integration test using a QEMU emulated USB to serial adapter

### DIFF
--- a/src/device/xhci/endpoint_handle.rs
+++ b/src/device/xhci/endpoint_handle.rs
@@ -744,14 +744,16 @@ impl<RIEH: RealInEndpointHandle> EndpointHandle for TdBasedInEndpointHandle<RIEH
 
     fn next_completion(&mut self) -> Self::TrbCompletionFuture<'_> {
         Box::pin(async {
-            match mem::take(&mut self.submission_state) {
-                TdBasedNormalSubmissionState::CollectingTd(trbs) => {
-                    self.submission_state = TdBasedNormalSubmissionState::CollectingTd(trbs);
-                    Ok(TrbProcessingResult::Ok)
+            match &mut self.submission_state {
+                TdBasedNormalSubmissionState::CollectingTd(_) => Ok(TrbProcessingResult::Ok),
+                TdBasedNormalSubmissionState::UnsupportedTrb => {
+                    self.submission_state = TdBasedNormalSubmissionState::default();
+                    Ok(TrbProcessingResult::TrbError)
                 }
-                TdBasedNormalSubmissionState::UnsupportedTrb => Ok(TrbProcessingResult::TrbError),
                 TdBasedNormalSubmissionState::AwaitingRealTransfer(trbs) => {
                     let completion = self.real_ep.next_completion().await?;
+                    let trbs = mem::take(trbs);
+                    self.submission_state = TdBasedNormalSubmissionState::default();
                     let processing_result = process_real_transfer_response(
                         self.endpoint_id,
                         self.slot_id,
--- a/src/device/xhci/endpoint_handle.rs
+++ b/src/device/xhci/endpoint_handle.rs
@@ -829,7 +829,7 @@ struct TdProcessingInfo<'a> {
 enum TdProcessingState {
     Default,
     // no more data, skip forward to next TD
-    ShortTransfer,
+    ShortTransfer(usize),
 }
 
 impl<'a> TdProcessingInfo<'a> {
@@ -884,7 +884,11 @@ impl<'a> TdProcessingInfo<'a> {
                                 );
                                 self.event_sender.send(transfer_event)?;
                             }
-                            self.state = TdProcessingState::ShortTransfer;
+                            self.state = TdProcessingState::ShortTransfer(bytes_missing);
+
+                            pcap::in_completion(self.pcap_meta, addr, bytes);
+
+                            return Ok(None);
                         }
                         _ => {
                             let (completion_code, processing_result) = match self.status {
@@ -938,9 +942,22 @@ impl<'a> TdProcessingInfo<'a> {
 
                 Ok(None)
             }
-            TdProcessingState::ShortTransfer => {
+            TdProcessingState::ShortTransfer(bytes_missing) => {
                 // Skip all Normal TRBs.
                 // We will need more handling here once we support EventData TRBs.
+
+                if trb_data.interrupt_on_completion {
+                    let transfer_event = EventTrb::new_transfer_event_trb(
+                        addr,
+                        bytes_missing as u32,
+                        CompletionCode::ShortPacket,
+                        false,
+                        self.endpoint_id,
+                        self.slot_id,
+                    );
+                    self.event_sender.send(transfer_event)?;
+                }
+
                 Ok(None)
             }
         }
--- a/src/device/xhci/hotplug_endpoint_handle.rs
+++ b/src/device/xhci/hotplug_endpoint_handle.rs
@@ -219,6 +219,8 @@ impl<EH: EndpointHandle> BaseEndpointHandle for HotplugEndpointHandleImpl<EH> {
                 // After the lock release, the device will be gone, so no need to actually acquire the lock.
             }
 
+            self.submission_state = HotplugSubmissionState::NoTrbSubmitted;
+
             Ok(())
         })
     }


Noi0103:
As a side note: I tested this branch with two USB to serial adapters connected to each other. One in the host and one in the guest, communication between the host and guest over hardware serial so to speak. This worked fine with minicom (display of "ä", "ü", "ö" etc is broken but that looked intentional).

src/device/xhci/endpoint_handle.rs
```diff
@@ -744,14 +747,13 @@ impl<RIEH: RealInEndpointHandle> EndpointHandle for TdBasedInEndpointHandle<RIEH
...
-                    self.submission_state = TdBasedNormalSubmissionState::CollectingTd(trbs);
-                    Ok(TrbProcessingResult::Ok)
-                }
+            match &mut self.submission_state {
+                TdBasedNormalSubmissionState::CollectingTd(_) => Ok(TrbProcessingResult::Ok),
                 TdBasedNormalSubmissionState::UnsupportedTrb => Ok(TrbProcessingResult::TrbError),
                 TdBasedNormalSubmissionState::AwaitingRealTransfer(trbs) => {
                     let completion = self.real_ep.next_completion().await?;
+                    let trbs = mem::take(trbs);
+                    self.submission_state = TdBasedNormalSubmissionState::default();
```
snue, line 759: Is there an actual bug with the old logic? The behavior differs now regarding resetting the state to default.

src/device/xhci/endpoint_handle.rs
```diff
@@ -883,6 +885,7 @@ impl<'a> TdProcessingInfo<'a> {
                                 self.event_sender.send(transfer_event)?;
                             }
                             self.state = TdProcessingState::ShortTransfer;
+                            return Ok(None);
```
snue: This path does not record pcap completions now. But previously it would send two IOC events, one with short packet, and one with success.

src/device/xhci/endpoint_handle.rs
```diff
@@ -939,6 +942,13 @@ impl<'a> TdProcessingInfo<'a> {
             TdProcessingState::ShortTransfer => {
                 // Skip all Normal TRBs.
                 // We will need more handling here once we support EventData TRBs.
+
+                // We do need to handle the case from xhci specification page 193 in chapter 4.10.1.1.2:
+                // > If the Short Packet occurred while processing a Transfer TRB with only an ISP
+                // > flag set, then two events shall be generated for the transfer; one for the Transfer
+                // > TRB that the Short Packet occurred on, and a second for the last TRB with the
+                // > IOC flag set.
+
```
snue, line 964: We don't handle the `ISP` flag at all currently. We will have to generate one event for that.

It's a bit unclear to me what the "last TRB" means here. I would have guessed it is the "next" TRB with the IOC flag of the current TD. Probably the last of the TD, but may also just be a packet boundary? Or the spec really means the last TRB of the active TD.

Maybe we can approximate by just handling all the IOC flags here for now.

src/device/xhci/hotplug_endpoint_handle.rs
```diff
@@ -211,6 +211,7 @@ impl<EH: EndpointHandle> BaseEndpointHandle for HotplugEndpointHandleImpl<EH> {
             if let Ok(mut guard) = self.endpoint_handle.try_lock() {
                 if let Some(device) = guard.as_mut() {
                     device.cancel().await?;
+                    self.submission_state = HotplugSubmissionState::NoTrbSubmitted;
```
snue: Should this be outside the `if let`?

src/device/xhci/endpoint_handle.rs
Noi0103: Yes, this `next_completion` is called within a `select!()` and might wait for an undefined amount of time at the line `let completion = self.real_ep.next_completion().await?;`.
When the calling `select!()` cancels this path, `mem::take()` already replaced the `self.submission_state` with its default.

The return value would have given back the trbs in most cases, but it is not cancel safe as we need when calling with `select!()`.

src/device/xhci/endpoint_handle.rs
Noi0103: ~~In theory we need separate paths for IOC & ISP or only ISP on the specific TRB. implementing proper ISP only handling seemed like a future todo that should then also touch on this condition:~~
```diff
- trb_data.interrupt_on_completion || trb_data.interrupt_on_short
+ trb_data.interrupt_on_completion && trb_data.interrupt_on_short
```

~~For now I wanted to leave it as is and just add the comment. Should there be some sort of 
`// TODO` or `todo!()` marker to better find it?~~

src/device/xhci/endpoint_handle.rs
Noi0103: I understood it as "the last TRB of the current TD". That made the most sense to me in terms of events carry the address of TRB's that triggered it. The driver should still probably know how long the remaining TD/where the end of the TD is by itself.

src/device/xhci/hotplug_endpoint_handle.rs
Noi0103: I'm not sure. If something is submitted and then waiting for a response is forcefully canceled through detach, moving this should not make a functional difference.

lbeierlieb: It should be outside.
If a device is already detached, and the driver does SetTrDequeuePointer and a new doorbell before it realizes the device is gone, we will run into the `"submit_trb called twice without calling next_completion"` assertion in `HotplugEndpointHandleImpl::submit`.

We crash the endpoint worker, and, with the current  design, the crash will spread to the SlotWorker and CommandWorker through channel failures.

nix/checks/testutils.nix
```diff
@@ -274,6 +274,11 @@ let
   # Fill in a template for the qemu.options list for a USB keyboard.
   mkQemuKeyboard = deviceBus: devicePort: "-device usb-kbd,bus=${deviceBus}.0,port=${devicePort}";
 
+  # Add a usb to serial adapter
+  mkQemuSerialAdapter =
+    deviceId: deviceBus: devicePort:
+    "-chardev pipe,id=charusbserial,path=/tmp/usbserial -device usb-serial,chardev=charusbserial,id=${deviceId},bus=${deviceBus}.0,port=${devicePort}";
```
snue: Would it make sense to base the path on the device-id? Currently, you can only have one such device, besides the possibility to declare multiple. Adding more than one just breaks stuff.

nix/checks/testutils.nix
```diff
@@ -108,7 +108,7 @@ let
   # Putting the socket in a world-readable location is obviously not a
   # good choice for a production setup, but for this test it works
   # well.
-  usbvfiodSocket = "/tmp/usbvfio";
+  usbvfiodSocket = "/tmp/usbvfiod";
   usbvfiodSocketHotplug = "/tmp/hotplug";
```
snue: nit: I prefer socket paths to have a `.sock` suffix for clarity. It doesn't make any functional difference of course.

nix/checks/testutils.nix
```diff
@@ -471,6 +478,10 @@ let
           builtins.map (mkPrepareBlockdeviceImages args.name) args.virtualDevices
         )}
 
+        import subprocess
+        # prepare serial device communication (even if not used)
+        subprocess.run(["mkfifo", "/tmp/usbserial.in", "/tmp/usbserial.out"])
+
```
snue: This should be unneeded when we use a unix socket for the chardev instead of pipes.

nix/checks/testutils.nix
```diff
@@ -533,7 +544,7 @@ in
       debug :: Bool
       virtualDevices :: [
         {
-        type :: "blockdevice" || "hid-device"
+        type :: "blockdevice" || "hid-device" || "serial"
```
snue: nit: the new entry does not match the "pattern". I think the device type does not need to spell out "device" in general, and we should probably drop it from the other types.

```suggestion
        type :: "block" || "hid" || "serial"
```

snue (commented):
I also tested this with an actual system providing console and getty over a passed through USB FTDI adapter. There is a recognizable slow-down compared to native host performance, but it works and is usable now. 🎉

nix/checks/testutils.nix
snue: I know you tested this and the description says there was an issue. I'll test this locally and see what the issue with the unix socket is. There is no obvious reason why it shouldn't just work.

src/device/xhci/endpoint_handle.rs
snue: The way I interpret the spec, they just want to explain what will happen due to a rather naive controller implementation. The expectation is that in a multi-TRB TD all have ISP set (optional on the last TRB), and the last has IOC. The ISP is only handled once per TD, on the one where the short read appears, and IOC is handled generically for every TRB in the TD for which it is set. If IOC and ISP are set on the short read TRB, only one event is generated. There are more parts of the spec that mention multiple events can be generated for TDs with errors.

So, correct handling seems to result when we just remember the TRB address for the Short Transfer state and handle IOC there as well (generating redundant events for the previous short transfer).

snue: Maybe we don't even need the old address for the redundant events, just the old completion code. Need to check the spec again.

src/device/xhci/endpoint_handle.rs
Noi0103: When revisiting I see i totally missed the behavior change you mentioned. The match arm `TdBasedNormalSubmissionState::UnsupportedTrb => Ok(TrbProcessingResult::TrbError),`, after moving `mem::take`, needs to explicitly reset the state to keep the behavior on that path.

src/device/xhci/endpoint_handle.rs
Noi0103: I re-read the chapter "4.10.1.1.2 Short Transfers when not using Event Data TRBs" again, with the context of "4.11.3.1 Transfer Event TRB".
The short transfer chapter does not actually make any assertions about rules to trigger events.

Considering "Table 6-22: Offset 0Ch – Normal TRB Field Definitions"'s description of the ISP flag:
> Note that if the ISP and IOC flags are both ‘1’ and a Short Packet is detected, then only one Transfer Event TRB shall be queued to the Event Ring. Also refer to section 4.10.1.1.

and 4.10.1.1.2 on page 194
> After the TD Completion Event, if any subsequent Transfer TRBs are
encountered with their IOC flag set while advancing to the end of the TD then those TRBs shall also generate an Event, where the Completion Code field shall return the same value as TD Completion Event and TRB Transfer Length field should return the same value as TD Completion Event. There are a couple of exceptions to this rule: [...]

We should add the IOC handling on skipped TRB's after a short transfer occured. And in that event we need to keep using the same value for the `trb_transfer_length` field we used in the initial TRB that triggered the short transfer event.

snue: Yes, as I said above, redundant IOC events on the short packet state are the way to go. And yes, we only need to keep transfer length and completion code, not the original short trb address for the redundant events.

And the first part about IOC and ISP both being set is what I meant with "If IOC and ISP are set on the short read TRB, only one event is generated." - that's already implemented by the existing check.

nix/checks/testutils.nix
Noi0103: If we want to enable multiple serial devices we could probably derive the deviceId from the index in the `arg.virtualDevices` to make the current static deviceID unique for multiple ones.
This could then also be done for the symlinks of the other device types. We could then remove the doc comment mentioning that when using multiple devices the `arg.virtualDevices.[i].udevRule.symlink` needs to be provided (port probably should still be).

I would be fine to just add a doc comment to restrict to one `usb-serial` qemu device in this PR and do this refactor later. Might we need multiple serial devices?

Noi0103:
Since this PR touches the testutils.nix anyway, I removed the boilerplate systemd config snippet that was passed as argument to `mkUsbTest` and prepared a open todo for using the `--fd` option.

nix/checks/testutils.nix
```diff
@@ -274,6 +325,11 @@ let
   # Fill in a template for the qemu.options list for a USB keyboard.
   mkQemuKeyboard = deviceBus: devicePort: "-device usb-kbd,bus=${deviceBus}.0,port=${devicePort}";
 
+  # Add a usb to serial adapter
+  mkQemuSerialAdapter =
+    deviceId: deviceBus: devicePort:
+    "-chardev pipe,id=charusbserial,path=/tmp/usbserial -device usb-serial,chardev=charusbserial,id=${deviceId},bus=${deviceBus}.0,port=${devicePort}";
+
   # Create a blockdevice or USB keyboard on our QEMU bus-id corresponding with the declared usb version.
```
snue:
```suggestion
  # Create a USB device on our QEMU bus-id corresponding with the declared usb version.
```

nix/checks/testutils.nix
snue: I figured out what the problem with the socket is. Our test script prelude ensures SSH to cloud-hypervisor works, but the pass-through device only shows up once a socket connection is established. Unfortunately the preparation snippets we have for block device images are too early (before QEMU creates the socket), and the actual `testScript` is too late to make the connection. Pipes are a decent workaround and will offer enough usability.

However, I'd like to move the `mkfifo` into a `prepareDevice` (currently `prepareBlockDeviceImages`) part of the test utils. And we can directly open the pipes and bind them to python objects for convenience there. We can base this on the symlink name, as that must be unique and exist anyway.

nix/checks/blockdevice.nix
```diff
@@ -105,11 +58,11 @@ builtins.listToAttrs (
...
       virtualDevices = [
         {
-          type = "blockdevice";
+          type = "block";
           inherit usbVersion;
         }
       ];
-      testScript = singleBlockDeviceTestScript;
-    } systemd-config;
+      testScript = BlockDeviceTestScript;
```
snue: We can also refactor these to just require `inherit testScript` here.

nix/checks/usb-serial.nix
```diff
@@ -0,0 +1,79 @@
...
+    # listen on the guest
+    t_cloud_hypervisor = threading.Thread(target=read_cloud_hypervisor)
+    t_cloud_hypervisor.start()
+
+    # write in host
+    with open('/tmp/usbserial.in', "w") as fd:
+      # for some reason removing any of those lines breaks the test
+      fd.write('some\n')
+      fd.flush()
+      time.sleep(1)
```
snue: That looks fishy. It should work with a more simple logic. Pipes buffer writes before they block. That's fine for us and the test logic can basically rely on non-blocking writes (for small data).

nix/checks/usb-serial.nix
```diff
@@ -0,0 +1,79 @@
...
+    out = cloud_hypervisor.succeed("lsusb -v -d 0403:6001")
+    print(out)
+
+    out = cloud_hypervisor.succeed("ls -ls /dev/ttyUSB0")
+    print(out)
+
+    cloud_hypervisor.succeed('echo "preparations done"')
+
+    import time
+    import threading
```
snue: I don't think you need the threading logic. Just send before you receive. And with added convenience in the prelude you can get away without `subprocess` as well.

nix/checks/testutils.nix
```diff
@@ -542,73 +611,48 @@ in
...
+    } -> a
     ```
 
     # Examples
-    :::{.example}
 
-    When Using more than one device each device should define its usbPort and udevRule.symlink (the default value is static).
+    When Using more than one block or hid device each device should define its usbPort and udevRule.symlink (the default value is static).
+
+    Only one serial device is currently supported.
```
snue, line 633: It's a general rule:

```suggestion
    When using more than one device, each shall define its usbPort and udevRule.symlink (the default values are static).
```

And we can easily support multiple serial devices (even if unneeded) by basing the fifo names and associated python objects on the symlink name.

snue (commented):
The refactor could have gone to a separate PR to focus on the actual fixes, but it makes the tests much cleaner.

nix/checks/testutils.nix
```diff
@@ -365,13 +375,33 @@ let
...
+    in
+    ''
+      import os
+      import socket
+      ${socket}="/tmp/${socket}.sock"
+      if os.path.exists(${socket}):
+          os.remove(${socket})
+      ${server} = socket.socket(socket.AF_UNIX, socket.SOCK_STREAM)
+      ${server}.bind(${socket})
+      ${server}.listen(1)
```
snue: Oh well, somewhere we will have to `accept` the socket connection before QEMU will launch. This may actually be tricky to get right. I'll think about it a bit more, but we may need to fall back to pipes here.

snue: I was hoping to get the open/accepted connection in the `testutils` prelude and have the test script just use it. I don't see how to achieve this right now with less effort than just using pipes. Let's just go back to pipes that the test script can make direct use of through `symlink.{in/out}`.

src/device/xhci/endpoint_handle.rs
```diff
@@ -936,9 +945,23 @@ impl<'a> TdProcessingInfo<'a> {
 
                 Ok(None)
             }
-            TdProcessingState::ShortTransfer => {
+            TdProcessingState::ShortTransfer(bytes_missing) => {
                 // Skip all Normal TRBs.
                 // We will need more handling here once we support EventData TRBs.
+
+                // always handle IOC generic since we don't yet have event data
```
snue: What does event data have to do with it? Guests drivers shouldn't use EventData and IOC, but we will still handle IOC on every transfer TRB we encounter.

snue (commented):
Did you revert fully to a previous state? That's not what I meant 😅 

We should keep the various improvements from previous discussions, just create the python pipe accessors in the testutils and not the test script.

Noi0103:
Yeah I was a bit too fast to push there again. I have the changes still locally and am working on bringing those to pipes.

src/device/xhci/endpoint_handle.rs
Noi0103: To be completely hones I have no clue why I wrote that something 3 weeks ago. Thanks, I'll remove it.

nix/checks/testutils.nix
```diff
@@ -457,12 +480,6 @@ let
         ];
 
         services = {
-          # The framework automatically forwards all journal output to ttyS0,
-          # slowing down the test significantly if there is a lot of logs.
-          journald.extraConfig = lib.mkForce ''
-            ForwardToConsole=yes
-            TTYPath=/dev/hvc1
-          '';
```
snue: Why remove that?

nix/checks/testutils.nix
```diff
@@ -365,13 +374,27 @@ let
...
+  # Create socket for the backend chardev of the usb-serial qemu device.
+  mkPrepareSerialdevicePipe =
+    testname: device:
+    let
+      pipename = "${testname}-${device.udevRule.symlink}";
+    in
+    ''
+      import subprocess
+      # prepare serial device communication (even if not used)
+      subprocess.run(["mkfifo", "/tmp/${pipename}.in", "/tmp/${pipename}.out"])
```
snue: I really dislike shelling out to the OS for such things. Let's use `os.mkfifo` instead. And please create a python object here with both opened fifos for the test to use.

snue (commented):
Tested the latest version with a 115200 baud serial.

I observed warnings for input overruns in the guest, but they are mostly harmless (we lose some bytes sent to the guest when it does not handle the interrupts fast enough). This was with `tmux` over `screen` showing `dmesg` and `top`.

The connection gets stuck after reopening the serial device in the guest. It's missing `NoOp` TRB support on the IN endpoint.

```
2026-08-21T07:40:11.608139Z  WARN usbvfiod::device::xhci::endpoint_handle: Encountered unsupported TRB on In Endpoint (slot 1, ep 7): NoOp
2026-08-21T07:40:11.608182Z  WARN usbvfiod::device::xhci::endpoint: invalid endpoint action: Doorbell in state Error
```

src/device/xhci/endpoint_handle.rs
```diff
@@ -936,9 +945,22 @@ impl<'a> TdProcessingInfo<'a> {
...
-            TdProcessingState::ShortTransfer => {
+            TdProcessingState::ShortTransfer(bytes_missing) => {
                 // Skip all Normal TRBs.
                 // We will need more handling here once we support EventData TRBs.
+
+                if trb_data.interrupt_on_completion {
+                    let transfer_event = EventTrb::new_transfer_event_trb(
+                        addr,
+                        bytes_missing as u32,
+                        CompletionCode::Success,
```
snue: Spec?

I read in `4.10.1.1.2`:

> After the TD Completion Event, if any subsequent Transfer TRBs are
encountered with their IOC flag set while advancing to the end of the TD then
those TRBs shall also generate an Event, where the Completion Code field shall
return the same value as TD Completion Event and TRB Transfer Length field
should return the same value as TD Completion Event. There are a couple of
exceptions to this rule:
•If the IOC flag is set in an Event Data TRB then an Event Data Transfer Event
shall be generated only if PAE = '1'.
•If the IOC flag is set in a Link TRB then the Transfer Event shall be generated
with Completion Code = Success, and the TRB Transfer Count = 0

The "TD Completion Event" was the initial `ShortPacket`.

> If a Short Packet occurs, then a Transfer Event shall be generated with the
Completion Code = Short Packet, its TRB Pointer field pointing to the Transfer
TRB that the Short Packet occurred on, and its TRB Transfer Length field shall
indicate the residue bytes in the buffer.

> If Event Data TRBs are not used, then the total number of received bytes for a
Short Packet TD is the sum of the TRB Transfer Length fields in all Transfer TRBs
up to and including the one that generated the Short Packet Event, minus the
residue value of the TRB Transfer Length field in the Short Packet Event.

So, completion code needs fixing, but `bytes_missing` for the transfer length is correct.

For the exceptions:
* Event Data will come with #238 
* LinkTRB does not generate events for IOC currently (see #303), but the exception is very valuable for our design if we want to change that.

Noi0103: But do we additionally need to have special handling for the last TRB of the TD? This is still in the chapter about short transfer but worded a bit ambiguous IMO.

> TD Completion Events are generated by few basic conditions:
• If the IOC flag = ‘1’ and the TD completes successfully, then a Transfer Event
shall be generated with its Completion Code = Success and TRB Transfer
Length identifying the number bytes transferred.

snue: That comment is for the successful completion case (no short transfer, no error). We already generate that event.

nix/checks/testutils.nix
Noi0103: I struggle with creating the python objects prior to qemu starting. Simply moving the `os.open()` calls to the testutils results in a no such file error (even with a few seconds of `time.sleep()`).

nix/checks/testutils.nix
```diff
@@ -599,11 +626,28 @@ in
     ```
 
     # Examples
-    :::{.example}
 
-    When Using more than one device each device should define its usbPort and udevRule.symlink (the default value is static).
+    When Using more than one block or hid device each device should define its usbPort and udevRule.symlink (the default value is static).
```
snue: The old version still applies even when mixing with serial devices. 

But you can fix the capitalization mistake on `Using` -> `using`

nix/checks/testutils.nix
```diff
@@ -599,11 +626,28 @@ in
     ```
 
     # Examples
-    :::{.example}
```
snue, line 629: nit: was this invalid markup? Should it move?

nix/checks/testutils.nix
```diff
@@ -599,11 +626,28 @@ in
...
 
-    When Using more than one device each device should define its usbPort and udevRule.symlink (the default value is static).
+    When Using more than one block or hid device each device should define its usbPort and udevRule.symlink (the default value is static).
+
+    Only one serial device is currently supported.
```
snue: I still think this restriction should be easy to lift. The python socket object or pipe(s) can just have the symlink name. That just requires them to be created in the common test script. A separate phase/snippet after launching QEMU but before waiting for the nested guest service should do it.

nix/checks/testutils.nix
```diff
@@ -519,10 +543,8 @@ let
...
         ${nestedPythonClass}
 
-        # prepare blockdevice images if necessary
-        ${lib.concatStringsSep "\n" (
-          builtins.map (mkPrepareBlockdeviceImages args.name) args.virtualDevices
-        )}
+        # prepare devices
+        ${lib.concatStringsSep "\n" (builtins.map (mkPrepareDevice args.name) args.virtualDevices)}
 
         start_all()
```
snue, line 572: So, even with the pipes we cannot simply open the read and write ends before launching QEMU (`start_all`) without blocking? Have you tried opening them explicitly in non-blocking mode? Alternatively we can add another `mkPrepareTest` phase here after the `start_all` that will currently only establish the serial connection. I dislike that the serial testscript has to rely on known `pipename` locations. With this separate phase, a single `socket` managed by QEMU may also be an option again.

nix/checks/testutils.nix
```diff
@@ -569,7 +591,12 @@ in
 
   /**
     Create a pkgs.testers.runNixOSTest with specific purpose of testing Usbvfiod.
-    The Functions purpose is to remove duplicated lines, make comparing tests easier and write new tests with less boilerplate.
+    The Functions purpose is to remove duplicated lines, make comparing tests easier
+    and write new tests with less boilerplate.
+
+    For the testscript this function provides an object from the nested running
+    `cloud_hypervisor` vm, that can use `.succeed()` and `.wait_until_succeeds()`
+    just like the qemu nixos test `machine` object can.
```
snue, line 625: Nice. When we figure out how to conveniently grant access to the serial, this comment could also explain that mechanism.

nix/checks/testutils.nix
Noi0103: I tried to use this in the prepare step: 
```python
      import os
      os.mkfifo("/tmp/${pipename}.in")
      os.mkfifo("/tmp/${pipename}.out")
      pipein = os.open("/tmp/${pipename}.in", os.O_WRONLY | os.O_NONBLOCK)
      pipeout = os.open("/tmp/${pipename}.out", os.O_RDONLY | os.O_NONBLOCK)
```

I will try to go with a socket in a second prepare device step after the `start_all()`.

nix/checks/testutils.nix
Noi0103: This seemed to have no effect when tested with
```
nix run nixpkgs#nixdoc -- --file ./nix/checks/testutils.nix --category "" --description "" --prefix "" --anchor-prefix "" > testutils.md
``` 
so I did not see a reason to keep it.

Noi0103:
The name of the python object that provides the serial socket is very much subject to change, but I have not thought about what to use very much. Using the symlink as is probably is fine.

I was testing around what does and what does not work with the fishy script snippet where two strings are sent towards the nested guest. It seems to be necessary to (1) send any character, (2) append "\n" and (3) have a short sleep before finally sending the intended sequence.

src/device/xhci/endpoint_handle.rs
```diff
@@ -841,6 +851,23 @@ impl<'a> TdProcessingInfo<'a> {
...
+                        self.endpoint_id,
+                        self.slot_id,
+                    );
+                    self.event_sender.send(transfer_event)?;
+
+                    pcap::in_completion(self.pcap_meta, trb.addr, &[]);
+                    Ok(None)
+                }
+                TdProcessingState::ShortTransfer(_) => Ok(None),
+            },
```
snue: Do we need to differentiate state here? Shouldn't we just look at IOC? Also do we log completion for this? Where do we log submission?

src/device/xhci/trb.rs
```diff
@@ -1093,6 +1093,42 @@ impl TrbData for EventDataTrbData {
...
+    /// # Limitations
+    ///
+    /// The function currently does not check if the slice respects RsvdZ
+    /// fields.
+    fn parse(trb_bytes: RawTrbBuffer) -> Result<Self, TrbParseError> {
+        let trb_type = trb_bytes[13] >> 2;
+        assert_eq!(
+            trb_types::NO_OP,
+            trb_type,
+            "SetupStageTrbData::parse called on TRB data with incorrect TRB type ({trb_type:#x})"
```
snue:
```suggestion
            "NoOpTrbData::parse called on TRB data with incorrect TRB type ({trb_type:#x})"
```

src/device/xhci/trb.rs
```diff
@@ -851,7 +851,7 @@ impl TransferTrbVariant {
             trb_types::STATUS_STAGE => parse(Self::StatusStage, bytes),
             trb_types::ISOCH => Self::Isoch,
             trb_types::EVENT_DATA => parse(Self::EventData, bytes),
-            trb_types::NO_OP => Self::NoOp,
+            trb_types::NO_OP => parse(Self::NoOp, bytes),
```
snue, line 854: Do we need to differentiate between `NO_OP` and `NO_OP_COMMAND`?

Noi0103: I believe we should. Only the `NO_OP` has an IOC and CH bit (and ENT we are not using AFAIK). We could parse `NO_OP_COMMAND` just like the `NO_OP`, but I think this might bring more harm than good.

src/device/xhci/endpoint_handle.rs
Noi0103: Yeah, only IOC should be of interest, not state and not PCAP.

nix/checks/testutils.nix
Noi0103: From what I have seen and tested around with for the docs this should be broken due to missing closing delimiters `:::`. Because creating a sample render that uses those annotations will need more work I will leave this for a separate PR and avoid touching it here (i.e. not remove the line but also not touch the nix docs rendering flow).

src/device/xhci/trb.rs
snue: I meant, do we actually differentiate currently? And yes, our code uses `CommandTrbVariant::parse` for `NO_OP_COMMAND`. No overlap with the `NO_OP` handling here. This is what we want (for the reasons you mention).

nix/checks/testutils.nix
```diff
@@ -383,15 +393,27 @@ let
...
-      mkPrepareOneBlockdeviceImage testname device
-    else
-      "";
+    let
+      socket = "${device.udevRule.symlink}";
+    in
+    ''
+      import socket
+      ${socket} = socket.socket(socket.AF_UNIX, socket.SOCK_STREAM)
+      ${socket}.setsockopt(socket.SOL_SOCKET, socket.SO_SNDBUF, 0)
```
snue: nit: have you tried the latest version without the `setsockopt`? Would be good to know if we really need this for the test.

nix/checks/testutils.nix
```diff
@@ -587,7 +610,21 @@ in
...
+    `cloud_hypervisor` vm, that can use `.succeed()` and `.wait_until_succeeds()`
+    just like the qemu nixos test `machine` object can.
+
+    When using more than one device, each should define its `usbPort` and
+    `udevRule.symlink` (the default value is static).
+
+    For each serial device this function provides a python socket object with the
+    name equal to the symlink. Use this object to send and receive data over the
+    usb-serial adapter.
+    Note: Serial devices can not use `attachedOnStartup = "none";`.
```
snue: nit:
```suggestion
    Note: Serial devices cannot use `attachedOnStartup = "none";`.
```

nix/checks/testutils.nix
```diff
@@ -587,7 +610,21 @@ in
...
     Create a pkgs.testers.runNixOSTest with specific purpose of testing Usbvfiod.
-    The Functions purpose is to remove duplicated lines, make comparing tests easier and write new tests with less boilerplate.
+    The Functions purpose is to remove duplicated lines, make comparing tests easier
+    and write new tests with less boilerplate.
+
+    For the testscript this function provides an object from the nested running
+    `cloud_hypervisor` vm, that can use `.succeed()` and `.wait_until_succeeds()`
+    just like the qemu nixos test `machine` object can.
+
+    When using more than one device, each should define its `usbPort` and
```
snue: nit: using the RFC meaning of the words, this is a mandatory condition for correct operation.

```suggestion
    When using more than one device, each shall define its `usbPort` and
```

nix/checks/usb-serial.nix
```diff
@@ -0,0 +1,70 @@
...
+        time.sleep(1)
+        ${symlink}.sendall(b"host is writing into the emulated usb-serial device")
+        print(f'wrote loop {i}')
+
+    t_host_write = threading.Thread(target=host_write)
+    t_host_write.start()
+
+    # ...listen on `/dev/ttyUSB0` in the nested guest
+    print("starting to listen in cloud hypervisor")
+    out = cloud_hypervisor.succeed('socat /dev/ttyUSB0 - | head -c 51')
```
snue: Why 51? Twice the expected string length -1 because you don't know where the reading starts? Can we have a comment explaining the choice of constants?

src/device/xhci/trb.rs
```diff
@@ -1093,6 +1093,42 @@ impl TrbData for EventDataTrbData {
...
+    /// fields.
+    fn parse(trb_bytes: RawTrbBuffer) -> Result<Self, TrbParseError> {
+        let trb_type = trb_bytes[13] >> 2;
+        assert_eq!(
+            trb_types::NO_OP,
+            trb_type,
+            "NoOpTrbData::parse called on TRB data with incorrect TRB type ({trb_type:#x})"
+        );
+
+        let chain = trb_bytes[12] & 0x10 != 0;
```
snue, line 1122: > Software shall not define a No Op Transfer TRB within a multi-TRB TD, i.e.
software shall never set the Chain bit of a No Op TRB t o '1' and a No Op TRB
shall always be preceded by a TRB whose Chain bit is also set to '0'.

We don't enforce this, we just skip the NoOps and generate events as requested. But we do not actually need to parse/store the chain bit either? Or does anything look at the parsed bit currently? You can issue a warning here if `chain` is ever set.

snue (commented):
Nice. Does this now allow repeated connections to the serial from the guest?

Left some nits or wording suggestions. Looks good otherwise.

src/device/xhci/trb.rs
Noi0103: I only looked into the definition of the data structure in `6.4.1.4 No Op TRB` and assumed same CH bit rules.

We could not parse it and return a static `false` for this instead:
```rust
impl SupportedInEndpointTrb {
    const fn chain(&self) -> bool {
        match &self.variant {
            SupportedInEndpointTrbVariant::Normal(normal_trb_data) => normal_trb_data.chain,
            SupportedInEndpointTrbVariant::EventData(event_data_trb_data) => {
                event_data_trb_data.chain
            }
            SupportedInEndpointTrbVariant::NoOp(noop) => noop.chain,
        }
    }
...
```

A warning sounds good in any case.

nix/checks/usb-serial.nix
Noi0103: This is the exact string length and in theory should enable flaky behaviour, but I have yet to hit it when running the test.

> Twice the expected string length -1

Not sure if we should initially go for this or see where the current version takes us.

nix/checks/testutils.nix
Noi0103: I have not thought of that. Seems to work fine without :+1: .

src/device/xhci/trb.rs
Noi0103: To support a TD with interleaved NoOp ('shall not happen') we still parse and use the chain bit. If we assume the chain bit as 0 we would break a TD in half that is operating on undefined behaviour. It being undefined anyway we can also allow such TD.

snue: Agreed, let's just add the warning, because it indicates a driver issue, but handle it as if it was allowed.

nix/checks/usb-serial.nix
snue: Yes, let's stay with the exact length. It's nice to see this work reliably. But maybe add a comment that this depends on the string being received correctly from the start. It will help when the search below fails on a flaky run that doesn't read from the start.

nix/checks/usb-serial.nix
```diff
@@ -0,0 +1,74 @@
...
+    # Write in the nested guest...
+    # -dm ignore $STY for session creation in detached mode
+    # -S sessionname
+    cloud_hypervisor.succeed('screen -dm -S serialsession /dev/ttyUSB0 115200')
+    message = "guest is writing into the serial"
+
+    # ...and use the socket buffer to receive in the host.
+    # -X send command stuff and arg to the with -S specified session
+    cloud_hypervisor.succeed(f'screen -S serialsession -X stuff "{message}"')
+    cloud_hypervisor.succeed('screen -S serialsession -X quit')
```
snue: The `screen` cmdline usage seems very versatile. You use `socat` in the nested guest later. That basically proves the device is reusable. However, you could keep the `screen` session open and query it for the data sent by the host. This avoids the sending thread and the loop/sleep and the extra `socat`. Then close the session afterward. This could then be looped 2 or 3 times to keep re-exercising the device? If we get rid of the explicit `sleep` this is probably still faster than the current execution.

src/device/xhci/trb.rs
```diff
@@ -1093,6 +1093,47 @@ impl TrbData for EventDataTrbData {
...
+        if chain {
+            // specification chapter 4.11.7
+            // "[...] software shall never set the Chain bit of a No Op TRB t o '1' [...]"
```
snue, line 1125:
```suggestion
            // "[...] software shall never set the Chain bit of a No Op TRB to '1' [...]"
```

Noi0103: I wanted to enable searching and for whatever reason CTRL+F in Firefox will only find this string in the specification with this space. This is a copy and paste citation from the document.

nix/checks/usb-serial.nix
Noi0103: I have not thought of reusing screen, I like this version a lot more, thank you.

nix/checks/usb-serial.nix
```diff
@@ -0,0 +1,84 @@
...
+
+      # Write from the host...
+      ${symlink}.sendall(b"host is writing into the emulated usb-serial device")
+
+      # On consecutive runs we need more writes and some wait time, the beginning is cut off.
+      if i > 1:
+        for n in range(1, 9):
+          ${symlink}.sendall(b"\r\n")
+          ${symlink}.sendall(b"host is writing into the emulated usb-serial device")
+          time.sleep(1)
```
snue: Now this is really slowing down the test (+8s for each iteration).

`screen` has a command to redraw the window (`redisplay`). Also, the screen window is typically `80x25` characters with automatic wrapping. Maybe just start the host string with a newline to guarantee it doesn't wrap at the receiver. Maybe even explicit `\r\n` to start at column 0. Also, there is `log` to start writing a screenlog file instead of `hardcopy`, which is just a "screenshot" of the current window contents. Maybe that can help with the cutoff beginning?

nix/checks/usb-serial.nix
```diff
@@ -0,0 +1,84 @@
...
+      # clean exit
+      out = cloud_hypervisor.succeed('lsof /dev/ttyUSB0')
+      print(out)
+      out = cloud_hypervisor.succeed('screen -list')
+      print(out)
+
+      # The builtin quit sometimes hangs with i number of loops.
+      cloud_hypervisor.succeed(f'pkill -9 -f "SCREEN.*serialsession{i}"')
+      cloud_hypervisor.succeed('rm -rf /root/.screen')
+      cloud_hypervisor.succeed('while lsof /dev/ttyUSB0; do sleep 0.5; done')
```
snue: I like this check. In case this fails with timeout, do the command invocation logs contain the `lsof` output already? Or are we blind to what kept using the device?

nix/checks/usb-serial.nix
```diff
@@ -0,0 +1,84 @@
...
+
+      # clean exit
+      out = cloud_hypervisor.succeed('lsof /dev/ttyUSB0')
+      print(out)
+      out = cloud_hypervisor.succeed('screen -list')
+      print(out)
+
+      # The builtin quit sometimes hangs with i number of loops.
+      cloud_hypervisor.succeed(f'pkill -9 -f "SCREEN.*serialsession{i}"')
+      cloud_hypervisor.succeed('rm -rf /root/.screen')
```
snue: This hard kill is unfortunate and rather unclean. I really wonder why `quit` would hang. Can you try with a SIGHUP instead of SIGKILL?  The `autodetach` setting is on by default and needs to be switched off for SIGHUP to terminate (and not just detach) the sessions. But that probably doesn't change anything if it is blocked while terminating.

Did you deliberately switch to separate session names to identify which one hangs or to avoid any reuse?

nix/checks/usb-serial.nix
```diff
@@ -0,0 +1,84 @@
...
+      print(out)
+      out = cloud_hypervisor.succeed('cat /tmp/screen_output.txt')
+      print("ttyUSB in guest received: " + out)
+      search("host is writing into the emulated usb-serial device", out)
+
+      # clean exit
+      out = cloud_hypervisor.succeed('lsof /dev/ttyUSB0')
+      print(out)
+      out = cloud_hypervisor.succeed('screen -list')
+      print(out)
```
snue: I suppose this is currently informational to help debugging the session issues.

nix/checks/usb-serial.nix
snue: The `rm -rf` should be avoided/replaced with `screen -wipe` to remove any dead sessions. I think it does not remove active or detached sessions.

snue (commented):
Thanks for adapting the test logic to use `screen` exclusively. I left some ideas on the current unexpected behavior in the test (hangs and missed input).

Noi0103:
I tested around a bit (also with a loop to 40 instead of 4) and it seems that `-X quit` is not the only screen command usage that can get stuck. I have seen the spawning of the session and writing into the session being stuck too.

There is also a small fix that caused us to drop one request worth of data when a NoOp was used.

nix/checks/usb-serial.nix
Noi0103: We should be blind right now.

nix/checks/usb-serial.nix
Noi0103: \r\n is only for human readability and to have a better time when counting lines.

> there is log to start writing a screenlog file instead of hardcopy

I read about `log` after finishing the version with `hardcopy` and considered changing it not worth. We write so little we are not even close to filling a screen page. I would also consider `redisplay` highly optional with how short-lived these screen sessions are.
The current state feels quite reproducible (with only 3 lines written and the 3rd being received). I have not tested how `log` might change the behaviour.

nix/checks/usb-serial.nix
Noi0103: Comparing quit and kill, both seem to have problems eventually and either the exiting or some following screen command hangs. With the kill it just takes longer when increasing the loop count. 

I tested with SIGHUP, but it was stuck within the first loop.

Separate session names help to identify the current loop iteration without any searching in previous logs (just convenience).

src/device/xhci/endpoint_handle.rs
```diff
@@ -772,6 +789,24 @@ impl<RIEH: RealInEndpointHandle> EndpointHandle for TdBasedInEndpointHandle<RIEH
                     )?;
                     Ok(processing_result)
                 }
+                TdBasedNormalSubmissionState::NoOpTD(trbs) => {
+                    let completion = InTrbProcessingResult {
+                        status: Success,
+                        data: vec![],
+                    };
+                    let trbs = mem::take(trbs);
+                    self.submission_state = TdBasedNormalSubmissionState::default();
```
snue: I don't quite get this. This only covers the Spec-compliant TRB NoOp TDs (with a single TRB)? But we still allow NoOps basically anywhere? Then you throw away all TRBs collected so far? If you want to go this route, we must raise `TrbError` on all the invalid NoOp locations (with chain bit or after TRBs that have the chain bit set).

src/device/xhci/endpoint_handle.rs
```diff
@@ -772,6 +789,24 @@ impl<RIEH: RealInEndpointHandle> EndpointHandle for TdBasedInEndpointHandle<RIEH
...
+                    let processing_result = process_real_transfer_response(
+                        self.endpoint_id,
+                        self.slot_id,
+                        completion,
+                        trbs,
+                        &self.event_sender,
+                        &self.dma_bus,
+                        self.pcap_meta,
+                    )?;
```
snue: But there was no real transfer?

src/device/xhci/endpoint_handle.rs
```diff
@@ -736,6 +738,21 @@ impl<RIEH: RealInEndpointHandle> EndpointHandle for TdBasedInEndpointHandle<RIEH
...
+                replace_with_or_abort(&mut self.submission_state, |old_state| {
+                    let TdBasedNormalSubmissionState::CollectingTd(trbs) = old_state else {
+                        unreachable!(
+                            "verified the state is CollectingTd at the start of the function"
+                        );
+                    };
+                    TdBasedNormalSubmissionState::NoOpTD(trbs)
+                });
+
+                return Ok(());
```
snue: Ok, only if the complete TD has a length of zero ... TDs with interleaved NoOps (not quite spec compliant) still do a real hardware request.

src/device/xhci/endpoint_handle.rs
```diff
@@ -736,6 +738,21 @@ impl<RIEH: RealInEndpointHandle> EndpointHandle for TdBasedInEndpointHandle<RIEH
...
+            }
+
             self.real_ep.submit(td_request_length)?;
```
snue: Isn't this an implementation detail of the `real_ep` / `RealDevice` abstraction (i.e `nusb`)? Should we really handle this here? I imagine we could push the responsibility to the `real_ep.submit()` and we do not need the extra `NoOpTD` state here.

src/device/xhci/endpoint.rs
```diff
@@ -194,7 +194,8 @@ impl<EH: HotplugEndpointHandle> EndpointWorker<EH> {
                     }
                     EndpointMessage::Doorbell => {
                         self.context.set_state(endpoint_state::RUNNING);
-                        self.state = WorkerState::WaitForTrbCompletion;
+                        self.real_endpoint.cancel().await?;
+                        self.state = WorkerState::LookForTrb;
                     }
```
snue, line 208: Now there is no path that would ever continue with the previously submitted TRB (as indicated by the state name)? We always reload/resubmit the TRB from the current dequeue pointer? How does the guest identify this as the "wrong TD"? Did it modify the TRB while stopped? The TRB address should still match, right? The `next_trb()` (called from `LookForTrb` state) just reads from the dequeue pointer and this should not have advanced yet.

If this is indeed the right behavior, I would argue we should cancel the request before updating the context state to `STOPPED`. And maybe we don't even need the `WithContinuableTrb` state with this at all?

nix/checks/testutils.nix
```diff
@@ -619,9 +654,23 @@ in
...
+    ```nix
+    myTest = mkUsbTest {
+      name = "foo";
+      virtualDevices = [
+        {
+          type = "block";
+        }
+      ];
+      testScript = ''
+        cloud_hypervisor.succeed("echo hello", timeout=60)
```
snue, line 678: nit: will need an update after #308. (`timeout=ONE_MINUTE`)

nix/checks/testutils.nix
snue: nit: Also various occurrences of outdated timeout in this example.

nix/checks/testutils.nix
```diff
@@ -162,6 +162,14 @@ let
       def __init__(self, vm_host: BaseMachine) -> None:
         self.vm_host = vm_host
 
+      def execute(self, *commands: str, timeout: int | None = None) -> str:
```
snue: nit: `timeout` should be `datetime.timedelta` after rebasing.

snue (commented):
Quite a lot of fixes that were required. I think some of these should be solved differently. The tests seem to be stable now. You can squash the last two commits that add the test.

Please adapt the PR title and description to mention the various fixes that should be relevant to other device types. With 13 commits you could have split off some of these early into their own PRs as well. I would suggest something like `Various endpoint handling fixes to enable serial devices`.

src/device/xhci/endpoint.rs
snue: Could it be this is an artifact of not updating the context dequeue pointer (only the state) when switching to this stopped/continuable state? In `endpoint.rs:156`. We also don't update it in `endpoint.rs:172` (Stop while Halted -> `ContextStateError` according to Spec), or `endpoint.rs:168` (Reset from Halted state). That may be redundant but probably won't hurt either. We should probably just ensure there is a full context update when moving to Halted state? 

(Two of these missing updates were flagged by a Claude Fable-5 report for the `main` branch.)

snue: > Did it modify the TRB while stopped?

Linux will rewrite TRBs to `NoOp` for any cancelled TDs, but here there was no cancellation, so probably no need to re-read/"submit" anything. It's likely just the context dequeue pointer mismatch the guest is unhappy about.

Noi0103: The old requests are cancelled by the Linux driver and kernel logs show this:
```
[   23.045459] xhci_hcd:xhci_dbg_trace:31: xhci_hcd 0000:00:04.0: Cancel URB 0000000050a2160f, dev 1, ep 0x81, starting at offset 0x75a22020
[   23.045459] xhci_hcd:xhci_ring_cmd_db:426: xhci_hcd 0000:00:04.0: // Ding dong!
[   23.049475] xhci_hcd:xhci_dbg_trace:31: xhci_hcd 0000:00:04.0: Removing canceled TD starting at 0x75a22020 (dma) in stream 0 URB 0000000050a2160f
[   23.049475] xhci_hcd:xhci_giveback_invalidated_tds:947: xhci_hcd 0000:00:04.0: xhci_giveback_invalidated_tds: Giveback cancelled URB 0000000050a2160f TD
[   23.060316] xhci_hcd:xhci_dbg_trace:31: xhci_hcd 0000:00:04.0: Cancel URB 000000004a1983a6, dev 1, ep 0x81, starting at offset 0x75a22010
[   23.060316] xhci_hcd:xhci_ring_cmd_db:426: xhci_hcd 0000:00:04.0: // Ding dong!
[   23.063089] xhci_hcd:xhci_dbg_trace:31: xhci_hcd 0000:00:04.0: Removing canceled TD starting at 0x75a22010 (dma) in stream 0 URB 000000004a1983a6
[   23.063682] xhci_hcd:xhci_giveback_invalidated_tds:947: xhci_hcd 0000:00:04.0: xhci_giveback_invalidated_tds: Giveback cancelled URB 000000004a1983a6 TD
[   23.719486] xhci_hcd 0000:00:04.0: Event dma 0x0000000075a22010 for ep 2 status 13 not part of TD at 0000000075a22030 - 0000000075a22030
```

snue: Ok, we were handling/continuing a TD that contains TRB `0x0000000075a22010`. But Linux overwrote it as `NoOp`. This is only allowed/expected for TDs *outside* the current one. So Linux indeed just isn't aware what the current TD is (because context has a stale Dequeue Pointer value). It should instead issue an explicit `SetTrDequeuePtr` to skip the TRB at `0x0000000075a22010`.

We never read-ahead more than the current TD, so we never need to "invalidate" our cached TRBs.

snue: Since the explicit context updates on `STOP` alone did not seem to resolve this, it may be tied to the missing generation of `Stopped Transfer Events` (which contain the dequeue pointer value). We only generate the Command Completion event.

But we may still need to cancel the current transfer in order to generate a correct transfer event. And continuing the TD will be rather complicated in that case. We would need to schedule a new `real_ep` request for the missing part of the TD. => Let's not deal with this in this PR.

I am almost certain Linux has to set the dequeue pointer here instead of simply ringing the doorbell. If Linux assumes we are "in between TDs", it's basically free to touch anything on the ring, though.

snue: The `Table 4-2` in the spec describes the handling for Stop Endpoint commands. In most situations we shall advance the dequeue pointer after receiving a doorbell in Stopped state. And we don't do that currently. And our event generation is completely off as well. We won't fix all of that in this PR. 

**Cancelling our request on `STOP` is definitely correct.**

We will currently drop all partial transfers from the cancelled request. But in this case we assume this to be `0` (we are still waiting for new serial data).

**"Transfer TRB (Incomplete) / Residual Length > 0"** is the row in the table.
  * Resuming the TRB on Doorbell does not advance the dequeue pointer.
    * We can re-read the same TRB for now, as we do not remember any partial transfers.
  * We are stopped "within the TRB".
     * I would expect a SetTrDequeuePtr next
  * A Transfer Event with `CompletionCode::Stopped` and the residual length should be generated
    * but we can skip this for now (just make a note maybe?) as skipping NoOps will work just the same.
  * No ISSE (short packet) handling here for now. That needs the EDTLA from #238.

src/device/xhci/endpoint_handle.rs
```diff
@@ -17,7 +17,8 @@ use crate::device::{
         hotplug_endpoint_handle::BaseEndpointHandle,
         interrupter::EventSender,
         real_endpoint_handle::{
-            ControlRequestProcessingResult, InTrbProcessingResult, InTrbProcessingStatus,
+            ControlRequestProcessingResult, InTrbProcessingResult,
+            InTrbProcessingStatus::{self},
```
snue, line 21: nit: unnecessary change

src/device/xhci/endpoint_handle.rs
```diff
@@ -736,6 +737,7 @@ impl<RIEH: RealInEndpointHandle> EndpointHandle for TdBasedInEndpointHandle<RIEH
                 "Submitting on ep {} a real request for {td_request_length} bytes",
                 self.endpoint_id
             );
+
```
snue, line 740: You don't need to touch this file at all anymore. Changes are confined to `nusb`. Please fix the scope of the commit title.

snue (approved):
Two small requests. Please fix them before merging.

Stopping and resuming an endpoint correctly will be done in a separate PR. For now this handling enables the use with serial adapters, and at least it doesn't break the CI tests for other devices.

src/device/xhci/endpoint_handle.rs
snue: To be clear, this is just about the whitespace change in the one commit. You can of course move that change to the correct commit that added the endpoint id to the debug log. Just don't touch that file in the zero-length transfer fix commit.

PR #297, opened 2026-08-06
Fix handling for WorkerMessage::Stop for the CommandRing worker

Noi0103:
A new unit test and a fix discovered when adding the test.

### DIFF
--- a/src/device/xhci/command_ring.rs
+++ b/src/device/xhci/command_ring.rs
@@ -383,7 +383,12 @@ impl CommandWorker {
                 )
             }
             CommandTrbVariant::ForceHeader => todo!(),
-            CommandTrbVariant::NoOp => todo!(),
+            CommandTrbVariant::NoOp => EventTrb::new_command_completion_event_trb(
+                trb.address,
+                0,
+                CompletionCode::Success,
+                0,
+            ),
             CommandTrbVariant::Unrecognized(_, trb_parse_error) => {
                 warn!("Failed to parse command TRB {trb_parse_error:?}");
                 EventTrb::new_command_completion_event_trb(
--- a/src/device/xhci/command_ring.rs
+++ b/src/device/xhci/command_ring.rs
@@ -413,3 +413,307 @@ impl CommandWorker {
         Ok(())
     }
 }
+
+#[cfg(test)]
+mod tests {
+    use tokio::{runtime::Handle, sync::mpsc::UnboundedReceiver};
+
+    use crate::{
+        device::{
+            bus::{testutils::TestBusDevice, BusDevice},
+            pci::constants::xhci::{
+                operational::crcr::{self},
+                rings::trb_types,
+            },
+            xhci::{
+                interrupter::tests::testutils::MockInterrupter,
+                registers::UsbcmdRegister,
+                slot_manager::{test::testutils::MockSlotManager, SlotMessage},
+                trb::testutils::RawTrbBuilder,
+            },
+        },
+        dynamic_bus::DynamicBus,
+    };
+
+    use super::*;
+
+    // We can never set a 64 bit Dequeu Pointers lowest 6 bits:
+    // "This field defines high order bits of the initial value of the 64-bit Command Ring Dequeue Pointer."
+    const FIRST_ADDRESS: u64 = 0x10 << 6;
+    const SECOND_ADDRESS: u64 = FIRST_ADDRESS + 0x10;
+    const THIRD_ADDRESS: u64 = SECOND_ADDRESS + 0x10;
+    const FOURTH_ADDRESS: u64 = THIRD_ADDRESS + 0x10;
+
+    const SLOT_ID: u8 = 0;
+
+    /// the ring is not running
+    /// dequeue_pointer points to FIRST_ADDRESS
+    fn init_test() -> (
+        CommandRing,
+        MockInterrupter,
+        UnboundedReceiver<SlotMessage>,
+        Arc<DynamicBus>,
+        UsbcmdRegister,
+    ) {
+        let dma_bus = Arc::new(DynamicBus::new());
+        let dma_backing = vec![99; 16384];
+        dma_bus
+            .add(0x0, Arc::new(TestBusDevice::new(&dma_backing[..])))
+            .unwrap();
+        let async_runtime = Handle::current();
+
+        let (event_sender, interrupter) = MockInterrupter::new();
+
+        let (slot_manager, receiver) = MockSlotManager::new();
+
+        let usbcmd = UsbcmdRegister::new();
+
+        let command_ring = CommandRing::new(
+            dma_bus.clone(),
+            &async_runtime,
+            event_sender,
+            slot_manager.create_slot_worker_handle(),
+            usbcmd.value_reference(),
+        );
+
+        assert!(interrupter.is_empty());
+        assert!(!command_ring.running.load(Ordering::Relaxed));
+
+        // write a dequeue pointer value so the first command trb is next in line
+        command_ring.control(FIRST_ADDRESS).unwrap();
+
+        (command_ring, interrupter, receiver, dma_bus, usbcmd)
+    }
+
+    #[tokio::test]
+    async fn process_many_command_trb_with_one_doorbell() {
+        let (command_ring, mut interrupter, mut receiver, dma_bus, usbcmd) = init_test();
+
+        // place command trb on a ring segment
+        let command_1 = RawTrbBuilder::new(FIRST_ADDRESS)
+            .with_type(trb_types::NO_OP_COMMAND)
+            .build();
+        let command_2 = RawTrbBuilder::new(SECOND_ADDRESS)
+            .with_type(trb_types::ENABLE_SLOT_COMMAND)
+            .build();
+        let command_3 = RawTrbBuilder::new(THIRD_ADDRESS)
+            .with_type(trb_types::DISABLE_SLOT_COMMAND)
+            .with_byte(15, SLOT_ID)
+            .build();
+        let command_4 = RawTrbBuilder::new(FOURTH_ADDRESS)
+            .with_data_field(0x1 << 4)
+            .with_type(trb_types::ADDRESS_DEVICE_COMMAND)
+            .with_byte(15, SLOT_ID)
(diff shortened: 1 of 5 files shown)


snue:
Can we rename the PR to focus on the actual stop bit handling fix? And make the actual fix a separate commit that explains what is wrong please.

Great that you found this with the suggested unit test additions, but that's less important for the PR.

src/device/xhci/command_ring.rs
```diff
@@ -232,18 +198,12 @@ impl CommandWorker {
...
-                                completion.send(()).ok();
-                                break;
-                            }
                             WorkerMessage::Doorbell => {
                                 // we are already active and running, silently consume
                             }
                             WorkerMessage::Stop => {
                                 self.state = WorkerState::Stopping;
-                                break;
+                                continue 'run_loop;
```
snue, line 206: This should not actually change the logic, given the outer loop doesn't do anything after the match. The `try_next_msg()` match above still uses the "break out of inner loop -> continue outer loop" logic. If we want to be explicit about the `continue` here, we should at least do the same above.

src/device/xhci/command_ring.rs
```diff
@@ -273,6 +233,14 @@ impl CommandWorker {
...
                 WorkerState::Stopping => {
                     self.commandring_running.store(false, Ordering::Relaxed);
+                    let (dequeue_pointer, _) = self.ring.get_dequeue_pointer();
+                    let event = EventTrb::new_command_completion_event_trb(
+                        dequeue_pointer,
+                        0,
+                        CompletionCode::CommandRingStopped,
+                        0,
+                    );
+                    self.event_sender.send(event)?;
```
snue, line 243: So, the actual "fix" is just generating valid Stop Events when stopping the command ring?

src/device/xhci/command_ring.rs
```diff
@@ -400,3 +373,280 @@ impl CommandWorker {
...
+                }
+                EventTrb::CommandCompletion(event_trb)
+                    if event_trb.get_completion_code() == CompletionCode::CommandRingStopped =>
+                {
+                    assert_eq!(command_ring.status() & crcr::CRR, 0);
+                    break;
+                }
+                _ => panic!(),
+            }
+        }
```
snue, line 675: So, what is the expected logic on abort? We should reach a conclusive state "eventually", but we cannot guarantee we hit a specific order of operations? The Comment above should explain what this is looking for. There should not be any more events after the `CommandRingStopped`, but the test does not look for that either, just that we see the stopped event at all.

Can we easily assert there are no more events here?

src/device/xhci/command_ring.rs
```diff
@@ -400,3 +373,280 @@ impl CommandWorker {
...
+        // await above acted as "wait until succeeded"
+        assert_eq!(command_ring.status(), crcr::CRR);
+
+        // command_2
+        assert!(!receiver.is_empty());
+        if let SlotMessage::EnableSlot(sender) = receiver.recv().await.unwrap() {
+            sender.send(Ok(SLOT_ID)).expect("");
+        } else {
+            panic!()
+        }
```
snue, line 504: nit: would probably look better with a `match` instead of the `unwrap else panic`. When the test panics, the message will not be very clear about the expectation.

src/device/xhci/trb.rs
```diff
@@ -126,6 +126,11 @@ impl CommandCompletionEventTrbData {
 
         trb
     }
+
+    #[cfg(test)]
+    pub const fn get_completion_code(&self) -> CompletionCode {
+        self.completion_code
+    }
```
snue, line 133: Is this helper really necessary? Couldn't you `assert_eq` on `expected_event` like for the others?

snue (commented):
Looks like the only thing we were truly missing is the Stop Event generation? I left some comments to clarify some parts. Thanks for adding the unit tests.

src/device/xhci/command_ring.rs
Noi0103: If we get a WorkerMessage::Stop and transition to WorkerState::Stopping and do not continue the outer loop, we would wait for a next trb and while doing so overwrite the WorkerState::Stopping without ever having the chance to react to it.

> we should at least do the same above

Do you mean also labeling the inner loop and using that with the break?

snue: > If we get a WorkerMessage::Stop and transition to WorkerState::Stopping and do not continue the outer loop,

The "break out of the inner loop" is exactly the same as "continue with the outer loop" here, or am I missing something? It's just a bit more explicit.

> Do you mean also labeling the inner loop and using that with the break?

No, that break will also continue with the outer loop. That's what the `break` causes. Leave the inner loop, continue in outer (and since there is no additional code in the outer loop before the next outer iteration, both the break 'inner' and continue 'outer' do the same). You could `continue 'run_loop` there as well. Or just leave it as is in both cases and just add the Event Sending.

Have you confirmed it is still broken with the old `break` logic but the added event sending? I just fail to spot any functional difference in that code.

snue: Ah sorry, the diff view just hid the additional code after the inner loop from me 😕 . The implemented fix is correct, and the break above is correct as well.

src/device/xhci/command_ring.rs
Noi0103: Since the CommandRing has a async worker running in the background I struggle to find a way to guarantee the workers progress state.
So in other words, how to guarantee a expected number of `CompletionCode::Success` will exist until the `CompletionCode::CommandRingStopped` arrives while also guaranteeing that not all of the NoOp will be consumed when the abort bit crcr::CA is actually processed to actually hit the targeted code path.

An assertion is added now.

src/device/xhci/command_ring.rs
```diff
@@ -648,5 +652,7 @@ mod tests {
                 _ => panic!(),
             }
```
snue, line 674: I think I would prefer a nested match for the Completion Code, and not printing the event as debug statement. Instead, just include the event in the error message for the unexpected events. We can keep the completion code test helper for this. It's much more straightforward here than expecting concrete events.

Something like this (untested/unformatted):

```suggestion
            match event {
                EventTrb::CommandCompletion(event_trb) =>
                    match event_trb.get_completion_code() { 
                      CompletionCode::Success =>
                      {
                        assert_eq!(command_ring.status() & crcr::CRR, crcr::CRR);
                      }
                      CompletionCode::CommandRingStopped =>
                      {
                        assert_eq!(command_ring.status() & crcr::CRR, 0);
                        break;
                      }
                unexpected => panic!("received unexpected event {unexpected:?}"),
            }
```

src/device/xhci/command_ring.rs
```diff
@@ -400,3 +373,286 @@ impl CommandWorker {
...
+        // start the ring through usbcmd and doorbell
+        usbcmd.write(usbcmd::RS);
+        command_ring.doorbell().unwrap();
+
+        // verify running state by waiting for the first command completion event
+        if let EventTrb::CommandCompletion(event) = interrupter.await_event().await.unwrap() {
+            assert!(event.get_completion_code() == CompletionCode::Success);
+        } else {
+            panic!()
+        }
```
snue: Making this into `match` statements would allow you to much better show the unexpected results. Same for all the other `else panic!()` statements. Maybe `assert_eq!` on a fully specified expected result is an option here as well.

snue (approved):
Thanks for the fix. Took me a while to spot the actual issue. Maybe instead of saying 'jump to the wrong loop' in the commit message you rather say 'immediately handle the Stopping state when receiving a Stop message instead of checking for more TRBs first'.

I think we should replace the `panic!()`s in the unit tests with more fitting messages that explicitly mention what we got. But that's a minor suggestion, and I am already approving.

src/device/xhci/command_ring.rs
snue: Maybe we can just add an excessive amount of `NoOp` to almost guarantee we didn't handle everything? Can we check the dequeue pointer and verify it's not at the end after stopping the ring?

src/device/xhci/trb.rs
Noi0103: I want this functionality so i can avoid setting the Address in the `write_to_command_abort_bit` test since I do not know when the `CompletionCode::CommandRingStopped` will be handled. This is the same issue I struggle with, that I attempt to explain in the comments in this test.

On (I think) all others i should be fine without this and using the `expected_event` and `assert_eq!`.

src/device/xhci/command_ring.rs
Noi0103: I did not realize how often i used this.

I replaced those "if x or else panic" sections with matches and added error messages.

src/device/xhci/command_ring.rs
Noi0103: > add an excessive amount of NoOp to almost guarantee

I had the same idea but it still would not give any guarantees sadly so I did not follow up on this.

> Can we check the dequeue pointer and verify it's not at the end after stopping the ring?

Even if we were I don't see what use this information has, except for a confirmation that we did not hit the code path we intended.

Noi0103:
The failing USB reset test I could only reproduce locally for usb 1.1, not USB 2.0 like in the CI.

snue:
> The failing USB reset test I could only reproduce locally for usb 1.1, not USB 2.0 like in the CI.

Does it only reproduce on this branch / with the Stop message changes? The guest logs are printed (or attempted) but are just empty?

Noi0103:
I do not suspect this branch is at fault. But I will test locally with a `lsblk` print before expecting `/dev/sda`.

A renovate PR #300 encountered the error on the USB 1.1 test.

snue:
> I do not suspect this branch is at fault. But I will test locally with a `lsblk` print before expecting `/dev/sda`.
> 
I'd rather be interested in seeing the missing guest output in that situation.

> A renovate PR #300 encountered the error on the USB 1.1 test.

Ok, then it looks like a flaky test we should stabilize and not an issue with the stop handling. But just adding `lsblk` won't help much and doesn't explain the issue. Rather focus on obtaining relevant logs.

Noi0103:
Only a rebase on main. This should fail in CI like the merge CI job did.

PR #299, opened 2026-08-07
Refactor nix tests to remove boilerplate systemd config snippets

Noi0103:
While doing #293 a refactor for the boilerplate systemd snippets used in every test came up.
This is removing those boilerplate snippets and will ad an option int the mkUsbTest args to use the alternative file descriptor and systemd sockets setup.
It is currently a todo and will fail on evaluation when used.

### DIFF
--- a/nix/checks/testutils.nix
+++ b/nix/checks/testutils.nix
@@ -1,6 +1,8 @@
 {
   lib,
   pkgs,
+  cloud-hypervisor,
+  usbvfiod,
   ...
 }:
 let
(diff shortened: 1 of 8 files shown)


nix/checks/testutils.nix
```diff
@@ -266,6 +268,55 @@ let
...
+                    --user-device socket=${usbvfiodSocket} \
+                    --net "tap=tap0,mac=,ip=192.168.100.1,mask=255.255.255.0"
+                '';
+              };
+            };
+        };
+      }
+    else
+      assert false; # not yet implemented
+      { };
```
snue: Use `assertMsg` to throw a helpful comment when attempted.

nix/checks/testutils.nix
```diff
@@ -400,7 +452,8 @@ let
       nodes.machine = _: {
         imports = [
           basicMachineConfig
-          (systemd args)
+          #(systemd args)
```
snue: remnant? can probably be removed.

snue (commented):
Pretty nice boilerplate reduction. Left some minor comments.

nix/checks/testutils.nix
Noi0103: ah, that is nicer than showing the line of the assert with the comment included

nix/checks/testutils.nix
```diff
@@ -266,6 +268,54 @@ let
...
+                    --initramfs ${netboot.initrd} \
+                    --user-device socket=${usbvfiodSocket} \
+                    --net "tap=tap0,mac=,ip=192.168.100.1,mask=255.255.255.0"
+                '';
+              };
+            };
+        };
+      }
+    else
+      lib.assertMsg false "not yet implemented" { };
```
snue: Let's be specific about what caused the failure, so the caller has a better chance to repair it without looking into this code. `useFileDescriptor = true is not yet implemented for the usbvfiod service setup`

snue (approved):
nit: you have a typo in the commit title `s/snipped/snippet`. You may fix that before merge,

Noi0103:
Thanks, all done.

PR #302, opened 2026-08-10
Fix device reattachment and add test

ziyifu225:
Add an integration test for [PR#273](https://github.com/cyberus-technology/usbvfiod/pull/273).
The device should be detached before re-attaching the same instance.

I found some weird logs in the CI test:
```
Aug 07 13:26:22 machine usbvfiod[614]: 2026-08-07T13:26:22.522987Z  INFO usbvfiod::device::xhci::port: A device with the same identifier is already attached and will be detached first
Aug 07 13:26:22 machine usbvfiod[614]: 2026-08-07T13:26:22.523016Z DEBUG usbvfiod::device::xhci::port: Device to detach is connected to port 1
Aug 07 13:26:22 machine usbvfiod[614]: 2026-08-07T13:26:22.523194Z  INFO usbvfiod::device::xhci::port: Detached device (5, 2) from port 1
Aug 07 13:26:22 machine usbvfiod[614]: 2026-08-07T13:26:22.523227Z  INFO usbvfiod::device::xhci::port: Attached SuperSpeed (5 Gbps) device (5, 2) to port 1 (USB3 port)
Aug 07 13:26:22 machine usbvfiod[614]: 2026-08-07T13:26:22.523277Z DEBUG usbvfiod::device::xhci::port: Device to detach is connected to port 1
Aug 07 13:26:22 machine usbvfiod[614]: 2026-08-07T13:26:22.523457Z DEBUG nusb::platform::linux_usbfs::device: Released interface 0 on device 0: Ok(())
Aug 07 13:26:22 machine usbvfiod[614]: 2026-08-07T13:26:22.531922Z DEBUG nusb::platform::linux_usbfs::device: Reattached kernel drivers for interface 0 on device 0: Ok(())
Aug 07 13:26:22 machine usbvfiod[614]: 2026-08-07T13:26:22.531967Z DEBUG nusb::platform::linux_usbfs::device: Closing device 0
Aug 07 13:26:22 machine usbvfiod[614]: 2026-08-07T13:26:22.532078Z  INFO usbvfiod::device::xhci::port: Detached device (5, 2) from port 1
Aug 07 13:26:22 machine usbvfiod[614]: 2026-08-07T13:26:22.532146Z DEBUG usbvfiod::device::xhci::interrupter: Sent event: PortStatusChange(PortStatusChangeEventTrbData { port_id: 1 })
Aug 07 13:26:22 machine usbvfiod[614]: 2026-08-07T13:26:22.532187Z DEBUG usbvfiod::device::xhci::interrupter: Sent event: PortStatusChange(PortStatusChangeEventTrbData { port_id: 1 })
Aug 07 13:26:22 machine usbvfiod[614]: 2026-08-07T13:26:22.532226Z DEBUG usbvfiod::device::xhci::interrupter: Sent event: PortStatusChange(PortStatusChangeEventTrbData { port_id: 1 })
Aug 07 13:26:22 machine usbvfiod[614]: 2026-08-07T13:26:22.532269Z DEBUG usbvfiod::device::xhci::port: Could not find the device to detach
Aug 07 13:26:22 machine usbvfiod[614]: 2026-08-07T13:26:22.532603Z DEBUG nusb::platform::linux_usbfs::device: Released interface 0 on device 1: Err(Os { code: 22, kind: InvalidInput, message: "Invalid argument" })
Aug 07 13:26:22 machine usbvfiod[614]: 2026-08-07T13:26:22.532661Z DEBUG nusb::platform::linux_usbfs::device: Closing device 1
Aug 07 13:26:22 machine usbvfiod[614]: 2026-08-07T13:26:22.574709Z DEBUG usbvfiod::device::xhci::command_ring: Doorbell for the controller
Aug 07 13:26:22 machine usbvfiod[614]: 2026-08-07T13:26:22.574841Z DEBUG usbvfiod::device::xhci::command_ring: Processing command CommandTrb { address: 1988386928, variant: ConfigureEndpoint(ConfigureEndpointCommandTrbData { input_context_pointer: 1989033984, deconfigure: false, slot_id: 1 }) }
```
While adding the reattach integration test, I found that reattach initially follows the expected path: the controller detects the existing device, detaches it, and attaches the new instance successfully.

The explicit detach cancels the old device's detach token, which wakes its asynchronous detach listener. The listener then sends another detach request using only the device identifier. Because the new instance has the same identifier, this delayed request matches and detaches the newly attached device, causing it to disappear from the guest.

This change adds an attach-lifecycle ID to each attached device. The ID is based on the device's `CancellationToken`, which is created once per device instance and shared through clones.

Detach requests from the asynchronous listener now carry this lifecycle ID. Device lookup matches both the device identifier and lifecycle ID when one is provided, so a delayed detach from an old instance no longer matches a reattached instance. Explicit detach and the detach performed during reattach continue to match only by device identifier.

### DIFF
--- a/src/device/xhci/port.rs
+++ b/src/device/xhci/port.rs
@@ -103,7 +103,7 @@ impl<CRD: CompleteRealDevice> PortArray<CRD> {
 
 #[derive(Debug)]
 struct PortWorker<CRD: CompleteRealDevice> {
-    devices: OneIndexed<Option<Arc<CRD>>, { MAX_PORTS as usize }>,
+    devices: OneIndexed<Option<AttachedDevice<CRD>>, { MAX_PORTS as usize }>,
     portsc: Arc<OneIndexed<PortscRegister, { MAX_PORTS as usize }>>,
     event_sender: EventSender,
     // the worker does not use the sender itself but needs to pass clones of the sender to detach listeners
@@ -112,10 +112,23 @@ struct PortWorker<CRD: CompleteRealDevice> {
     async_runtime: runtime::Handle,
 }
 
+/// The detach token is created once per device instance and cloned by its users, so its identity
+/// distinguishes attached device lifecycles. Token cancellation does not change this identity.
+#[derive(Debug, Clone, PartialEq, Eq)]
+pub struct DeviceInstanceId(CancellationToken);
+
+#[derive(Debug)]
+struct AttachedDevice<CRD: CompleteRealDevice> {
+    device: Arc<CRD>,
+    // The lifecycle ID of this attached device instance.
+    instance_id: DeviceInstanceId,
+}
+
 #[derive(Debug)]
 pub enum PortMessage<CRD: CompleteRealDevice> {
     Attach(CRD, oneshot::Sender<Response>),
-    Detach(CRD::ID, oneshot::Sender<Response>),
+    // Detaches a device, optionally limited to a specific attached device lifecycle.
+    Detach(CRD::ID, Option<DeviceInstanceId>, oneshot::Sender<Response>),
     ListAttached(oneshot::Sender<Vec<CRD::ID>>),
     // port id
     GetDevice(usize, oneshot::Sender<Option<Arc<CRD>>>),
@@ -136,8 +149,8 @@ impl<CRD: CompleteRealDevice> PortWorker<CRD> {
                 PortMessage::Attach(device, responder) => {
                     responder.send_anyhow(self.attach(device)?)?;
                 }
-                PortMessage::Detach(identifier, responder) => {
-                    responder.send_anyhow(self.detach(identifier)?)?;
+                PortMessage::Detach(identifier, instance_id, responder) => {
+                    responder.send_anyhow(self.detach(identifier, instance_id)?)?;
                 }
                 PortMessage::ListAttached(responder) => {
                     responder.send_anyhow(self.attached_devices())?;
@@ -146,7 +159,7 @@ impl<CRD: CompleteRealDevice> PortWorker<CRD> {
                     let device = self
                         .devices
                         .get(port_id)
-                        .and_then(|opt| opt.as_ref().map(|dev| dev.clone()));
+                        .and_then(|opt| opt.as_ref().map(|attached| attached.device.clone()));
                     responder.send_anyhow(device)?;
                 }
             };
@@ -165,7 +178,7 @@ impl<CRD: CompleteRealDevice> PortWorker<CRD> {
             info!(
                 "A device with the same identifier is already attached and will be detached first"
             );
-            self.detach(device.identifier())?;
+            self.detach(device.identifier(), None)?;
         }
 
         let speed = match device.realdevice_ref().speed() {
@@ -185,13 +198,18 @@ impl<CRD: CompleteRealDevice> PortWorker<CRD> {
                 };
 
         let identifier = device.identifier();
+        let instance_id = DeviceInstanceId(device.detach_token());
         self.async_runtime.spawn(detach_listener(
             device.detach_token(),
             identifier,
+            instance_id.clone(),
             self.msg_sender.clone(),
         ));
 
-        self.devices[available_port_id] = Some(Arc::new(device));
+        self.devices[available_port_id] = Some(AttachedDevice {
+            device: Arc::new(device),
+            instance_id,
+        });
 
         let new_portsc = match version {
             UsbVersion::USB3 => {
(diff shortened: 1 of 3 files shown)


snue:
Adding the test and uncovering the flawed logic. 

> My current suspicion is that the old device's detach listener is triggered asynchronously after the explicit detach.

This could indeed happen. How do we fix this cleanly?

src/device/xhci/port.rs
```diff
@@ -259,13 +281,24 @@ impl<CRD: CompleteRealDevice> PortWorker<CRD> {
             }
         };
 
+        if requested_instance_id
+            .as_ref()
+            .is_some_and(|requested_instance_id| {
+                requested_instance_id != &self.devices[port_id].as_ref().unwrap().instance_id
```
snue: This needs a safety comment for the `unwrap()` (we have one below).

src/device/xhci/port.rs
```diff
@@ -227,16 +242,23 @@ impl<CRD: CompleteRealDevice> PortWorker<CRD> {
...
+            .map(|attached| attached.device.identifier())
             .collect()
     }
 
-    fn detach(&mut self, id: CRD::ID) -> anyhow::Result<Response> {
+    fn detach(
+        &mut self,
+        id: CRD::ID,
+        requested_instance_id: Option<DeviceInstanceId>,
+    ) -> anyhow::Result<Response> {
```
snue, line 261: The file is pretty sparse on doc comments currently. We should explain the logic behind the Option in the Port Message and for the detach function, as well as the instance id in the `AttachedDevice` struct.

src/device/xhci/port.rs
```diff
@@ -182,16 +191,22 @@ impl<CRD: CompleteRealDevice> PortWorker<CRD> {
                 {
                     Some(port) => port,
                     None => return Ok(Response::NoFreePort),
-                };
+        };
```
snue: nit: did auto-formatting change this indentation?

src/device/xhci/port.rs
```diff
@@ -112,10 +112,19 @@ struct PortWorker<CRD: CompleteRealDevice> {
     async_runtime: runtime::Handle,
 }
 
+#[derive(Debug, Clone, PartialEq, Eq)]
+pub struct DeviceInstanceId(CancellationToken);
```
snue, line 118: What do these tokens look like? Do they change once `cancelled().await` was called? You only use that as an implicit "generation ID" and not for cancelling anything, right? This might be fine, but I want to be sure it is actually sane. And I would appreciate comments that explain the intention behind it.

src/device/xhci/port.rs
```diff
@@ -112,10 +112,19 @@ struct PortWorker<CRD: CompleteRealDevice> {
...
+
+#[derive(Debug)]
+struct AttachedDevice<CRD: CompleteRealDevice> {
+    device: Arc<CRD>,
+    instance_id: DeviceInstanceId,
+}
```
snue, line 125: I like the approach of keeping the generation/instance ID to the port implementation and not touching the `CRD::ID`.

src/device/xhci/port.rs
```diff
@@ -182,16 +191,22 @@ impl<CRD: CompleteRealDevice> PortWorker<CRD> {
...
                     None => return Ok(Response::NoFreePort),
-                };
+        };
 
         let identifier = device.identifier();
+        let cancel = device.detach_token();
+        let instance_id = DeviceInstanceId(cancel.clone());
         self.async_runtime.spawn(detach_listener(
-            device.detach_token(),
+            cancel,
```
snue: nit: The cancellation token doubles as instance ID here. Do we really need to pass it twice, or can we just extract the token in the listener? That obfuscates its use a bit of course.

src/device/xhci/port.rs
```diff
@@ -259,13 +281,24 @@ impl<CRD: CompleteRealDevice> PortWorker<CRD> {
             }
         };
 
+        if requested_instance_id
+            .as_ref()
+            .is_some_and(|requested_instance_id| {
```
snue: nit: maybe this could be part of the above `.filter_map()` chain already. That would make the above comment about "This message is expected once per soft detach" actually cover this soft re-attach case as well.

snue (commented):
Nice. 🎉  Left some comments. Does it fix the new integration test reliably?

Is there anything missing for adding a unit test as well?

snue:
Please adapt the PR title (something like "Fix device reattachment and add test"). Then I think this can move out of Draft with the next iteration.

src/device/xhci/port.rs
ziyifu225: Each `CompleteRealDeviceImpl::new()` creates a fresh CancellationToken. The port and its detach listener store clones of that token as the instance ID. `cancelled().await` only waits for cancellation; equality remains based on the same underlying token allocation.
After reattach, the old listener carries token A while the new device carries token B, so the instance-ID comparison rejects the old detach request.

src/device/xhci/port.rs
ziyifu225: Good idea. The previous logic had too many separate checks, now I merged them into a more centralized one.

src/device/xhci/port.rs
ziyifu225: cancel is used to wait for cancellation, while instance_id is forwarded for instance matching. So it's sure that extracting the token only inside the listener would make this relationship less clear.

src/device/xhci/port.rs
ziyifu225: yes... maybe😅 I'm not sure why it happens. But I fix it now.

snue (approved):
Nice that this is fully contained within `port.rs`. I also really like the added comments and latest code structure. Thanks!

You can merge this after squashing the fixups.

PR #303, opened 2026-08-12
Allow some consecutive Link TRB and add tests for it

Noi0103:
A very simple test to check that the get and set methods on the linked ring work as intended.

### DIFF
--- a/src/device/xhci/linked_ring.rs
+++ b/src/device/xhci/linked_ring.rs
@@ -122,6 +122,9 @@ mod tests {
 
     use super::*;
 
+    const FIRST_ADDRESS: u64 = 0x00;
+    const THIRD_ADDRESS: u64 = 0x20;
+
     // test summary:
     //
     // This test checks the retrieval of raw TRBs according to the cycle state and cycle bits,
@@ -249,5 +252,28 @@ mod tests {
         expected_data[12] &= 0xfe;
 
         assert_eq!(actual_data, expected_data);
+    }
+
+    #[test]
+    fn get_and_set_dequeue_pointer_with_cycle_bit() {
+        // construct memory segment for a ring that can contain TRBs
+        let dma = Arc::new(TestBusDevice::new(&[0; TRB_SIZE * 32]));
+        let mut ring = LinkedRing::new(dma, THIRD_ADDRESS, true);
+
+        // the dequeue pointer is initialized on some address in the middle
+        // with a cycle bit preventing `next_trb()` to see any trbs
+        assert_eq!(ring.get_dequeue_pointer(), (THIRD_ADDRESS, true));
+        assert_eq!(ring.next_trb(), None);
+
+        // setting the pointer backwards and changing the cycle bit interpretation
+        // so we interpret the zero buffer as trb from the current cycle
+        ring.set_dequeue_pointer(FIRST_ADDRESS, false);
+        assert_eq!(ring.get_dequeue_pointer(), (FIRST_ADDRESS, false));
+        check_trb(ring.next_trb(), FIRST_ADDRESS, [0; 16]);
+
+        // with the cycle bit we can see the previously none trb
+        ring.set_dequeue_pointer(THIRD_ADDRESS, false);
+        assert_eq!(ring.get_dequeue_pointer(), (THIRD_ADDRESS, false));
+        check_trb(ring.next_trb(), THIRD_ADDRESS, [0; 16]);
     }
 }
--- a/src/device/xhci/linked_ring.rs
+++ b/src/device/xhci/linked_ring.rs
@@ -381,4 +381,85 @@ mod tests {
         assert_eq!(ring.get_dequeue_pointer(), (SEG_3, false));
         check_trb(trb, SEG_3, [0; 16]);
     }
+
+    // This is technically allowed by the specification. We don't want to allow
+    // endless Link TRB since the guest would be able to block the hosts resources.
+    #[test]
+    fn gracefully_handle_endless_link_td() {
+        let dma = Arc::new(TestBusDevice::new(&[0; TRB_SIZE * 2]));
+        let mut ring = LinkedRing::new(dma.clone(), FIRST_ADDRESS, false);
+
+        // place a link trb that create an endless loop on itself
+        let link = RawTrbBuilder::new(FIRST_ADDRESS)
+            .with_data_pointer(FIRST_ADDRESS)
+            .with_trb_type(trb_types::LINK)
+            .build();
+        dma.write_bulk(link.address, &link.buffer);
+
+        for _ in 0..9 {
+            // LinkTrb TD --> LinkTrb TD --> ... --(hit some limit)--> returned None
+            let trb = ring.next_trb();
+            assert_eq!(ring.get_dequeue_pointer(), (FIRST_ADDRESS, false));
+            assert_eq!(trb, None);
+        }
+    }
+
+    // This is technically not allowed by the specification. We allow this for
+    // the sake of simplicity and handle it the same as we handle the endless
+    // Link TD.
+    #[test]
+    fn gracefully_handle_endless_chained_link_trb() {
+        let dma = Arc::new(TestBusDevice::new(&[0; TRB_SIZE * 2]));
+        let mut ring = LinkedRing::new(dma.clone(), FIRST_ADDRESS, false);
+
+        // place a link trb that create an endless loop on itself
+        let link = RawTrbBuilder::new(FIRST_ADDRESS)
+            .with_data_pointer(FIRST_ADDRESS)
+            .with_trb_type(trb_types::LINK)
+            .with_chain()
+            .build();
+        dma.write_bulk(link.address, &link.buffer);
+
+        for _ in 0..9 {
+            // LinkTrb(chain) --> LinkTrb(chain) --> ... --(hit some limit)--> returned None
+            let trb = ring.next_trb();
+            assert_eq!(ring.get_dequeue_pointer(), (FIRST_ADDRESS, false));
+            assert_eq!(trb, None);
+        }
+    }
+
+    #[test]
+    fn gracefully_handle_endless_link_trb_with_flipping_cycle_bit() {
+        let dma = Arc::new(TestBusDevice::new(&[0; TRB_SIZE * 4]));
+        let mut ring = LinkedRing::new(dma.clone(), FIRST_ADDRESS, false);
+
+        // cycle is false when encountering this link trb
+        let link = RawTrbBuilder::new(FIRST_ADDRESS)
+            .with_data_pointer(THIRD_ADDRESS)
+            .with_trb_type(trb_types::LINK)
+            .with_toggle_cycle()
+            .build();
+        dma.write_bulk(link.address, &link.buffer);
+
+        // cycle is true when encountering this link trb
+        let link_cycle = RawTrbBuilder::new(THIRD_ADDRESS)
+            .with_data_pointer(FIRST_ADDRESS)
+            .with_trb_type(trb_types::LINK)
+            .with_cycle()
+            .with_toggle_cycle()
+            .build();
+        dma.write_bulk(link_cycle.address, &link_cycle.buffer);
+
+        for _ in 0..9 {
+            // LinkTrb(!cycle) --> LinkTrb(cycle) --> LinkTrb(!cycle) --> ... --(hit some limit)--> returned None
+            let trb = ring.next_trb();
+            assert_eq!(trb, None);
+
+            let dequeue_pointer = ring.get_dequeue_pointer();
+            assert!(
+                dequeue_pointer == (FIRST_ADDRESS, false)
+                    || dequeue_pointer == (THIRD_ADDRESS, true)
+            );
+        }
+    }
 }
--- a/src/device/xhci/linked_ring.rs
+++ b/src/device/xhci/linked_ring.rs
@@ -153,7 +153,7 @@ mod tests {
             0x0, 0x0, 0x0, 0x0, 0x0, 0x0, 0x0, 0x0, 0x0, 0x0, 0x0, 0x0, 0x2, 0x18, 0x0, 0x0,
         ];
 
-        // construct memory segment for a ring that can contain 5 TRBs and an endpoint context
+        // construct memory segment for a ring that can contain 5 TRBs
         let ram = Arc::new(TestBusDevice::new(&[0; TRB_SIZE * 5]));
         let mut ring = LinkedRing::new(ram.clone(), 0x0, true);
 
@@ -164,17 +164,16 @@ mod tests {
             "When no fresh TRB is on the ring, next_trb should return None, instead got: {trb:?}"
         );
 
-        // place three TRBs
-        // set cycle bit
-        // place setup
+        // place three TRBs and set their cycle bit
+        // setup
         ram.write_bulk(0, &setup);
         ram.write_bulk(12, &[0x1]);
 
-        // place data
+        // data
         ram.write_bulk(TRB_SIZE as u64, &data);
         ram.write_bulk(TRB_SIZE as u64 + 12, &[0x1]);
 
-        // place status
+        // status
         ram.write_bulk(TRB_SIZE as u64 * 2, &status);
         ram.write_bulk(TRB_SIZE as u64 * 2 + 12, &[0x1]);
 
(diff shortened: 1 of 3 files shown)


src/device/xhci/linked_ring.rs
```diff
@@ -251,4 +251,60 @@ mod tests {
...
+        const DMA_POINTER_1: u64 = 0x200;
+
+        const SETUP_WLENGTH: u16 = 512;
+        const TRANSFER_LENGTH: u32 = SETUP_WLENGTH as u32;
+
+        const TRB_TYPE_SETUP_STAGE: u8 = 0x2;
+        const TRB_TYPE_DATA_STAGE: u8 = 0x3;
+        const TRB_TYPE_STATUS_STAGE: u8 = 0x4;
+
+        const SETUP_BM_REQUEST_TYPE_IN: u8 = 0x80;
```
snue: The types and contents really don't matter in this test, right? Let's just use the simplest possible types, with some way to discern the ones you are interested in checking.

src/device/xhci/linked_ring.rs
```diff
@@ -251,4 +251,36 @@ mod tests {
...
+        assert_eq!(ring.get_dequeue_pointer(), (FIRST_ADDRESS, false));
+
+        assert_eq!(ring.next_trb(), Some(first));
+
+        // skipping trb on SECOND_ADDRESS
+        ring.set_dequeue_pointer(THIRD_ADDRESS, false);
+
+        let next = ring.next_trb();
+        assert_ne!(next, Some(second));
+        assert_eq!(next, Some(third));
```
snue: Calling `next_trb` does not advance the ring. We would never expect the second TRB here. Rather it would still point to the first without explicit `advance()` (or setting the dequeue pointer of course)

Noi0103: Ah, thanks, I did not look into the implementation for `next_trb()` and assumed it would advance for me (to the next trb after requesting the latest). I will change that.

src/device/xhci/linked_ring.rs
```diff
@@ -251,4 +251,33 @@ mod tests {
...
+    fn get_and_set_dequeue_pointer_to_skip_trb() {
+        const DMA_SEGMENT_LENGTH: usize = TRB_SIZE * 4;
+
+        const FIRST_ADDRESS: u64 = 0x10;
+        const SECOND_ADDRESS: u64 = 0x20;
+        const THIRD_ADDRESS: u64 = 0x30;
+
+        let first = RawTrbBuilder::new(FIRST_ADDRESS).build();
+        let second = RawTrbBuilder::new(SECOND_ADDRESS).build();
+        let third = RawTrbBuilder::new(THIRD_ADDRESS).build();
```
snue: You don't really need the builder. You can use zeroed buffers in the dma writes and the `check_trb` function used in the other test.

You may not even need any TRB definitions or DMA writes at all. Only the first and third address.

src/device/xhci/linked_ring.rs
```diff
@@ -251,4 +251,33 @@ mod tests {
...
+
+        // construct memory segment for a ring that can contain 5 TRBs and an endpoint context
```
snue: That comment looks wrong. There is no endpoint context here? (And neither in the other test this came from). You also changed the constant to 4 TRBs. You don't need the constant if you have a comment explaining the computation and only use it once.

src/device/xhci/linked_ring.rs
snue: The existing unit test is a good example of how it is supposed to work :wink:

src/device/xhci/linked_ring.rs
Noi0103: I found the `assert_eq!()` somewhat easier to read with the RawTrbBuilder since the `ring.next_trb()` returns a `Option<RawTrb>`.

src/device/xhci/linked_ring.rs
```diff
@@ -251,4 +250,123 @@ mod tests {
...
+
+        assert_eq!(ring.next_trb(), None);
+
+        // since we did not encounter a actionable link trb we remained at this address
+        assert_eq!(ring.get_dequeue_pointer(), (SEG_1 + LINK_TRB_ADDRESS, true));
+    }
+
+    // Encountering multiple chained LinkTrb is described in the xhci specification as:
+    // - undefined behavior (chapter 4.11.7) and
+    // - performance degrading effects (chapter 6.4.4.1).
```
snue: Maybe add a final statement what our implementation is expected to do. We would like it to just work now.

However, in case the driver gives us a cycle of `LinkTrb`s, we will run into a busy endless loop. This is probably already true for a single `LinkTrb` that references itself. We are plainly within the "performance degrading effects" here, but we should not allow guests to hog host resources. Robust cycle detection is not trivial, but an upper limit on chained `LinkTrb`s is probably good enough. When the limit is hit, we should just return `None` (and print a warning of course). The current limit is 1 (no chaining allowed), but it panics, which is not ideal.

src/device/xhci/linked_ring.rs
```diff
@@ -251,4 +250,123 @@ mod tests {
...
+        check_trb(ring.next_trb(), FIRST_ADDRESS, [0; 16]);
+
+        // with the cycle bit we can see the previously none trb
+        ring.set_dequeue_pointer(THIRD_ADDRESS, false);
+        assert_eq!(ring.get_dequeue_pointer(), (THIRD_ADDRESS, false));
+        check_trb(ring.next_trb(), THIRD_ADDRESS, [0; 16]);
+    }
+
+    #[test]
+    fn next_trb_changing_dequeue_pointer_value_when_encountering_link_trb() {
```
snue: nit: can be a little shorter. you can also keep it as is.

```suggestion
    fn next_trb_advances_dequeue_pointer_over_link_trb() {
```

src/device/xhci/linked_ring.rs
snue: This comment needs updating when we change the logic to allow more than one `LinkTrb`.

snue (commented):
Unfortunately, we can't trust the guest to play by the rules. Thanks for the added tests and removing the panic. But we need to have bounds on the `LinkTrb` loop - and test them as well :wink:

src/device/xhci/linked_ring.rs
Noi0103: > Maybe add a final statement what our implementation is expected to do.

I added some more description to the `next_trb()` doc comment and made 10 (arbitrarily chosen) the upper limit.

src/device/xhci/linked_ring.rs
```diff
@@ -38,35 +38,37 @@ impl LinkedRing {
     ///
     /// This function only returns `TransferTrb`s that are not Link TRBs.
     /// Instead, Link TRBs are handled correctly, which is the reason why the
-    /// function might read two TRBs to return a single one.
+    /// function might read multiple TRBs to return a single one.
+    ///
+    /// There is an upper limit on how many Link TRB are tolerated before this
+    /// function will return None instead of following more Link TRB.
```
snue: It should also mention the effect on the dequeue pointer. It may end up anywhere in the linked structure with unknown cycle state. We do not save/restore it. The spec allows this ("undefined behavior"). The guest driver will have to recover, but that is fine. The comment should mention that the "upper limit" demanded by the spec is 1.

src/device/xhci/linked_ring.rs
```diff
@@ -38,35 +38,37 @@ impl LinkedRing {
...
     pub fn next_trb(&mut self) -> Option<RawTrb> {
         // retrieve TRB at dequeue pointer and return None if there is no fresh
         // TRB
-        let first_trb_buffer = self.next_trb_raw()?;
+        let mut buffer = self.next_trb_raw()?;
+
+        let mut link_trb_upper_limit = 10;
```
snue: It's an arbitrary limit. Given the spec says it is not allowed, I would have chosen a small multiple of 2 (2, 4, or 8)

src/device/xhci/linked_ring.rs
```diff
@@ -38,35 +38,37 @@ impl LinkedRing {
...
+        while let Some(link_trb) = LinkTrb::parse(buffer) {
+            if link_trb_upper_limit == 0 {
+                warn!("encountered too many chained Link Trb (undefined behaviour)");
```
snue: I think any chained link TRBs should cause a warning, as they are never spec compliant. Exceeding the limit should then return `None`. And let's not say "undefined behavior" in the logs - it sounds like bad things can happen in our implementation when we hit this. The behavior is well-defined for our controller emulation. Just the guest cannot rely on anything per spec.

src/device/xhci/linked_ring.rs
```diff
@@ -38,35 +38,37 @@ impl LinkedRing {
...
+    /// function will return None instead of following more Link TRB.
     pub fn next_trb(&mut self) -> Option<RawTrb> {
         // retrieve TRB at dequeue pointer and return None if there is no fresh
         // TRB
-        let first_trb_buffer = self.next_trb_raw()?;
+        let mut buffer = self.next_trb_raw()?;
+
+        let mut link_trb_upper_limit = 10;
+        while let Some(link_trb) = LinkTrb::parse(buffer) {
+            if link_trb_upper_limit == 0 {
```
snue: nit: I'd rather start from 0 and count up. Should be a little more efficient when there is no Link TRB. And Link TRB is rather the exception.

src/device/xhci/linked_ring.rs
```diff
@@ -251,4 +260,143 @@ mod tests {
...
+    }
+
+    #[test]
+    fn encounter_endless_link_trb() {
+        const FIRST_ADDRESS: u64 = 0x00;
+
+        let dma = Arc::new(TestBusDevice::new(&[0; TRB_SIZE * 8]));
+        let mut ring = LinkedRing::new(dma.clone(), FIRST_ADDRESS, false);
+
+        // place a link trb that create an endless loop on itself
```
snue, line 396: This is the simplest possible loop. I'd like an additional test that also toggles the cycle state around (which needs multiple Link TRBs to end up in a valid loop).

Such a test can then of course not rely on the final dequeue pointer without knowing details of the implementation (which it shouldn't). It's enough to see we don't get stuck and don't read valid TRBs from that. As we may stop anywhere on the cycle, we should also check more than once.

src/device/xhci/linked_ring.rs
```diff
@@ -251,4 +260,143 @@ mod tests {
...
+        dma.write_bulk(link.address, &link.buffer);
+
+        // LinkTrb -> LinkTrb -> zero buffer
+        let trb = ring.next_trb();
+        assert_eq!(ring.get_dequeue_pointer(), (SEG_3, false));
+        check_trb(trb, SEG_3, [0; 16]);
+    }
+
+    #[test]
+    fn encounter_endless_link_trb() {
```
snue: The test name should really say what the expectation is. Something like `gracefully_handle_link_trb_cycle` and `gracefully_handle_complex_link_trb_cycles`.

src/device/xhci/linked_ring.rs
snue: The test still does not mention the expectation. It's probably best to just give the test a fitting name like `some_chained_link_trbs_are_supported`.

src/device/xhci/linked_ring.rs
```diff
@@ -153,7 +163,7 @@ mod tests {
             0x0, 0x0, 0x0, 0x0, 0x0, 0x0, 0x0, 0x0, 0x0, 0x0, 0x0, 0x0, 0x2, 0x18, 0x0, 0x0,
         ];
 
-        // construct memory segment for a ring that can contain 5 TRBs and an endpoint context
+        // construct memory segment for a ring that can contain 5 TRBs
```
snue, line 180: Thanks for fixing this comment. The other comment changes in this test look a bit noisy without obvious improvement to me, but you can keep it as is.

snue (commented):
With the fixed chained Link TRB behavior, please change the PR title to reflect that. A title like the current one will surely be missed when generating high-level release descriptions, when the change is actually noteworthy.

src/device/xhci/linked_ring.rs
```diff
@@ -38,35 +38,45 @@ impl LinkedRing {
     ///
     /// This function only returns `TransferTrb`s that are not Link TRBs.
     /// Instead, Link TRBs are handled correctly, which is the reason why the
-    /// function might read two TRBs to return a single one.
+    /// function might read multiple TRBs to return a single one. When a Link
+    /// TRB is handled the dequeue pointer is set to the Link TRB's data field
+    /// value.
+    ///
+    /// Encountering multiple chained LinkTrb is described in the xhci specification as:
+    /// - allowed in the form of Link TD (chapter 4.11.7)
```
snue: We need to be a bit careful with "chained" here. We do not explicitly look at the chain bits. The right word is "consecutive". "Chained" Link TRBs are not allowed. But "stand-alone" Link-TD (chain bit not set) is allowed. In our handling we do not check the chain bit (and we shouldn't). We just follow any LinkTRBs we find, until we detect too many consecutive links.

src/device/xhci/linked_ring.rs
```diff
@@ -251,4 +268,175 @@ mod tests {
...
+            .with_tc()
+            .build();
+        dma.write_bulk(link.address, &link.buffer);
+
+        // cycle is true when encountering this link trb
+        let link_cycle = RawTrbBuilder::new(THIRD_ADDRESS)
+            .with_data_field(FIRST_ADDRESS)
+            .with_type(TRB_TYPE_LINK)
+            .with_c()
+            .with_tc()
```
snue: Some of the methods on this builder are not very descriptive. This is "cycle" and "toggle cycle"? There is also an explicit chain bit (`with_ch()`?) - that we never set in these tests.

src/device/xhci/linked_ring.rs
Noi0103: > we should also check more than once.

What exactly do you mean by that? A current test loops through two Link TRB and could stop on either one of them.

snue: I mean you receive `None` on `next_trb` and the dequeue pointer ends "somewhere". You can call `next_trb` again a couple of times to ensure it stays that way, and doesn't jump somewhere unexpected where it then finds a TRB.

src/device/xhci/linked_ring.rs
```diff
@@ -38,35 +38,45 @@ impl LinkedRing {
     ///
     /// This function only returns `TransferTrb`s that are not Link TRBs.
     /// Instead, Link TRBs are handled correctly, which is the reason why the
```
snue, line 40: Unfortunately, on non-command-rings, a Link TRB can have IOC set, which is currently not handled.

src/device/xhci/linked_ring.rs
Noi0103: The short names are taken as is from the specification tables in chapter 6.4 since I found the long names too long at times and did not want to introduce a new alias. For writing these, the doc comments for the functions were sufficient, but I can see how reading without having the doc comments is no fun. I would like to have a consistent schema, should those all use the long names instead? Something like `.with_interrupt_on_completion()` would probably be on the longer side of those names.

> we do not check the chain bit (and we shouldn't).

I will add the chain bit to one of the loops to have it somewhere in the tests, even if it is not meant make a difference.

snue: I see how some of the names can become rather long. Personally I would still prefer if they were self-describing.

src/device/xhci/linked_ring.rs
```diff
@@ -38,35 +38,49 @@ impl LinkedRing {
...
+        let mut link_trb_counter = 0;
+        while let Some(link_trb) = LinkTrb::parse(buffer) {
+            if link_trb_counter == 256 {
+                warn!("encountered unreasonable amount of consecutive Link TRB");
+                return None;
+            }
+            link_trb_counter += 1;
+
+            if link_trb.interrupt_on_completion {
+                warn!("Link TRB with set IOC bit is being ignored");
```
snue: nit:

```suggestion
                warn!("ignoring the set IOC bit on a Link TRB");
```

snue (approved):
Making sure the names match the spec is less important to me than just making sure they are sufficiently clear and self-describing (e.g. `with_type(trb_types::Something)` on the `RawTrbBuilder` doesn't really benefit from calling it `with_trb_type`). But you can keep the renames as is.

The LinkTRB handling should be robust enough now. Thanks for fixing it. Let's see if we ever skip an interrupt/event with the new warning.

PR #304, opened 2026-08-12
Testutils to mock a RealDevice

Noi0103:
Introduce testutils and demonstrate how to test against a mock RealDevice with the example of the PortArray happy path.

Introducing these testutils here will shrink #238.

### DIFF
--- a/src/device/xhci/port.rs
+++ b/src/device/xhci/port.rs
@@ -389,3 +389,85 @@ impl<CRD: CompleteRealDevice> DeviceRetriever<CRD> {
         Ok(device)
     }
 }
+
+#[cfg(test)]
+mod tests {
+    use std::time::Duration;
+
+    use tokio::{runtime::Handle, time::timeout};
+
+    use crate::device::xhci::{
+        interrupter::tests::testutils::MockInterrupter,
+        real_device::{tests::testutils::MockRealDevice, CompleteRealDeviceImpl},
+    };
+
+    use super::*;
+
+    const ASYNC_TIMEOUT_SECS: u64 = 30;
+    const PORT_ID: u8 = 1;
+    const USB_BUS_NR: u8 = 1;
+    const USB_DEV_NR: u8 = 1;
+    const IDENTIFIER: (u8, u8) = (USB_BUS_NR, USB_DEV_NR);
+
+    #[tokio::test]
+    async fn port_array_hotplug_control_can_attach_list_and_detach() {
+        let async_runtime = Handle::current();
+        let (event_sender, mut interrupter) = MockInterrupter::new();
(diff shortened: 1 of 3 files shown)


src/device/xhci/real_device.rs
```diff
@@ -62,7 +62,7 @@ impl Identifier for (u8, u8) {}
 // the best bet for identification. However, with two identical devices, the
 // approach fails to uniquely identify the devices. Complete real device allows
 // distinction of devices by storing an externally-decided, unique identifier.
-// Host bus-/device-number is an option,// but so are randomly chosen (but unique
+// Host bus-/device-number is an option, but so are randomly chosen (but unique
```
snue: Also missing a closing brace.

```suggestion
// Host bus-/device-number is an option, but so are randomly chosen (but unique)
```

src/device/xhci/port.rs
```diff
@@ -389,3 +389,76 @@ impl<CRD: CompleteRealDevice> DeviceRetriever<CRD> {
...
+        .expect("local timeout on await");
+
+        // expect a successful response for the command and an event
+        assert_eq!(response, Response::SuccessfulOperation);
+        assert_eq!(
+            interrupter.await_event().await,
+            Some(EventTrb::new_port_status_change_event_trb(PORT_ID))
+        );
+
+        assert!(interrupter.is_empty());
```
snue, line 471: Could also add a final check of the `list_devices()` response (for empty vector).

src/device/xhci/port.rs
```diff
@@ -389,3 +389,76 @@ impl<CRD: CompleteRealDevice> DeviceRetriever<CRD> {
...
+    use super::*;
+
+    const ASYNC_TIMEOUT_SECS: u64 = 30;
+    const PORT_ID: u8 = 1;
+    const USB_BUS_NR: u8 = 1;
+    const USB_DEV_NR: u8 = 1;
+    const IDENTIFIER: (u8, u8) = (USB_BUS_NR, USB_DEV_NR);
+
+    #[tokio::test]
+    async fn happy_path_for_attach_and_list_and_detach() {
```
snue:
```suggestion
    async fn port_array_hotplug_control_can_attach_list_and_detach() {
```

snue (commented):
Some minor comments and suggestions. The new unit tests are appreciated. Thanks.

snue:
We could add a test for the issues seen in #302 here. Does the mock trigger cancellation token handling? But then I have no idea how to reliably solve that without more explicit IDs that complicate the handling.

We can of course add such a unit test to #302 after merging this.

snue (approved):

PR #301, opened 2026-08-10
Fix missed context updates and flaky controller reset integration test

Noi0103:
With some open PR's that attempted to merge we ran into issues that the usb device sometimes did not work after a second or later reset (#297  #300). One of the reset tests seemed to fail almost reliable, but it switched back and forth from the tests for usb versions 1.1 and 2.0 not working.

Removing the scsi device before reloading the kernel module seems to resolve the issue.

I tested both mentioned open PR's rebased on this change locally without any issue.

PR #305, opened 2026-08-12
Add missing transition for configure endpoint command

Noi0103:
The transition from addressed state into addressed state is legal according to the endpoint state diagram, but is currently implemented as `ContextStateError`.  

~~Also include a refactor that aligns more with the state machine pattern used in other places.~~

### DIFF
--- a/src/device/xhci/slot_manager.rs
+++ b/src/device/xhci/slot_manager.rs
@@ -540,16 +540,14 @@ impl Slot {
         input_context_pointer: u64,
         deconfigure: bool,
     ) -> anyhow::Result<CompletionCode> {
+        // xhci specification: Figure 4-2: Slot State Diagram
         // configure endpoint with DC=0 transitions from Addressed/Configured to Configured
-        // configure endpoint with DC=1 transitions from Configured to Addressed
+        // configure endpoint with DC=1 transitions from Addressed/Configured to Addressed
         let base_address = match self.state {
             SlotState::Enabled | SlotState::Default(_) => {
                 return Ok(CompletionCode::ContextStateError)
             }
-            SlotState::Addressed(base_address) => match deconfigure {
-                true => return Ok(CompletionCode::ContextStateError),
-                false => base_address,
-            },
+            SlotState::Addressed(base_address) => base_address,
             SlotState::Configured(base_address) => base_address,
         };
 


src/device/xhci/slot_manager.rs
```diff
@@ -563,37 +550,52 @@ impl Slot {
...
-                self.dma_copy_ep_context(ep, input_context_pointer, base_address);
-                self.configure_endpoint(ep, base_address).await?;
+            SlotState::Configured(base_address) => {
+                if deconfigure {
+                    for ep_id in 2..=31 {
+                        self.deconfigure_endpoint(ep_id).await?;
+                    }
+                    self.state = SlotState::Addressed(base_address);
+                } else {
+                    self.state = SlotState::Configured(base_address);
```
snue: That doesn't do anything. And it omits copying endpoint context and configuring the endpoint.

src/device/xhci/slot_manager.rs
```diff
@@ -540,19 +540,6 @@ impl Slot {
...
-        // configure endpoint with DC=1 transitions from Configured to Addressed
-        let base_address = match self.state {
-            SlotState::Enabled | SlotState::Default(_) => {
-                return Ok(CompletionCode::ContextStateError)
-            }
-            SlotState::Addressed(base_address) => match deconfigure {
-                true => return Ok(CompletionCode::ContextStateError),
-                false => base_address,
-            },
-            SlotState::Configured(base_address) => base_address,
```
snue, line 553: It's just this part that was wrong (specifically the `return ... ContextStateError` in `Addressed` state). Both cases should just return the `base_address`.

snue (commented):
> Also include a refactor that aligns more with the state machine pattern used in other places.

It is really beneficial to do refactor and fix in separate commits. With the fix first of course. It helps in the review.

I also spotted an issue with the refactor and find the old logic easier to follow.

snue:
Does this fix the stalled endpoint recovery flow as well? Should it be part of #301?

Noi0103:
I removed the refactor and kept the current version with the fix.

I tested #301 with this added and it did not resolve the issue, so I left it separate.

In theory the transition from the default state is possible and not in the state diagram. Testing that did also not resolve the issue but reduced the `ContextStateError` the linux driver was triggering. But I don't think that we want to allow configure endpoint in the default state without a better reason than that.

snue:
> In theory the transition from the default state is possible and not in the state diagram. Testing that did also not resolve the issue but reduced the ContextStateError the linux driver was triggering. But I don't think that we want to allow configure endpoint in the default state without a better reason than that.

Sounds like the driver and the controller don't agree on the internal controller state. Do you see how the controller transitioned to `Default` state before raising the `ContextStateError`?

Noi0103:
These ContextStateError occur for the attempted transition `Default --(Reset Device)--> ...`. Fixing that might be more fitting for the flaky CI PR mentioned above.

snue (approved):
Ok, it is obviously correct regarding the spec and it simplifies the code. Let's merge it.

Noi0103:
No code changes but a new commit hash to trigger CI and hoping for a good run (current CI flakiness). The store derivation was not available on the builder to press rebuild in the hercules interface.

PR #307, opened 2026-08-26
Fix unit test for Command Abort

Noi0103:
This test did not previously hit the code path it was supposed to when using the crcr::CA bit.

The CommandWorker is in the state `WorkerState::LookingForNewCommand` and needs to handle a `WorkerMessage::Stop` even when there are more Command TRB waiting to be handled.

The tests `flavor = multi_thread` will avoid being stuck after sending the Doorbell. Otherwise, the CommandWorker will hog the available compute.

There is no timeout like behavior for the loop, and it might go on infinitely when the behavior is broken.


src/device/xhci/command_ring.rs
```diff
@@ -657,24 +658,32 @@ mod tests {
...
             .with_trb_type(trb_types::NO_OP_COMMAND)
             .build();
 
+        // Do not use a toggle cycle bit to run an infinite loop.
+        let link_trb = RawTrbBuilder::new(FIFTH_ADDRESS)
+            .with_data_pointer(command.address)
+            .with_trb_type(trb_types::LINK)
+            .build();
+
+        // A chain of NoOp TRB is not allowed, but we do not use the chain bit.
```
snue: True, but rather explain what you set up here

```suggestion
        // Construct a loop of NoOp Command TDs on the ring.
```

src/device/xhci/command_ring.rs
```diff
@@ -657,24 +658,32 @@ mod tests {
         assert!(interrupter.is_empty());
     }
 
-    #[tokio::test]
+    #[tokio::test(flavor = "multi_thread", worker_threads = 1)]
```
snue, line 661: Interesting. We need multi_threading, but a single worker thread is enough. So this changes the task preemption/yield behavior? It's only an issue for the Command Ring? We didn't seem to need it for testing loops on the transfer rings.

Noi0103: Testing the transfer ring did not spawn a worker task in the background. In this case specifically the loop of NoOp is keeping the worker busy infinitely. The scheduler used with "multi_thread"  seems to give us the option to continue even with one thread.

```rust
        // Start the ring through usbcmd and doorbell.
        usbcmd.write(usbcmd::RS);
        command_ring.doorbell().unwrap();

        // stuck and never getting past this with the default tokio::test flavor

        let event = interrupter.await_event().await.unwrap();
```

snue (approved):

PR #309, opened 2026-09-07
Fix swapped ERSTSZ and ERSTBA reads

lbeierlieb:
While comparing the usbvfiod implementation against the xhci spec, Fable found the reads for ERSTSZ and ERSTBA were swapped.

Thanks @phip1611 for running the Fable check and sharing the results with us!


snue (approved):
❤️ 

There are more good findings in the report of course. There will be more PRs soon.

PR #310, opened 2026-09-08
Refactor the default devshell and add a profiling utility devshell

Noi0103:
While gathering profiling information it was helpful to prepare some standard commands packaged as script.

Some of that is already helpful for the default development environment.
The extended list of packages (incl. kcachegrind as a GUI application) is placed in a separate new devShell specifically meant for gathering profiling information.

### DIFF
--- a/nix/devshell.nix
+++ b/nix/devshell.nix
@@ -1,0 +1,23 @@
+{
+  craneLib,
+  pkgs,
+  shellHook,
+}:
+let
+  sshhost = pkgs.writeShellScriptBin "sshhost" ''
+    ssh -p 2000 root@localhost -o UserKnownHostsFile=/dev/null -o StrictHostKeyChecking=no
+  '';
+
+  sshguest = pkgs.writeShellScriptBin "sshguest" ''
+    ssh -p 2000 root@localhost -o ProxyCommand="ssh -W %h:%p -p 2000 root@localhost -o UserKnownHostsFile=/dev/null -o StrictHostKeyChecking=no" -o UserKnownHostsFile=/dev/null -o StrictHostKeyChecking=no
+  '';
+in
+{
+  default = craneLib.devShell {
+    inherit shellHook;
+    packages = [
+      sshhost
+      sshguest
+    ];
+  };
+}
--- a/flake.nix
+++ b/flake.nix
@@ -188,12 +188,9 @@
           };
         };
 
-        devShells.default = craneLib.devShell {
-          shellHook = ''
-            ${self.checks.pre-commit-check.shellHook}
-            alias sshhost=ssh\ -p\ 2000\ root@localhost\ -o\ UserKnownHostsFile=/dev/null\ -o\ StrictHostKeyChecking=no
-            alias sshguest=ssh\ -o\ ProxyCommand="ssh\ -W\ %h:%p\ -p\ 2000\ root@localhost\ -o\ UserKnownHostsFile=/dev/null\ -o\ StrictHostKeyChecking=no"\ -o\ UserKnownHostsFile=/dev/null\ -o\ StrictHostKeyChecking=no\ root@192.168.100.2
-          '';
+        devShells = import ./nix/devshell.nix {
+          inherit craneLib pkgs;
+          inherit (self.checks.pre-commit-check) shellHook;
         };
 
         herculesCI = {


nix/devshell.nix
```diff
@@ -0,0 +1,121 @@
...
+    cargo build --bin usbvfiod
+    cargo build --bin remote
+    nix-build -A netboot.x86_64-linux '<nixpkgs/nixos/release.nix>'
+
+    echo ""
+    echo "FIND YOUR BUS/DEVICE NUMBERS"
+    lsusb
+    echo ""
+    echo "LIST OF COMMANDS TO USE IN THE GUEST ON THE BLOCKDEVICE:"
+    echo "sudo su"
```
snue, line 48: nit: that's basically `sudo -i` (interactive) ... but why should that be needed in the guest? The scripts just use the `root` login directly.

nix/devshell.nix
```diff
@@ -0,0 +1,121 @@
...
+    rm ${vfio-user-socket}
+    rm ${hotplug-socket}
+    ./target/debug/usbvfiod --socket-path ${vfio-user-socket} --hotplug-socket-path ${hotplug-socket} --no-color
+  '';
+
+  runchv = pkgs.writeShellScriptBin "runchv" ''
+    nix-build -A netboot.x86_64-linux '<nixpkgs/nixos/release.nix>'
+    NETBOOT=$(grep -o "init=[^$]*" result/netboot.ipxe)
+    echo "$NETBOOT"
+    cloud-hypervisor --memory size=4G,shared=on --serial tty --user-device socket=${vfio-user-socket} --console off --kernel result/bzImage --initramfs result/initrd --cmdline "$NETBOOT console=ttyS0"
```
snue, line 29: But `cloud-hypervisor` is not in the `devShell`.

snue (commented):
I like the cleanup and the new profiling shell. Left minor remarks. The names for the various shell helpers may not be the most ergonomic, but let's not waste any time with naming for the dev shell helpers.

nix/devshell.nix
Noi0103: Thanks for catching. (..something something direnv impurities from the system config...)

nix/devshell.nix
Noi0103: Running the net boot image the cloud-hypervisor command will auto login as normal user. The line is gone now since error messages about missing privileges will suffice here.

snue: I see this was intended for the interactive shell on `ttyS0`/`stdio` from the `runchv` helper, which has auto-login as `nixos`, not through `sshguest`. That's fine.

nix/devshell.nix
```diff
@@ -0,0 +1,189 @@
...
+
+  # info howto shell -> describing what the helper do
+  infoMessage = pkgs.writeShellScriptBin "infoMessage" ''
+    echo "INFO: Set \$BUILD to 'release' for release binaries (default: 'debug')."
+    echo "INFO: This devShell provides a few simple QoL scripts."
+    echo "  sshhost       connect to a qemu guest of a nix integration test"
+    echo "  sshguest      connext to a nested cloud hypervisor guest of a nix integration test"
+    echo "  runusbvfiod   usbvfiod wrapper with hotplug enabled; foreground job using stdio for logs"
+    echo "  runchv        cloud hypervisor wrapper expecting a usbvfiod socket; foreground job and uses stdio for interaction & logs"
+    echo "  attach        remote wrapperto attach a usb device to the running usbvfiod; foreground job and uses stdio for logs"
```
snue: nit: missing a space, and as it is a short-lived command invocation, we don't really need the foreground mention.
 
```suggestion
    echo "  attach        remote wrapper to attach a usb device to the running usbvfiod; call with BUS DEV arguments"
```

nix/devshell.nix
```diff
@@ -0,0 +1,189 @@
...
+    else
+      cargo build --bin remote --release
+    fi
+
+    ./target/$BUILD/remote --socket ${hotplug-socket} --attach /dev/bus/usb/$1/$2
+  '';
+
+  # info howto shell -> describing what the helper do
+  infoMessage = pkgs.writeShellScriptBin "infoMessage" ''
+    echo "INFO: Set \$BUILD to 'release' for release binaries (default: 'debug')."
```
snue, line 53: nit: I hope this generic name does not conflict with the actual build commands the scripts invoke.

nix/devshell.nix
```diff
@@ -0,0 +1,189 @@
...
+
+      if [ "$answer" = "y" ] || [ "$answer" = "yes" ]; then
+        cat > .cargo/config.toml << 'EOF'
+      [build]
+      rustflags = [
+        "-C",
+        "force-frame-pointers=yes",
+        "-C",
+        "symbol-mangling-version=v0",
+      ]
```
snue: Can we instead just add these flags directly in the build invocation?

https://doc.rust-lang.org/cargo/reference/config.html#command-line-overrides

(or maybe through a profile, see other comment)

nix/devshell.nix
```diff
@@ -0,0 +1,189 @@
...
+    pkgs.cloud-hypervisor
+    sshhost
+    sshguest
+    runusbvfiod
+    runchv
+    attach
+  ];
+
+  # Prepare cached binaries for other scripts and give some very likely needed output.
+  preparegrind = pkgs.writeShellScriptBin "preparegrind" ''
```
snue, line 74: Should this include at least a `cargo clean`? I currently don't see the benefit of this, as every other script contains basically the same snippet.

nix/devshell.nix
```diff
@@ -0,0 +1,189 @@
...
+    if [ "$BUILD" = "debug" ]; then
+      cargo build --bin usbvfiod
+      cargo build --bin remote
+    else
+      cargo build --bin usbvfiod --release
+      cargo build --bin remote --release
+    fi
+    nix-build -A netboot.x86_64-linux '<nixpkgs/nixos/release.nix>'
```
snue: nit: I feel this should provide an explicit result symlink directory. But the default location is at least ignored by git already.

nix/devshell.nix
```diff
@@ -0,0 +1,189 @@
...
+
+  # Prepare cached binaries for other scripts and give some very likely needed output.
+  preparegrind = pkgs.writeShellScriptBin "preparegrind" ''
+    if [ "$BUILD" = "debug" ]; then
+      cargo build --bin usbvfiod
+      cargo build --bin remote
+    else
+      cargo build --bin usbvfiod --release
+      cargo build --bin remote --release
+    fi
```
snue, line 81: For profiling runs, a full `release` profile is probably not good. There is no debuginfo available for the binaries. I have not checked how useful the generated reports are, though.

We may want this flag to be `BUILD_PROFILE` that defaults to `dev` instead, supply this with `--profile $...` instead of the if/else. And maybe add a separate one for optimized profiling?

See https://doc.rust-lang.org/cargo/reference/profiles.html

nix/devshell.nix
```diff
@@ -0,0 +1,189 @@
...
+  };
+
+  profiling = craneLib.devShell {
+    # used as guidance: https://nnethercote.github.io/perf-book/profiling.html
+    shellHook = ''
+      ${shellHook}
+      ${infoMessage}/bin/infoMessage
+      echo "INFO: This devShell, meant for profiling, also provides the following."
+      echo "  valgrind      valgrind package"
+      echo "  kcachegrind   kcachegrind package"
```
snue:
```suggestion
      echo "  valgrind      profiling toolkit"
      echo "  kcachegrind   GUI for valgrind reports"
```

nix/devshell.nix
```diff
@@ -0,0 +1,189 @@
...
+
+    ./target/$BUILD/remote --socket ${hotplug-socket} --attach /dev/bus/usb/$1/$2
+  '';
+
+  # info howto shell -> describing what the helper do
+  infoMessage = pkgs.writeShellScriptBin "infoMessage" ''
+    echo "INFO: Set \$BUILD to 'release' for release binaries (default: 'debug')."
+    echo "INFO: This devShell provides a few simple QoL scripts."
+    echo "  sshhost       connect to a qemu guest of a nix integration test"
+    echo "  sshguest      connext to a nested cloud hypervisor guest of a nix integration test"
```
snue:
```suggestion
    echo "  sshguest      connect to a nested cloud hypervisor guest of a nix integration test"
```

snue (commented):
Some more comments and spotted typos. If it turns out a separate `profiling` cargo profile is helpful, we can add that to the project. Or maybe the standard `bench` profile is already a good fit.

nix/devshell.nix
Noi0103: Ah, I found profiles and the environment variable alternatives but somehow not the `--config` argument.

nix/devshell.nix
Noi0103: This is just a personal preference to have the binaries and netboot image already cached instead of building things on demand and have the reminder for the `dd` command to use.

The `dd` command example could move to the `shellHook`, building be done on demand and `lsusb` be a user responsibility. I'm not sure what is best here.

nix/devshell.nix
Noi0103: I would like to avoid creating symlinks _somewhere_, unless we can assume they are periodically removed by the system like in /tmp.

nix/devshell.nix
Noi0103: Maybe `result-netboot` is a good symlink. It is not `result`, already in the `.gitignore` and keeps it in the project directory.

snue: From the name I expected this to ensure you get a clean build with the ideal settings for a profiling run ("grind"). That still sounds useful to me.

nix/devshell.nix
Noi0103: For the moment I prefer the command line arguments over the profiles. I see how this is then leaning into nix more than it would have to.

I would like to keep it with the argument flags mentioned above until we have a stronger incentive for multiple profiles.

nix/devshell.nix
snue: How would something like `result-netboot` be different from the default `result` in the working directory? Our `.gitignore` already contains `result*`, so we are basically prepared for this. I would like to avoid confusion with concurrent use of `nix-build` by the user in the working directory.

```
     •  --out-link / -o outlink
        Change the name of the symlink to the output path created from result to outlink.
```

Noi0103: The difference would only be in the name. I would expect that to be clear enough for nix users to not confuse it with `usbvfiod` under `result`.

```diff
- nix-build -A netboot.x86_64-linux '<nixpkgs/nixos/release.nix>'
- NETBOOT=$(grep -o "init=[^$]*" result/netboot.ipxe)
+ nix-build -A netboot.x86_64-linux '<nixpkgs/nixos/release.nix>' --out-link result-netboot
+ NETBOOT=$(grep -o "init=[^$]*" result-netboot/netboot.ipxe)
```

nix/devshell.nix
Noi0103: Doesn't `cargo clean` just remove old binary artifacts like a garbage collection? That should not impact profiling if the `cargo build` places the latest version under `./target` as far as I understand it.

nix/devshell.nix
snue: Exactly, let's just use that.

My concern was if a user invokes `nix-build`, then any of the helpers that call `nix-build` and then wants to use the previous `result` in an invocation, it will fail because `result` was replaced within the helper.

nix/devshell.nix
Noi0103: I have not had any perceivable issues so far at least.

nix/devshell.nix
```diff
@@ -0,0 +1,162 @@
...
+    rm -rf ${vfio-user-socket} ${hotplug-socket}
+
+    ${pkgs.valgrind}/bin/valgrind --tool=callgrind \
+      --callgrind-out-file=callgrind.out.%p \
+      --compress-strings=no \
+      --separate-threads=yes \
+      --dump-line=yes \
+      --collect-systime=yes \
+      --collect-bus=yes \
+      ./target/debug/usbvfiod --socket-path ${vfio-user-socket} --hotplug-socket-path ${hotplug-socket} --no-color
```
snue, line 123: Some of the `./target` calls don't respect the actual build target.

nix/devshell.nix
```diff
@@ -0,0 +1,162 @@
...
+
+  attach = pkgs.writeShellScriptBin "attach" ''
+    if [ "$#" -ne 2 ]; then
+      echo "Error: expected 2 arguments: /dev/bus/usb/\$1/\$2"
+      exit 1
+    fi
+
+    cargo build --bin remote
+
+    ./target/debug/remote --socket ${hotplug-socket} --attach /dev/bus/usb/$1/$2
```
snue, line 49: nit: put all bash variable expansion into double-quotes to prevent word splitting. Especially for tools you expect to call as root occasionally.

nix/devshell.nix
```diff
@@ -0,0 +1,162 @@
...
+  sshguest = pkgs.writeShellScriptBin "sshguest" ''
+    ssh -p 2000 root@localhost -o ProxyCommand="ssh -W %h:%p -p 2000 root@localhost -o UserKnownHostsFile=/dev/null -o StrictHostKeyChecking=no" -o UserKnownHostsFile=/dev/null -o StrictHostKeyChecking=no
+  '';
+
+  runusbvfiod = pkgs.writeShellScriptBin "runusbvfiod" ''
+    if [ "$BUILD" = "debug" ]; then
+      cargo build --bin usbvfiod
+    else
+      cargo build --bin usbvfiod --release
+    fi
```
snue, line 23: nit: It looks like a target definition, but it's just "debug" or "something else" ... I'd much rather see something where you can expand the variable here and not duplicate the cargo call in the if/else.

snue (commented):
I have some minor comments, but this looks helpful already.

nix/devshell.nix
Noi0103: Thanks for catching those three.

With the `attach` script I have some trouble and decided always using debug will be just fine.
For running this in a `nix develop` and needing permissions to USB devices, a fast way is sudo. Doing so will run this script without the environment variables and fail with the error:
```
/nix/store/kf2ij3mzb3kvdkjfrhjprrz4yh4fsrw7-attach/bin/attach: line 13: ./target//remote: No such file or directory
```

Ironically, direnv integration in zsh does not have the same issue.

snue: Yes, `sudo` will clean the environment by default to avoid evil. This is expected. One can use `-E` (`--preserve-env`), but you have to remember or even know that.

Using the `debug` binary is fine, but make sure it is actually built/available also when running `release` builds otherwise.

snue (approved):
Latest version looks good. There are still some opportunities to quote bash variables, but they don't block merging.

Noi0103:
Adding quotes for the `$1` and `$2` arguments in the `attach` script. I thought I did at least those after @snue mentioned that, but it seems I lost the changes, so here they are again.

PR #311, opened 2026-09-10
Add a missing slot context update

Noi0103:
The xhci specification chapter 4.6.11 Reset Device:

> - If the Device Slot is in the Addressed or Configured state:
>   - [...]
>   - Set the Slot State field of Slot Context to the Default state

When doing a block device formatting test on Windows, this resolved a `ContextStateError`.


snue (approved):
Yes, this is an obvious miss. I did a short check, and it looks like this is the only location we missed for writing back the slot state.

PR #315, opened 2026-09-17
Implement long Control Transfers

Noi0103:
Attempting to split #238 into smaller pieces.

Currently, only two variants for Control Transfers are possible:
1. `SetupStageTrb` --> `StatusStageTrb`
2. `SetupStageTrb` --> `DataStageTrb` --> `StatusStageTrb`

With these changes longer chains will also be possible, e.g.:
- `SetupStageTrb` --> `DataStageTrb` --> `NormalTrb` --> `StatusStageTrb`
- `SetupStageTrb` --> `DataStageTrb` --> `NormalTrb` --> `NormalTrb` --> `StatusStageTrb`


Noi0103:
For some function names I did not yet come up with something better. Using terms like initialize and realize is likely not optimal.

src/device/xhci/endpoint_handle.rs
```diff
@@ -105,66 +199,438 @@ impl<RCEH: RealControlEndpointHandle> ControlEndpointHandle<RCEH> {
...
+
+    /// for Control Out
+    fn collect_transfer_data<T: TrbDmaInfo>(&mut self, trb: &T) {
+        if trb.has_immediate_data() {
+            // Only event data should follow when immediate data is used here
+            // but we do not check for that and allow multiple immediate data
+            // TRB in the data stage TD.
+
+            self.transfer_state.usb_request.data.append(
+                &mut trb.data_pointer().to_le_bytes()[..trb.transfer_length() as usize].to_vec(),
```
snue: You need the `min(8, transfer_length)` logic here too to prevent OOB access.

src/device/xhci/endpoint_handle.rs
```diff
@@ -105,66 +199,438 @@ impl<RCEH: RealControlEndpointHandle> ControlEndpointHandle<RCEH> {
...
+    fn collect_transfer_length<T: TrbDmaInfo>(&mut self, trb: &T) {
+        let transfer_length = if trb.has_immediate_data() {
+            min(8, trb.transfer_length())
+        } else {
+            trb.transfer_length()
+        };
+
+        // SAFETY: The UsbRequest struct reasons, that the wLength field is not
+        // usable, but we can still use the field size to assume a maximum
+        // Control Request transfer length that will fit in a u16.
```
snue: I don't understand this comment. You just accumulate transfer_length and don't look at `wLength` here?

src/device/xhci/endpoint_handle.rs
```diff
@@ -105,66 +199,438 @@ impl<RCEH: RealControlEndpointHandle> ControlEndpointHandle<RCEH> {
...
+        self.event_sender.send(event)?;
+
+        Ok(())
+    }
+
+    fn realize_slice<T: TrbDmaInfo>(
+        &self,
+        trb: &T,
+        hardware_data: &mut Vec<u8>,
+    ) -> anyhow::Result<()> {
```
snue: What do you plan to use the result for? Currently, it never returns an `Err` value. Maybe returning the amount of data transferred, or the residue makes more sense if you don't want to generate the events for this slice here but in the caller. Or rather an enum value with the required info for the event generation (in case the caller wants to generate an event).

src/device/xhci/endpoint_handle.rs
```diff
@@ -105,66 +199,438 @@ impl<RCEH: RealControlEndpointHandle> ControlEndpointHandle<RCEH> {
...
+        };
+
+        // SAFETY: The UsbRequest struct reasons, that the wLength field is not
+        // usable, but we can still use the field size to assume a maximum
+        // Control Request transfer length that will fit in a u16.
+        self.transfer_state.usb_request.length = self
+            .transfer_state
+            .usb_request
+            .length
+            .wrapping_add(transfer_length as u16);
```
snue: The `transfer_length` is a 17 bit value.

src/device/xhci/endpoint_handle.rs
```diff
@@ -105,66 +199,438 @@ impl<RCEH: RealControlEndpointHandle> ControlEndpointHandle<RCEH> {
...
+            ControlTransferState::MoreData => match &supported_trb.variant {
+                SupportedControlEndpointTrb::SetupStage(trb) => {
+                    info!(
+                        "received Setup Stage TRB abort ongoing control transfer in favour of this new one"
+                    );
+
+                    // backtrack to send healthy events before we drop the trbs
+                    for trb in &self.submission_state.trbs {
+                        self.transfer_event_success(trb.addr)?;
+                    }
```
snue: What healthy events? We didn't do any hardware request here yet? We shouldn't signal anything for transfers that did not happen.

src/device/xhci/endpoint_handle.rs
```diff
@@ -174,69 +640,105 @@ impl<RCEH: RealControlEndpointHandle> EndpointHandle for ControlEndpointHandle<R
...
+                    );
+                    self.event_sender.send(event)?;
+
+                    // these trb are reported back, we reset now and await to continue
+                    self.submission_state.trbs.clear();
+                    self.transfer_state.state = ControlTransferState::ExpectSetupStageTrb;
+
+                    TrbProcessingResult::TrbError
+                }
+                ControlSubmissionState::OutOfOrderTrb(supported_trb) => {
```
snue: Does this differ from `UnexpectedTrb` except for the message?

src/device/xhci/endpoint_handle.rs
```diff
@@ -174,69 +640,105 @@ impl<RCEH: RealControlEndpointHandle> EndpointHandle for ControlEndpointHandle<R
...
-                                CompletionCode::Success,
-                                false,
-                                self.endpoint_id,
-                                self.slot_id,
-                            );
-                            self.event_sender.send(event)?;
+                    // backtrack to send healthy events until the bad is encountered
+                    for trb in &self.submission_state.trbs {
+                        self.transfer_event_success(trb.addr)?;
+                    }
```
snue: We didn't actually submit these on the `real_ep`. Should we really claim the transfers happened? What does the spec say? (Not generating events for these if requested also seems wrong, though.)

src/device/xhci/endpoint_handle.rs
```diff
@@ -105,66 +199,438 @@ impl<RCEH: RealControlEndpointHandle> ControlEndpointHandle<RCEH> {
...
+        }
+
+        Ok(())
+    }
+
+    /// if direction { IN } else { OUT }
+    fn realize_control_chain(
+        &self,
+        hardware_data: &mut Vec<u8>,
+        direction: bool,
```
snue: Isn't `direction` redundant with the availability of hardware data here? Can you make that an `Option` instead?

snue (commented):
I left some comments, but I didn't review everything completely yet. Thanks for splitting up #238, that should help us get this merged more quickly.

src/device/xhci/endpoint_handle.rs
Noi0103: Correct, I do not look at the value of `wLength`. I do however use the size of the `wLength` field to justify using only a `u16` to store the accumulated `transfer_length`.

To avoid writing the same, I want to reference the explanation in `usb_request.rs`.

src/device/xhci/endpoint_handle.rs
Noi0103: I considered using an option, but found this explicit boolean to be easier to understand with the specification.

I'm open to change it into an Option and rewriting the comment to be something along the lines of
```rust
/// if hardware_data.is_some() { IN } else { OUT }
```

src/device/xhci/endpoint_handle.rs
Noi0103: Yes, right now this is dead code I did not remove. When adding a proper short packet handling parts of those results should become useful.

src/device/xhci/endpoint_handle.rs
Noi0103: I differentiated for debugging purposes.
`UnexpectedTrb` is a successfully parsed TRB that should not appear on any the Control Endpoint as per the list of supported Endpoints.
`OutOfOrder` on the other hand is a whitelisted TRB type for Control Endpoints. If the state machine for the Control endpoint is wrong this will help us find that it is an error in state transitions and not allowed TRB types.

src/device/xhci/endpoint_handle.rs
Noi0103: This is tricky and with the proposed approach we might be able to desync some state between device and system software.
Not sending any event and jumping to the bad TRB is worse according to some very very minimal testing where the event received did not match the expected events (since those were skipped).

Just sending a success has worked in some cases is the past. Should this be logged at info level until we have an issue or better way how to do this?

src/device/xhci/endpoint_handle.rs
Noi0103: After talking offline this will change and avoid deriving anything from `wLength`.

snue: We saw `nusb` control requests also take the length as `u16`, so with the current backend we cannot really supply anything larger than `wLength` would allow. It's most likely not required at all, but we should at least warn when we see large transfers here that we truncate.

src/device/xhci/endpoint_handle.rs
snue: Ideally the state machine issues would all be covered/caught by unit tests already and a smart compiler could prove this is dead code. But the guest may still issue illegal `OutOfOrder` sequences?  The extra code is quite some duplication. Maybe you can at least cover both in the same match arm if they imply the same response?

src/device/xhci/endpoint_handle.rs
snue: Yes, let's log this at info level. And please guard it with an actual `IOC` flag check before generating events.

src/device/xhci/endpoint_handle.rs
```diff
@@ -349,8 +357,8 @@ impl<RCEH: RealControlEndpointHandle> ControlEndpointHandle<RCEH> {
                     }
                 }
                 SupportedControlEndpointTrb::Normal(normal) => {
-                    if direction {
-                        self.realize_slice(normal, hardware_data)?;
+                    if hardware_data.is_some() {
+                        self.realize_slice(normal, hardware_data.as_mut().unwrap());
                     }
```
snue, line 375: nit: can this be `if let Some()`?

src/device/xhci/endpoint_handle.rs
```diff
@@ -530,14 +533,11 @@ impl<RCEH: RealControlEndpointHandle> EndpointHandle for ControlEndpointHandle<R
                         "received Setup Stage TRB abort ongoing control transfer in favour of this new one"
                     );
 
-                    // backtrack to send healthy events before we drop the trbs
-                    for trb in &self.submission_state.trbs {
-                        self.transfer_event_success(trb.addr)?;
-                    }
+                    self.clean_current_trbs()?;
 
                     self.submission_state.trbs.clear();
```
snue: redundant with the `clean_current_trbs()`

src/device/xhci/endpoint_handle.rs
```diff
@@ -576,14 +576,11 @@ impl<RCEH: RealControlEndpointHandle> EndpointHandle for ControlEndpointHandle<R
...
```
snue: also redundant now.

src/device/xhci/endpoint_handle.rs
```diff
@@ -264,6 +261,29 @@ impl<RCEH: RealControlEndpointHandle> ControlEndpointHandle<RCEH> {
...
+                SupportedControlEndpointTrb::SetupStage(setup) => setup.interrupt_on_completion,
+                SupportedControlEndpointTrb::DataStage(data) => data.interrupt_on_completion,
+                SupportedControlEndpointTrb::Normal(normal) => normal.interrupt_on_completion,
+                SupportedControlEndpointTrb::StatusStage(status) => status.interrupt_on_completion,
+                SupportedControlEndpointTrb::EventData(event_data) => {
+                    event_data.interrupt_on_completion
+                }
+            };
+            if interrupt_on_completion {
+                self.transfer_event_success(trb.addr)?;
```
snue, line 291: Yes this is better. Just needs an `info` message. I hope we don't hit that for every short string descriptor transfer later.

Noi0103: I tried to search for the message in a full `nix flake check` output log and have not been able to find them (of course that does not guarantee anything with real hardware).

src/device/xhci/endpoint_handle.rs
Noi0103: Combined the `OutOfOrderTrb` into the `UnexpectedTrb`. It originated from debugging ease, but it is too much duplication for a very limited “might happen” case.

src/device/xhci/trb.rs
```diff
@@ -1149,7 +1149,10 @@ pub enum TrbParseError {
 }
 
 #[derive(Debug, Clone)]
-pub struct SupportedEndpointTrb<T> {
+pub struct SupportedEndpointTrb<T>
+where
+    T: TryFrom<TransferTrbVariant, Error = TransferTrbVariant>,
+{
```
snue: nit: this landed in the wrong commit.

I also wonder if you need `TryFrom` here when the error value is the same type (just the `UnrecognizedTrb` variant). The conversion is infallible with that mapping, but `TryFrom`/`Err` may still make sense with the use.

src/device/xhci/endpoint_handle.rs
```diff
@@ -79,16 +80,121 @@ impl BaseEndpointHandle for DummyEndpointHandle {
...
+
+    fn try_from(trb: TransferTrbVariant) -> Result<Self, Self::Error> {
+        match trb {
+            TransferTrbVariant::SetupStage(t) => Ok(Self::SetupStage(t)),
+            TransferTrbVariant::DataStage(t) => Ok(Self::DataStage(t)),
+            TransferTrbVariant::StatusStage(t) => Ok(Self::StatusStage(t)),
+            TransferTrbVariant::Normal(t) => Ok(Self::Normal(t)),
+            TransferTrbVariant::EventData(t) => Ok(Self::EventData(t)),
+            variant => Err(variant),
+        }
```
snue: Ah yes, this use of `TransferTrbVariant` for the error type is alright.

src/device/xhci/endpoint_handle.rs
```diff
@@ -105,138 +211,530 @@ impl<RCEH: RealControlEndpointHandle> ControlEndpointHandle<RCEH> {
...
+                        let event = EventTrb::new_transfer_event_trb(
+                            trb.addr,
+                            0,
+                            CompletionCode::Success,
+                            false,
+                            self.endpoint_id,
+                            self.slot_id,
+                        );
+                        self.event_sender.send(event)?;
+                    }
```
snue: Don't you have `self.transfer_event_success()` for that? You still generate `CC::Success` events manually in some places.

src/device/xhci/endpoint_handle.rs
```diff
@@ -105,138 +211,530 @@ impl<RCEH: RealControlEndpointHandle> ControlEndpointHandle<RCEH> {
...
+                    info!(
+                        "invalid control transfer sequence; expected Setup Stage Trb, got: {:?}",
+                        supported_trb
+                    );
+
+                    self.transfer_state.state = ControlTransferState::ExpectSetupStageTrb;
+
+                    self.submission_state.state = ControlSubmissionState::UnexpectedTrb(
+                        supported_trb.addr,
+                        TransferTrbVariant::from(supported_trb.variant),
```
snue: nit: would `supported_trb.variant.into()` work here?

src/device/xhci/endpoint_handle.rs
```diff
@@ -105,138 +211,530 @@ impl<RCEH: RealControlEndpointHandle> ControlEndpointHandle<RCEH> {
...
+            0,
+            CompletionCode::Success,
+            false,
+            self.endpoint_id,
+            self.slot_id,
+        );
+        self.event_sender.send(event)
+    }
+
+    fn realize_slice<T: TrbDmaInfo>(&self, trb: &T, hardware_data: &mut Vec<u8>) {
```
snue: This handles Control IN DMA transfers. I think the name should reflect it.

`copy_to_guest` maybe? Or just `data_dma_write`.

src/device/xhci/endpoint_handle.rs
```diff
@@ -105,138 +211,530 @@ impl<RCEH: RealControlEndpointHandle> ControlEndpointHandle<RCEH> {
...
+        // check length for short packet
+        if hardware_data.len() < trb.transfer_length() as usize {
+            // TODO This is a very minimal handling of a short packet that the
+            // linux driver will tolerate. Windows and others might need
+            // proper/spec compliant handling.
+
+            warn!(
+                "ControlEndpoint in slot {} encountered ShortPacket (incomplete implementation)",
+                self.slot_id
```
snue: Do we hit this warning in the CI logs already or just with actual devices?

snue (commented):
This tolerates quite the abuse of `wLength` now. Thanks for adding the unit tests as well. The devices described in the spec would only use a non-zero `wLength` without actual data stage. I don't know if `nusb` control requests can represent/translate that. In any case, the new handling should be fine.

I left some comments, but no complaints about the general approach.

src/device/xhci/endpoint_handle.rs
Noi0103: On a full `nix flake check`we hit this message 373 times. A follow-up PR will touch this again soon.

snue: Yes, a bit ugly, but with #320 already open I won't block the merge for this.

src/device/xhci/endpoint_handle.rs
```diff
@@ -1152,4 +1650,959 @@ pub mod tests {
...
+    /// > - If a Data Stage TD does not follow a Setup Stage TD, where wLength > ‘0’.
+    /// >
+    /// > [...]
+    /// >
+    /// > This condition violates the definition of a USB Control Transfer, however this condition should be ignored by the
+    /// > xHC to ensure legacy device compatibility. The Setup Stage Transfer Type (TRT) field strictly indicates the
+    /// > presence and the Direction of the Data Stage TD, and determines the direction of the Status Stage TD so the
+    /// > wLength field should be ignored by the xHC.
+    #[tokio::test]
+    async fn submit_control_out_with_some_wlength_but_do_not_automatically_expect_data() {
```
snue: The test and comment is good, but we should mention that the guest provided mismatching `wLength` will not be transmitted/forwarded to the real device abstraction. Such non-compliant devices are most likely still broken behind the virtual controller.

Luckily we don't plan to support any such devices. We just want the controller to handle such requests gracefully and in a spec-compliant way.

PR #318, opened 2026-09-23
Load kvm module in initrd to prevent an initial service error

Noi0103:
Enables #317

This should prevent the cloud hypervisor systemd service to attempt running before the KVM module is loaded.
This currently leads to a flaky file descriptor test that relies on clean service logs without any error messages.

I did a rebase onto #317 and the CI did not encounter the issue. ~~On the PR itself CI hit the issue in 2 of the 2 attempts.~~ On #317 itself the CI hit the issue in 2 of the 2 attempts so far.


nix/checks/testutils.nix
```diff
@@ -131,7 +131,7 @@ let
       isSystemUser = true;
       group = "usbaccess";
     };
-    boot.kernelModules = [ "kvm" ];
+    boot.initrd.kernelModules = [ "kvm" ];
```
snue: I think we want the specific hardware modules here as well. Not sure what the architecture of the current runner is, but including both choices hopefully doesn't hurt.
 
```suggestion
    boot.initrd.kernelModules = [ "kvm" "kvm-intel" "kvm-amd" ];
```

Or maybe the `intel` and `amd` modules just need to be made "available" in the `initrd` to ensure they get loaded early when the hardware matches.

Noi0103: Do you mean even thought the current state does already seem to fix the issue, we should mention/add hardware specific modules?

I just realized I might have worded the initial descriptions last paragraph bad.

snue: I'm not entirely convinced this is a fully sufficient fix or that it could regress if the hardware/platform-specific module is not available at the time the regular `kvm` module loads. Being thorough and at least explicitly including these in the `initrd.availableKernelModules` if not in the explicitly loaded `kernelModules` seems more robust long term.

The logs show this change moved the `kvm` module load much earlier, so it is likely enough to fix the issues we observed. And eventually we will want to test this on ARM as well. So referencing x86-specific modules isn't ideal. We can go with just the `kvm` module until we see another regression/flakiness.

Noi0103: I tested both variants on temporary branches in the meantime and while I could run the full `nix flake check` locally, the CI runner fails all integration tests under `checks.aarch64-linux` because of a failed dependency/derivation not finding modules:
```
root module: kvm-amd
modprobe: FATAL: Module kvm-amd not found in directory /nix/store/l088wanwizihq5naji3f9h97nma1lqji-linux-6.18.49-modules/lib/modules/6.18.49
```

snue: Yeah, let's not bother with the architecture dependencies for now. Just loading `kvm` early is fine.

snue (approved):

